PR #4, opened 2017-05-31
Custom encoding support

ondrejpialek:
Adds a configuration setting that allows for custom encoding to be passed in the Run Source command.

In the future this could be taken directly off the editor as it knows the encoding of the file being edited, this property is however not exposed to the JS API yet.

One questionable thing about this PR is that it provides UTF-8 as a default and therefore changes the default Run Source call `source(FILE)` to `source(FILE, encoding="UTF-8")`, which might be a breaking change. UTF-8 is a default file encoding in VS Code so for people predominantly working in this IDE it is the preffered setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,6 +153,11 @@
           "default": "/usr/bin/R",
           "description": "R terms path for Linux."
         },
+        "r.rterm.encoding": {
+          "type": "string",
+          "default": "UTF-8",
+          "description": "An optional encoding to pass to R when executing the file, i.e. 'source(FILE, encoding=ENCODING)'"
+        },
         "r.lintr.linters": {
           "type": "string",
           "default": "default_linters",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -64,12 +64,15 @@ export function activate(context: ExtensionContext) {
 
     function runSource()  {
         const Rpath = ToRStringLiteral(window.activeTextEditor.document.fileName, '"');
-        
+        var encodingParam = config.get('rterm.encoding');
+        if (encodingParam) {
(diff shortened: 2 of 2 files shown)


Ikuyadeu:
@ondrejpialek  Thank you for your contribution! Looks good.
But I can't test in this week. Please wait 2,3 days for merge.
I think your change is very good.

Ikuyadeu:
In this page (https://stat.ethz.ch/R-manual/R-devel/library/base/html/source.html), source function has any args. Can you rename `encoding` in some values to `args`?
But I don't know R so much, so this may be bad suggestion.

Ikuyadeu:
Ok, It is very good! Thank you @ondrejpialek !

ondrejpialek:
Thank you @Ikuyadeu for accepting my PR!

Btw I noticed you made some changed to the code to make it more readable. I just wanted to clarify why it was coded like that: I wanted to allow people NOT to supply an encoding parameter and get the previous behaviour. They would do that by setting the `r.source.encoding` setting to an empty string.

Consider the following examples:

`"r.source.encoding": "UTF-8"`:

- Before change: `source("file.R", encoding="UTF-8")`
- After change: `source("file.R", encoding="UTF-8")` (no change)

`"r.source.encoding": ""`:

- Before change: `source("file.R")` (previous behaviour of this extension)
- After change: `source("file.R", )` (looks weird but still works)

Just wanted to point this out. The script still works, it's just that there is an extra comma. Up to you if it bothers you or not.

Many thanks again and 👍

Ikuyadeu:
@ondrejpialek Thank you for great advice! 
Now I fixed this problem that extra comma.

PR #31, opened 2017-10-28
Proposed fix for Load Chunk problems #27

### LINKED ISSUE #27: `Run line` should load code chunk
In #26,  It's very useful function.
> Okay, I found the way to run a line (just like RStudio, its Cmd + Enter, and the run source utility is useful. However, I'm wondering about running just individual code chunks, e.g.
> 
> 
> code_chunk <- function(x) {
>     cat("I am a chunk")
> }
> 
> code_chunk_2 <- x %>%
>   mutate(...) %>%
>   transmute(...)
> 
> If I play my cursor at the beginning of line 1 with the first code_chunk, and enter C-c (aka Ctrl+c, or some other shortcut, say to run a region, like Ctrl+r) then it would run all three lines at once (the whole chunk), and not run code_chunk_2, and if this could be done without necessarily highlighting the chunk prior to running the code.
> 
> Just wondering if you might have some insight on how this could be done, or alternately if this could be implemented in an update?

Ladvien:
UPDATE: I realize this didn't address #27. Instead, it address RTerm choking when Run.Selection is given too much text _and_ sets us up to fix #27, which I'll handle if this PR passes review.

After some testing, it seems when you send too much text to be processed by RTerm a buffer is overrun somewhere.  Luckily, R executes line-by-line.  

I've split the selected text into a string[] and then loop over the array, sending a line at a time to the RTerm, with a small delay in between.  

*Adding a delay is not the best way to do this.* I realize it.  But without having access to output from the RTerm, I'm not sure how else to hack it.  I'm open to suggestions.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,7 +49,6 @@ export function activate(context: ExtensionContext) {
         const selectedTextArray = selectedLineText.split("\n");
 
         return selectedTextArray;
-
     }
 
     async function runSelection() {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -38,28 +38,34 @@ export function activate(context: ExtensionContext) {
         setFocus();
     }
 
-    function getSelection(): string {
+    function getSelection(): string[] {
         const { start, end } = window.activeTextEditor.selection;
         const currentDocument = window.activeTextEditor.document;
         const range = new Range(start, end);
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
@Ladvien, Thank you as always!
And, It's good readability code.

PR #32, opened 2017-10-29
Added block detection and execute whole block

### LINKED ISSUE #27: `Run line` should load code chunk

Ladvien:
This PR provides functionality which should truly close #27.

If a partial block is selected, for example, 

```
myFunction <- function() {
   foo$Bar <- "FooBar"
...
```
It should parse one line at a time until it finds the termination of the block.  

It should also handle whole and half blocks.

```
myFunction <- function() {
   foo$Bar <- "FooBar"
}

myFunctionTwo <- function() {
   foo$Bar <- "FooBar"
```

And bidirectional block checking:

```
foo$Bar <- "FooBar"
}

myFunctionTwo <- function() {
   foo$Bar <- "FooBar"
}
```

Once I've received some feed back, I'll expand this to include looking at multiline functions.  For example,

```
myFunction(x = 2,
                     y = 3)
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,7 +1,7 @@
 "use strict";
 // The module 'vscode' contains the VS Code extensibility API
 // Import the module and reference it with the alias vscode in your code below
-import { commands, ExtensionContext, languages, Range, window, workspace} from "vscode";
+import { commands, ExtensionContext, languages, Position, Range, window, workspace} from "vscode";
 import { createGitignore } from "./rGitignore";
 import { installLintr, lintr } from "./rLint";
 import { createRTerm, deleteTerminal, rTerm } from "./rTerminal";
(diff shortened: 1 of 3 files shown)


Ikuyadeu:
@Ladvien Thank you for good PR!
I'll review this in this weekend.

Ikuyadeu:
OK, I checked.
Thank you good PR @Ladvien !

PR #33, opened 2017-11-04
add shebang support for R syntax highlight

dongzhuoer:
I copied from the answer of this [question](https://stackoverflow.com/questions/42498117). I'm sorry that I don't know how to create a pull request against the original repository.


Ikuyadeu:
@dongzhuoer Thank you for your PR!
This configure may be `language` option.
I think this should be written in https://github.com/Ikuyadeu/vscode-R/blob/7783dca9bce6b9411e246f21cd8a402193f41165/package.json#L60

By the way https://help.github.com/articles/creating-a-pull-request/ and https://guides.github.com/activities/forking/ are useful documents.

dongzhuoer:
Thanks for your guide.

dongzhuoer:
There is something wrong with the line ending when editing in GitHub GUI. Finally I clone to my computer and fix it. That's why there are so many commits.

Ikuyadeu:
OK, Thank you for fixed, @dongzhuoer !

PR #58, opened 2018-03-20
Add package dev commands

jacob-long:
I have no experience with VS Code extension development, so my apologies if I'm making any mistakes. This seems to work on my machine though.

The main thing I'd been missing compared to RStudio was shortcuts for package development. I think I'd like to add keyboard shortcuts, but I think I can do that in my user keymapping and just put these things in the command palette for everyone else.

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,31 @@
         "title": "Preview Environment",
         "category": "R",
         "command": "r.previewEnvironment"
+      },
+      {
+        "title": "Load All",
+        "category": "R",
+        "command": "r.loadAll"
+      },
+      {
+        "title": "Test Package",
+        "category": "R",
+        "command": "r.test"
+      },
+      {
+        "title": "Install Package",
+        "category": "R",
+        "command": "r.install"
+      },
+      {
+        "title": "Build Package",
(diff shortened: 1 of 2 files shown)


Ikuyadeu:
@jacob-long Thank you for good functions!
I think `r.loadAll` can be `r.package.loadAll` or `r.devtool.loadall`.
But not so important it. I want to know just how did you decide these names.

jacob-long:
I didn't put too much thought into the names. I think it might be smarter to do as you have suggested, using the prefix `r.package` or `r.devtools` to avoid confusing people who aren't developing packages.

jacob-long:
I would also mention that some of these might best be run in a separate R session, as RStudio does it. I think that's maybe best done via VS Code tasks, but I couldn't figure out how to have a package create tasks in that way.

Ikuyadeu:
@jacob-long Thank you for your thinking.
Separating these function is looks good for me.
But it looks hard to implement quickly.

First, I'll merge current good changes, it is useful enough.
If you would implement any fix or improvement, please submit again or ask me.

Thank you again for your submitting!

PR #79, opened 2018-06-23
Add Check Package and Build Vignettes commands

renkun-ken:


Ikuyadeu:
@renkun-ken Thank you for PR!

I forgot to add the TODO on README.md, and your work is great!

However, these functions look easy to run without command and not be called than other packages function.

In my opinion

* Adding the command is good :)
* In the future, this extension should add many important commands (like TODO that you added).
* But many commands without the button or menu make the user confuse important command (like `Run`, `Create Terminal`).

I think, first, we should add menu or button for these commands before merging.

renkun-ken:
Thanks for your suggestions! It is a bit like an initial attempt for me to start working on this extension in hopes of making it somewhat comparable with RStudio (both Desktop and Server) in the long term. 

I'd love to add menu or button for those.

PR #104, opened 2019-04-03
Add runSelectionInActiveTerm command #80 #102

### LINKED ISSUE #80: Only support sending code to local R Interactive
- VSCode Version: 1.24.1
- VSCode-R Version: 0.6.0
- OS Version: macOS 10.13.5

Currently, `Send Code` ignores the existing vscode terminal. An R Interactive is always open.

My case is that I ssh logined a server in the vscode terminal and start an R session there. Therefore, whenever I send code, it is desired that the code be sent to the existing terminal rather than always open a new local R session.

### LINKED ISSUE #102: R Terminal loses purchase of old terminals
Actual it isn't possible to switch between R-Terminals. After creating a new Terminal only the last one is active.

andycraig:
Fix #80

Fix #102

This PR adds a new command `r.runSelectionInActiveTerm`, accessed from the command palette as `R: Run Selection/Line in Active Terminal`. It functions like `r.runSelection`, but sends the selection to the currently active terminal instead of the terminal registered as `rTerm`. 

Can now send code to an R session not started via vscode-R (e.g., an R session started in an SSH session).

Can now run multiple R terminals at the same time and send code to the currently active terminal.

**Screenshot**

PR80

**Notes**

This PR increases the required VSCode engine to 1.29.0, to access `vscode.window.activeTerminal`.

There is no check for whether there is actually an R session running in the active terminal. The code will be sent regardless.

The new command is not bound to any keyboard shortcut by default.

An equivalent command is added for `r.runSelection` only. Other commands that send code (`r.nrow`, `r.length`, `r.head`, `r.thead`, `r.names`, `r.runSource`) still send code to `rTerm` only. If there is demand for further equivalent commands, they could be added in a future PR.


Ikuyadeu:
@andycraig Great work! Your PR information makes me easily understand to contents.
I will check this PR at the weekend.

Ikuyadeu:
@andycraig  OK, I checked. Great work.

I fixed the default runSelection behavior a bit.
Currently, the user needs to call a command.
But it wastes time as same as copy and pastes the source code.

In previous:
1. You make R terminal manually (And the terminal name is R or R interactive)
2. You call command from F1 key

Now:
1. You make R terminal manually (And the terminal name is R or R interactive)
2. You call just `runSelection`. If the terminal name is R or R interactive, `runSelection` command will choose that.

How do you think this?

Ikuyadeu:
Screen Recording 2019-04-06 at 15 33 39
Here is movie, I just put cmd + Enter key 3 times.

README.md
```diff
@@ -20,6 +20,8 @@ Requires [R](https://www.r-project.org/).
 
 ![Create R terminal](images/terminal.png)
 
+* Run code in terminal containing existing R session, for example over SSH (`Run Selection/Line in Active Terminal`)
+
```
Ikuyadeu: SSH support is awesome.
Do you have an example behavior of that?

andycraig:
@Ikuyadeu Thank you for having a look at this PR and streamlining it! I think using `runSelection` for multiple 'R'/'R Interactive' terminals is better than my approach.
 
When using `ssh` from the command line and then running R, the terminal name is 'ssh', not 'R'. So to deal with 'ssh' terminals you would still need to use `runSelectionInActiveTerm` (which a frequent user would probably assign to a keyboard shortcut).

Your changes look good to me. Later today, I can add a movie showing use with SSH to the README, and if it looks okay to you I think it will be ready for merging.

andycraig:
@Ikuyadeu I just noticed that now if there are no open terminals and the user tries to use `runSelection`, they get an error message 'There are no open terminals', but I think the preferred behaviour now would be for an R Interactive terminal to be created. I will add a commit for this.

Ikuyadeu:
@andycraig Thank you for your check.
I am looking forward to that, but please do not too hard.

andycraig:
Thank you, @Ikuyadeu. I wasn't able to spend much more time on it today so I will carry on with it this week.

andycraig:
It seems that creating and showing a terminal does not necessarily set `activeTerminal`, which is making this more complicated than it seems like it should be. From https://github.com/Microsoft/vscode/issues/64390, 'show is an async request which may or may not change the activeTerminal', and it sounds like `onDidChangeActiveTerminal` might be needed.

I will continue to work into this.

andycraig:
I didn't have any success using a listener (e.g., `onDidOpenTerminal`), so I have gone with a workaround approach that seems to solve the problem.

andycraig:
I'm going to make some improvements to that last commit, and then add a movie demonstrating SSH.

README.md
andycraig: @Ikuyadeu I added a movie showing use of R in an SSH session to the README.

andycraig:
@Ikuyadeu If everything looks okay to you, I am happy for this to be merged.

Ikuyadeu:
@andycraig Great works! I will publish a new version today.

Ikuyadeu:
I published a new one, also I added collaboraters (@andycraig and @Ladvien) on the `README`.
From the next version, I skip adding your name on `CHANGELOG` for separate other contributors.

andycraig:
@Ikuyadeu I’m fine with my name being skipped in the `CHANGELOG` since I’m now mentioned in the `README`. Up to you :)

PR #113, opened 2019-07-17
Fix Preview Environment for multi-class objects #111

### LINKED ISSUE #111: Preview Environment R code failed on Sys.time() objects
How to reproduce:

```
time <- Sys.time()
```
Then R: Preview Environment

Reason: class(get("time")) gives a vector of string with 2 elements

Too lazy to submit a pull-request; a quick fix on dist.js works for me.
Fix:
Change
```
write.csv(data.frame(name=ls(),class=sapply(ls(), function(x) {class(get(x))})...
```
to
```
write.csv(data.frame(name=ls(),class=sapply(ls(), function(x) {class(get(x))[1]})...
```

andycraig:
Fixes #111

Solution as suggested by @JimmyZJX - thank you! (I've applied it to `preview.ts` rather than `dist.js`.)

`R: Preview Environment` produces strange results when objects have more than one class.

Example (before PR) - note that `time` appears more than once in preview:

env1

**(If you have)Screenshot**

Example (after PR):

env2

@Ikuyadeu I can build vscode-R successfully on commit 6411a58 but not on commits after that. I get build errors and can't run the unit tests with this version too, but it seems to run and it's only a one-line change so I'm going ahead with the PR.

### DIFF
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -17,7 +17,7 @@ export async function previewEnvironment() {
     const tmpDir = makeTmpDir();
     const pathToTmpCsv = tmpDir + "/environment.csv";
     const envName = "name=ls()";
-    const envClass = "class=sapply(ls(), function(x) {class(get(x))})";
+    const envClass = "class=sapply(ls(), function(x) {class(get(x))[1]})";
     const envOut = "out=sapply(ls(), function(x) {capture.output(str(get(x)), silent = T)[1]})";
     const rWriteCsvCommand = "write.csv(data.frame("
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
@andycraig Great works! Thank you for your contribution!
@JimmyZJX Thank you for your suggestion.

Ikuyadeu:
I will publish this version after the security problem is solved.
It caused `lodash` package in mocha.

andycraig:
@Ikuyadeu Thank you!

PR #125, opened 2019-09-19
Fix send code for newlines on Windows #114

### LINKED ISSUE #114: Newlines when sending code to Radian
See https://github.com/randy3k/radian/issues/102

When using Radian with vscode-R, `R: Run Selection/Line` results in whitespace-only lines before the block of code sent being sent as well. I think this is probably due to vscode-R rather than Radian, although it is interesting that it doesn't happen with the normal R console.

Currently, comment lines are filtered out before being sent to the console; it should be straightforward to filter out whitespace-only lines too.

andycraig:
Fixes #114 

Currently, blank lines that end in '\r\n' (common on Windows) are treated as not being blank, and are sent to the console by `R: Run Selection/Line`. This PR treats '\r' as whitespace, and so these lines are not sent to the console.

**How can I check this pull request (on any OS)?**

Create a file `temp.R` containing:

    # Some blank lines
    
    
    x <- 1

If 'LR' is displayed at the bottom right of the VSCode window, click on it and change the end of line sequence to 'CRLF'.

Use `R: Run Selection/Line`. Only `x <- 1` is sent to the console.

### DIFF
--- a/src/selection.ts
+++ b/src/selection.ts
@@ -44,7 +44,7 @@ export function checkForBlankOrComment(line: string): boolean {
     let index = 0;
     let isWhitespaceOnly = true;
     while (index < line.length) {
-        if (!((line[index] === " ") || (line[index] === "\t"))) {
+        if (!((line[index] === " ") || (line[index] === "\t") || (line[index] === "\r"))) {
             isWhitespaceOnly = false;
             break;
         }


Ikuyadeu:
@andycraig Thank you for your work!
I checked this.

andycraig:
@Ikuyadeu Thank you!

PR #130, opened 2019-09-25
Automatically comment new lines in roxygen sections #124

### LINKED ISSUE #124: Auto create roxygen comment on new line
When I write roxygen comments, it would be convinient if the new line is automatically commented with `#'`.

For example, when I try to add a new line in the middle or at the last line, the new line can be automatically added `#'` to save key strokes.

```r
#' Test function
#' @param x an object
#' @param check a function
######## new line here #########
#' @export
#' @examples
#' test(NULL, is.null)
#' test(0, is.null)
######## new line here #########
test <- function(x, check) {
  if (check(x)) {
    x
  } else {
    stop("Check failed")
  }
}
```

andycraig:
Closes #124 

This PR makes it so that hitting <kbd>enter</kbd> on an roxygen comment line (starting with `#'`) starts the new line with `#'` as well.

It only applies if there is no indentation before `#'`.

roxygen-newline

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,7 +2,7 @@
 // The module 'vscode' contains the VS Code extensibility API
 // Import the module and reference it with the alias vscode in your code below
 import { isNull } from "util";
-import { commands, ExtensionContext, languages, Terminal, window } from "vscode";
+import { commands, ExtensionContext, languages, Terminal, window, IndentAction } from "vscode";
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
@andycraig Thank you for your work!
It looks fine. I will publish a new version this weekend.

andycraig:
@Ikuyadeu Thank you!

Issue #123, opened 2019-09-17
Run/Source commands not friendly to user started R session


renkun-ken:
In my use case, I usually work with remote development via SSH on remote machine and most of the time I also need to start a R session in a tmux window so that the session is preserved.

Therefore, Run or source command should work with the session I start in tmux rather than start a new session. PR #104 partially solves this problem stated in #80, but most run-code related commands should be able to send code to active terminal.

However, it may not make sense to have all commands split into original version and Active Terminal version. I'm wondering if it makes sense to add an option to opt out the behavior to always start a R session if no session started by vscode-R extension is detected?

andycraig:
This sounds good to me, and I think is definitely a better approach than having original/Active Terminal versions for all commands. I would use this option too.

hmassalha:
I would like to add also, 
The current behavior of ctrl+enter is to run selected lines in the current terminal. In case no R session is active, it will start one. 
Would be nice if ctrl+enter can detect if the active terminal is connected to ssh or not, and then launch an R session based on the ssh status (R over the remote machine or the local machine). This is important since making plots is not allowed via the remote R.

andycraig:
Hi @hmassalha, I don’t think there’s a way of checking whether the active terminal is connected to ssh or not.

It may be possible to get plots displaying using port forwarding if that’s available (https://stackoverflow.com/a/3453318) but I haven’t tried that myself.

hmassalha:
Thanks for your note, I am not managing to set the forwarding ports!

renkun-ken:
For plots, I find X11 forwarding is not working well even if the white window shows the experience is not good.

I write a default device in `.Rprofile` (see <https://github.com/renkun-ken/dotfiles/blob/master/.Rprofile#L6-L26>), and it produces png files to `.images` folder in the working directory and always write the last plot to `latest-%d.png`. Each time a new plot is generated, the current `latest-%d.png` files are renamed to time-stamp file name so that history plots are stored. In vscode, you only need to open `latest-1.png` and you can preview the file instantly when the device is closed.

![Kapture 2019-09-30 at 22 10 19](https://user-images.githubusercontent.com/4662568/65886602-3c3ee100-e3cf-11e9-8517-80a2815c8765.gif)

hmassalha:
Thanks for your suggestion. 
I found the following [https://github.com/cloudyr/rmote](url) is a nice solution. You can generate and update the same plot several times, while the browser will keep the previous ones. 

[...]

PR #133, opened 2019-10-06
Add alwaysUseActiveTerminal setting #123

andycraig:
Solves #123 

Currently, only command `R: Run Selection/Line in Active Terminal` works for terminals other than the R Interactive terminal. This PR adds a setting `r.alwaysUseActiveTerminal` (`false` by default). When this setting is enabled, all commands (`R: Run Source`, `R: Preview Dataframe`, `R: Knit Rmd` etc.) are sent to the active terminal. This allows SSH users etc. to access the full functionality of this extension.

**Screenshot**

Before this PR, both `R: Run Selection/Line` and `R: Preview Dataframe` would have opened a new R Interactive terminal.

always-active

**Notes**

I needed to create the function `chooseTerminalAndSendText` outside `activate` in `extension.ts` so that it could be exported for use in `preview.ts`. I decided to put it in `rTerminal.ts`. I moved the associated terminal choosing/sending functions in there, because they couldn't be exported while inside `activate` either. I created a separate commit just for moving these functions, to make it easier to compare the changes.

Command `R: Run Selection/Line in Active Terminal` is probably redundant now, but I left it in there so that this PR doesn't remove any functionality.


Ikuyadeu:
@andycraig Thank you for a lot of work!
I could not find any problem and other tasks that I should do (e.g. Refactoring, Running the tslint, Make a Readme).
So, I will publish this version soon.

andycraig:
@Ikuyadeu Happy to hear that :) Thank you!

PR #136, opened 2019-10-19
Show r.term.option value in settings UI

andycraig:

It is not clear what the current value, expected format or default of the `r.term.option` setting are. This PR fixes that.

**Screenshot**

Before:

options

After:

options-new

Items can be deleted/edited using buttons that appear when the items are hovered over.

### DIFF
--- a/package.json
+++ b/package.json
@@ -336,7 +336,10 @@
             "--no-restore",
             "--no-site-file"
           ],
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
@andycraig Thank you for your good PR!
It is a small, but very important fix.

andycraig:
@Ikuyadeu Thank you!

PR #138, opened 2019-10-24
fix behaviour when workplacefolders is Undefiend

masterhands:
VScode cannot load this extention when we open a single file "hogehoge.R"

Origina code have no error handling while ignorePath is defined.

**(If you have)Screenshot**
No

You'll open a file

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3158,13 +3158,24 @@
       "dev": true
     },
     "https-proxy-agent": {
-      "version": "2.2.1",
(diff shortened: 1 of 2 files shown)


Ikuyadeu:
@masterhands Thank you for your contribution! That problem was the most confusable point for new user.
I'll check this PR this weekend.

Ikuyadeu:
@masterhands I just checked this, and it is work well. Thank you for you fix!

PR #141, opened 2019-11-02
Use word under cursor for previewDataframe, nrow #137

### LINKED ISSUE #137: Use variable under cursor for r.previewDataframe, r.nrow etc.
When there is no selection, it would be good if the following commands operated on the variable the cursor is on:

`r.previewDataframe`, `r.nrow`, `r.length`, `r.head`, `r.thead`, `r.names`

**Details**

Consider the following file:

```
library(ggplot2)
ggplot(iris, aes(Sepal.Width, Sepal.Length)) +
    geom_point()
```

When the cursor is on `iris` in line 2, I think it would make sense for `r.previewDataframe` etc. to act on `iris`. But calling `r.previewDataframe` results in the message 'This does not appear to be a dataframe'. Calling `r.nrow` results in in following appearing in the console:

```
> nrow(ggplot(iris, aes(Sepal.Width, Sepal.Length)) +)
Error: unexpected ')' in "nrow(ggplot(iris, aes(Sepal.Width, Sepal.Length)) +)"
> nrow(geom_point())
NULL
```
 
The reason is that the above commands use `getSelection()` to decide what to act on. When there is no selection, `getSelection()` tries to find the 'minimum runnable block of code', which in the above example is lines 2-3.

@MaTo04 Thank you for adding the `r.nrow` family of commands! Does this suggestion sound okay to you? I don't want to break a use case I haven't thought of.

andycraig:
Closes #137 

This PR makes two changes:

1. If there is no selection, the word under the cursor is used for `r.previewDataframe` and the `r.nrow` family of commands (`r.nrow`, `r.length`, `r.head`, `r.thead`, `r.names`);
2. If multiple lines are selected, the `r.nrow` family of commands apply the function(s) to the whole selection, instead of to each line in the selection.

**Before**

Commands don't work when cursor is on `iris`:

word-before

Functions applied to each line in selection:

lines-before

**After**

word-after

lines-after

@MaTo04 Thank you for adding the r.nrow family of commands! Do you think these changes are okay?


andycraig:
I shortened the title to be within the recommended 50 characters for a commit subject.

Ikuyadeu:
@andycraig Great! Thank you for your fix!
After changing the LICENSE, I will publish this version.

andycraig:
@Ikuyadeu Thank you!

PR #149, opened 2019-11-27
Fix bracketed paste on Windows

### LINKED ISSUE #117: Messy indentation in console when sending text to Radian
- VSCode Version: 1.33.1
- VSCode-R Version: 1.1.1
- OS Version: Ubuntu 18.04.2 LTS

See https://github.com/randy3k/radian/issues/99

I have a working fix for this using bracketed paste mode (thanks @randy3k!): https://github.com/andycraig/vscode-R/tree/bracketed-paste

@Ikuyadeu I think there are two options here:
1. Introduce a 'Bracketed paste mode' option, that users can set if they're using Radian.
2. Always use bracketed paste mode. This works for both Radian and the normal R console, but slightly changes the appearance of how code is send to the R console:

Without bracketed paste mode (current version of vscode-R) - note the '+' on each line in the console: 

bracketed_paste_before

With bracketed paste mode - no '+':

bracketed_paste_after

So, functionally they are the same, but always using bracketed paste mode would slightly change the appearance of the R console from its current form.

andycraig:
Fixes #117

Quoting @Hong-Revo from https://github.com/Ikuyadeu/vscode-R/issues/117#issuecomment-536391701

> I think the only indentation issue I'm having is that sometimes, Radian will treat a single multi-line block of submitted text as a group of lines. For example:
> 
> ```
> f <- function(x)
> {
>     if(x)
>         return(42)
>     else
>     {
>         if(TRUE)
>             23
>         else 69
>     }
> }
> ```
> 
> When I submit this, it shows up as:
> 
> image
> 
> Note the "+" in the middle. Also, sometimes radian thinks there is more text coming and I have to press Enter in the terminal to get it to process the submitted code. This is with bracketed paste on.

This PR makes two changes:

1. When bracketed paste mode is set, sends code text all at once instead of line by line
1. When bracketed paste mode is set and OS is Windows, does not attach ANSI control characters, as Windows does not seem to support these in general

1. Use Windows
1. If `r.bracketedPaste` setting not enabled, enable it and restart VSCode
1. Paste this code into a file `temp-R/temp.R`:

```
f <- function(x) {
    if(x)
        return(42)
    else
    {
        if(TRUE)
            23
        else 69
    }
}
```
4. 'Open Folder...' and choose 'temp-R'
5. Start radian console
6. Position cursor on line 1 of file
7. `R: Run Selection/Line in Active Terminal`
8. Observe no `+` in the text sent to the console, and no need to hit <kbd>Enter</kbd> in the console for it to make it process the code

For full testing, this should also be tried in the following configurations:
1. Windows, with `r.bracketedPaste` not enabled, in normal R
1. Mac/Linux, with `r.bracketedPaste` enabled, in radian
1. Mac/Linux, with `r.bracketedPaste` not enabled, in normal R

I've tried these on Linux but I don't have access to a Mac.

**Notes**

This PR means that setting `r.bracketedPaste` on Windows doesn't actually use bracketed paste ANSI control characters. I think it's probably easier to just leave the setting name and description as-is, but open to discussion if anyone thinks it will make things too confusing.

I think this PR is probably fine as is but I would like to do a bit more testing. I will remove 'WIP' after that.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -93,11 +93,12 @@ export async function runTextInTerm(term: Terminal, textArray: string[]) {
         // Surround with ANSI control characters for bracketed paste mode
         textArray[0] = `\x1b[200~${textArray[0]}`;
         textArray[textArray.length - 1] += "\x1b[201~";
-    }
-
-    for (const line of textArray) {
-        await delay(8); // Increase delay if RTerm can't handle speed.
-        term.sendText(line);
+        term.sendText(textArray.join("\n"));
+    } else {
+        for (const line of textArray) {
+            await delay(8); // Increase delay if RTerm can't handle speed.
+            term.sendText(line);
+        }
     }
     setFocus(term);
 }
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -90,9 +90,11 @@ export function runSelectionInTerm(term: Terminal) {
 
 export async function runTextInTerm(term: Terminal, textArray: string[]) {
     if (textArray.length > 1 && config.get<boolean>("bracketedPaste")) {
-        // Surround with ANSI control characters for bracketed paste mode
-        textArray[0] = `\x1b[200~${textArray[0]}`;
-        textArray[textArray.length - 1] += "\x1b[201~";
+        if (process.platform !== "win32") {
+            // Surround with ANSI control characters for bracketed paste mode
+            textArray[0] = `\x1b[200~${textArray[0]}`;
+            textArray[textArray.length - 1] += "\x1b[201~";
+        }
         term.sendText(textArray.join("\n"));
     } else {
         for (const line of textArray) {


Hong-Revo:
Thanks! Is there a way to test this without Node.js etc installed?

andycraig:
@Hong-Revo Not that I know of. I’m hoping to finish testing it this weekend.

andycraig:
@Ikuyadeu I have done some more testing and I think this PR is ready.

Ikuyadeu:
@andycraig Thank you for your continuous work!
I can only check the code maintainability cause my laptop OS.
I just believe this function will work well.

For future maintenance work, I have one small question.
How did you decide `delay` function arguments? Currently `preview.ts` and `rTerminal.ts` have a that.
* 350
* 50
* 200
* 8
If you have any references for them, can you share it with me and name these values? After checking, I will merge this PR.

If it is based on your test experience,  OK! I merge now.

andycraig:
@Ikuyadeu Great, thank you!

@Ladvien I think you chose the delay amounts, perhaps you can answer @Ikuyadeu's question about them? Thank you in advance!

Ikuyadeu:
OK, I will merge this.
If that maintainability future work makes confusing, I will refactor them.

andycraig:
@Ikuyadeu Great, thank you!

randy3k:
I don’t know much about vscode terminal but radian indeed supports bracketed paste mode in Windows.

andycraig:
@randy3k I’m sure that’s true but something in the combination of Windows + VSCode + bracketed paste mode + radian doesn’t work. The VSCode Julia extension uses the same solution (bracketed paste mode on Mac/Linux, no bracketed paste mode on Windows) so I’m pretty sure the problem is on the Windows + VSCode side.

PR #157, opened 2019-12-23
Use empty order when creating DataTables in getTableHtml

### LINKED ISSUE #156: View data frame default order by first column
With `r.sessionWatcher` enabled,

```r
View(mtcars)
```

results in a table sorted by the first column (rownames) by default, which is not desired.

Instead, it should not sort by any column in default.

renkun-ken:
Closes #156.

According to https://datatables.net/examples/basic_init/table_sorting.html, explicitly use `order: []` to disable any initial ordering behavior.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -223,7 +223,8 @@ async function getTableHtml(webview: Webview, file: string) {
         data: data.data,
         columns: data.columns,
         paging: false,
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
Thank you @renkun-ken 
LGTM

PR #177, opened 2020-01-08
Fix session watcher init.R path on Windows

### LINKED ISSUE #176: Can't find init.R on Windows for session watcher
[Here](https://github.com/Ikuyadeu/vscode-R/issues/143#issuecomment-565773616) I wrote that on Windows `Sys.getenv("HOMEPATH")` returns `C:\Users\[USERNAME]`, but when I tried it just now it returned `\\Users\\[USERNAME]` (without the drive letter). So, when trying to use the session watcher on Windows, I get this error:

```
Error in file(filename, "r", encoding = encoding) :
  cannot open the connection
In addition: Warning message:
In file(filename, "r", encoding = encoding) :
  cannot open file '\Users\[USERNAME]/.vscode-R/init.R': No such file or directory
```

Sorry about this, I should have verified that it definitely worked.

Based on https://cran.r-project.org/bin/windows/base/rw-FAQ.html#What-are-HOME-and-working-directories_003f it seems that it should be:

```
source(file.path(if (.Platform$OS.type == "windows") file.path(Sys.getenv("HOMEDRIVE"), Sys.getenv("HOMEPATH")) else Sys.getenv("HOME"), ".vscode-R", "init.R"))
```

andycraig:
Fixes #176 

Updates `README` to give path for `init.R` that is correct on Windows. (It was already correct on Linux/Mac.)

I have tested that this works on Windows and Linux.

@jacob-long I think you are a Windows user too? If you would be able to confirm that this works for you too that would be great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ An opt-in experimental R session watcher is implemented to support the following
 To enable this feature, turn on `r.sessionWatcher` and append the following code to your `.Rprofile` (in your home directory):
 
 ```r
-source(file.path(Sys.getenv(if (.Platform$OS.type == "windows") "HOMEPATH" else "HOME"), ".vscode-R", "init.R"))
+source(file.path(if (.Platform$OS.type == "windows") file.path(Sys.getenv("HOMEDRIVE"), Sys.getenv("HOMEPATH")) else Sys.getenv("HOME"), ".vscode-R", "init.R"))
 ```
 
 This script writes the metadata of symbols in the global environment and plot file to `${workspaceFolder}/.vscode/vscode-R/PID` where `PID` is the R process ID. It also captures user input and append command lines to `${workspaceFolder}/.vscode/vscode-R/response.log`, which enables the communication between vscode-R and a live R sesson.


Ikuyadeu:
@andycraig Great! Thank you for applying for Windows!

renkun-ken:
@andycraig I'm playing around with a Windows 10 in virtualbox. I notice there's an environment variable `USERPROFILE` which happens to be the user home directory. Not sure if we can use this?

andycraig:
@renkun-ken Good find! I tried `USERPROFILE` it does indeed point to the right location on my Windows PC. But I did some more reading and apparently `USERPROFILE` and `HOMEDRIVE`/`HOMEPATH` don't necessarily point to the same place: https://stackoverflow.com/a/8797855

From [`os.homedir()` docs](https://nodejs.org/api/os.html#os_os_homedir): 'On Windows, it uses the `USERPROFILE` environment variable if defined. Otherwise it uses the path to the profile directory of the current user.'

A couple more options:
1. Rather than having users copy code from `README` into `.Rprofile`, provide a vscode-R command that writes the output of `os.homedir()` to `.Rprofile`. That would still fail if the user's environment variables changed after that, though.
2. Look at the source code of `os.homedir()` and try to duplicate that logic in the code to go in `.Rprofile`.

andycraig:
@renkun-ken For now, using `USERPROFILE` rather than `HOMEDRIVE`/`HOMEPATH` would probably be better.

Issue #179, opened 2020-01-08
Unable to attach terminal with Google Drive (Drive File Stream)


Koohoko:
I am using Windows 10 OS, and when I tried to use the session watcher, the R terminal cannot be attached. (neither auto attach or manually `getOption('vscodeR')$attach()`)

I observed `response.log` was changing over time (as below), but obviously there was no attached R session or any expected window pop-up for dataview.

```

{"time":"2020-01-08 18:10:23","pid":7220,"command":"attach"} 
{"time":"2020-01-08 18:10:26","pid":7220,"command":"attach"} 
{"time":"2020-01-08 18:10:44","pid":7220,"command":"dataview","source":"table","type":"json","title":"mtcars","file":"C:\\Users\\kooho\\AppData\\Local\\Temp\\RtmpOKdp0S\\file1c3430431451.json"} 
{"time":"2020-01-08 18:10:49","pid":7220,"command":"attach"} 
{"time":"2020-01-08 18:15:47","pid":3172,"command":"attach"} 
{"time":"2020-01-08 18:19:31","pid":5868,"command":"attach"} 
{"time":"2020-01-08 18:19:34","pid":5868,"command":"attach"} 
```

andycraig:
@Koohoko Good timing, we just updated the README today with a fix for Windows! Could you try changing the line in your `.Rprofile` to the below and see if that fixes it?

```
source(file.path(if (.Platform$OS.type == "windows") file.path(Sys.getenv("HOMEDRIVE"), Sys.getenv("HOMEPATH")) else Sys.getenv("HOME"), ".vscode-R", "init.R"))
```

Thank you.

renkun-ken:
@Koohoko Thanks for reporting!

It looks like `init.R` is correctly sourced in your `.Rprofile` but vscode-R does not manage to watch the `response.log` you refer to. The following cases may lead to such problem:

1. You don't launch R session with the vscode-R extension activated. Open an R file to activate the extension. If `r.sessionWatcher` is enabled, you'll see a status item `R: (not attached)`.
[...]

PR #184, opened 2020-01-10
Improve session watcher initialization

### LINKED ISSUE #196: More robust session watcher initialization
We could make session watcher script initialization more robust with the following applied:

* `init.R` detects if environment variable `TERM_PROGRAM=vscode`. Only in VSCode it is effective.
* `~/.Rprofile` only sources `init.R` when `~/.vscode-R/init.R` exists. Disabling `r.sessionWatcher` removes `~/.vscode-R/init.R` so that user does not have to comment out or remove the `source()` line manually.

The only drawback of detecting `TERM_PROGRAM=vscode` is that if R is started in a pre-existing environment inside a terminal multiplexer (e.g. `tmux` and `screen`) whose server program is not started inside VSCode terminal, it won't get a `TERM_PROGRAM=vscode` environment variable. But user could always set this environment variable to force it.

This is related to https://github.com/Ikuyadeu/vscode-R/issues/190, https://github.com/Ikuyadeu/vscode-R/issues/195, https://github.com/randy3k/radian/issues/139. When user tries to disable session watcher by setting `"r.sessionWatcher": false` but leaves `source(...)` line in `~/.Rprofile` or any other startup script, the use of all R sessions will be broken.

renkun-ken:
Closes #196.
Related to #177, #179.

In this PR:

* Use `USERPROFILE` for Windows according to https://github.com/nodejs/node/pull/1791: https://github.com/nodejs/node/pull/1791/files#diff-d9bb084c9b6137788ddd0314b2e4df6dR11.
* Clarify when a local `.Rprofile` exists.
* Use `R_PROFILE_USER` as @randy3k suggests so that there's no need for manual configuration if users starts R session using `R: Create R Terminal` command.
* `init.R` checks `TERM_PROGRAM=vscode` so  that it is only effective in VSCode terminal.
* Update `README.md` to cover basic usage, advanced usage and how to disable session watcher.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -1,8 +1,11 @@
 "use strict";
 
+import os = require("os");
+import path = require("path");
+
 import { pathExists } from "fs-extra";
 import { isDeepStrictEqual } from "util";
-import { commands, Terminal, window } from "vscode";
+import { commands, Terminal, window, TerminalOptions } from "vscode";
 
 import { getSelection } from "./selection";
 import { removeSessionFiles } from "./session";
@@ -18,7 +21,17 @@ export function createRTerm(preserveshow?: boolean): boolean {
         const termOpt: string[] = config.get("rterm.option");
         pathExists(termPath, (err, exists) => {
             if (exists) {
-                rTerm = window.createTerminal(termName, termPath, termOpt);
+                let termOptions: TerminalOptions = {
+                    name: termName,
+                    shellPath: termPath,
+                    shellArgs: termOpt
+                };
+                if (config.get("sessionWatcher")) {
+                    termOptions.env = {
+                        R_PROFILE_USER: path.join(os.homedir(), ".vscode-R", ".Rprofile")
+                    };
+                }
+                rTerm = window.createTerminal(termOptions);
                 rTerm.show(preserveshow);
                 return true;
             }
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -28,6 +28,7 @@ export function createRTerm(preserveshow?: boolean): boolean {
                 };
                 if (config.get("sessionWatcher")) {
                     termOptions.env = {
+                        R_PROFILE_USER_OLD: process.env.R_PROFILE_USER,
                         R_PROFILE_USER: path.join(os.homedir(), ".vscode-R", ".Rprofile")
                     };
                 }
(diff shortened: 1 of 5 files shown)


randy3k:
Alternatively, one could make use of the env variable `R_PROFILE_USER` that points to a vscode generated `.Rprofile`. For example, vscode-R could generate `~/.vscode-R/.Rprofile` and set `R_PROFILE_USER` to it in the integrated terminal.

```r
# ~/.vscode-R/.Rprofile

if (file.exists(".Rprofile")) {
	source(".Rprofile")
} else if (file.exists(normalizePath("~/.Rprofile"))) {
	source("~/.Rprofile")
}

if (<init.R has not been execuated>) {
	source(file.path(Sys.getenv(if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"), ".vscode-R", "init.R"))
}
```

This approach would require zero configuration for R sessions launched by vscode-R. Only if a user wants to use session watcher in external R processes, they would have to edit the `.Rprofile`.

renkun-ken:
Thanks @randy3k for your great suggestion. With this PR, user won't have to do any manual configuration if user only starts R session via `R: Create R Terminal` command.

Now, only users who want to use R session watcher with self-managed R session will have to do some extra configuration.

src/rTerminal.ts
```diff
@@ -18,7 +21,17 @@ export function createRTerm(preserveshow?: boolean): boolean {
...
             if (exists) {
-                rTerm = window.createTerminal(termName, termPath, termOpt);
+                let termOptions: TerminalOptions = {
+                    name: termName,
+                    shellPath: termPath,
+                    shellArgs: termOpt
+                };
+                if (config.get("sessionWatcher")) {
+                    termOptions.env = {
+                        R_PROFILE_USER: path.join(os.homedir(), ".vscode-R", ".Rprofile")
```
randy3k: You might want to check if the user has defined a custom `R_PROFILE_USER`.
If it is not empty, you might need to define `R_PROFILE_USER_OLD` and source `R_PROFILE_USER_OLD` instead of `~/.Rprofile`

renkun-ken: Yes, we should respect existing `R_PROFILE_USER`.

andycraig:
@renkun-ken This PR is now conflicted with `master`. Would you be able to resolve that and then I'll merge it? Thanks!

renkun-ken:
@andycraig Conflicts resolved. Thanks!

andycraig:
@renkun-ken Great, thanks again!

PR #203, opened 2020-01-26
WIP Add runCodeWithPath

### LINKED ISSUE #183: Run code using current file name
`runSource` and `knitRmd` wrap the current file’s path in some R code and send that code to the terminal.

We could generalise this to a function `wrapPath` (or something) that has as parameters the R function name and any additional parameters to pass to that R function. Both `runSource` and `knitRmd` could be refactored to use `wrapPath`.

The key benefit of this is that then users could add their own similar commands in their `settings.json` file, without needing vscode-R itself to be changed.

andycraig:
Closes #183 

Adds a command `r.runCodeWithPath` that can be used in `keybindings.json` to run code containing the current editor file's path.

This PR includes refactoring for `runSource` and `knitRmd` commands but does not affect their behaviour.

1. In `keybindings.json`, add:

    ```
    [
      {
        "key": "ctrl+alt+shift+1",
        "command": "r.runCodeWithPath",
        "when": "editorTextFocus",
        "args": [ "rmarkdown::run(", ")" ]
      }
    ]
    ```

2. Create an R Markdown file `temp.Rmd`:

```
    ---
    title: temp
    ---
 
    ```{r}
    iris
    ```
```

3. Open an R console
4. Make sure R extension is active, by, e.g., running `Run Selection/Line`
4. Press <kbd>Ctrl+Alt+Shift+1</kbd>. The following command will be sent to the R console 

```
rmarkdown::run("PATH/temp.Rmd")
```

**Note**

The `r.runCodeWithPath` command is only for use in `keybindings.json`. If the user runs the command `Run Code with Path` from the command palette, it fails with the vague message "Running the contributed command: 'r.runCodeWithPath' failed". I tried adding a check in `runCodeWithPath` that gives a sensible error message if `args.length === 0`, but the same message still appeared. The failure might occur before `runCodeWithPath` is actually called.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,21 @@ Requires [R](https://www.r-project.org/).
 
 * Package development short cut (`Load All`, `Test Package`, `Install Package`, `Build Package` and `Document`)
 
+* Commands for use in `keybindings.json`:
+
+    * `r.runCodeWithPath`: Wraps current file path in `args`. Example `keybindings.json`:
+
+        ```
+        [
+            {
+                "key": "ctrl+alt+shift+1;"",
+                "command": "r.runCodeWithPath",
+                "when": "editorTextFocus",
+                "args": [ "rmarkdown::run(", ")" ]
+            }
+        ]
+        ```
+
 ## Requirements
 
 * R base from <https://www.r-project.org/>
--- a/package.json
+++ b/package.json
@@ -242,6 +242,11 @@
         "title": "Show Plot History",
         "category": "R",
         "command": "r.showPlotHistory"
+      },
+      {
+        "title": "Run Code with Path",
+        "category": "R",
+        "command": "r.runCodeWithPath"
       }
     ],
     "keybindings": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -36,34 +36,20 @@ export function activate(context: ExtensionContext) {
     // Now provide the implementation of the command with  registerCommand
     // The commandId parameter must match the command field in package.json
 
-    function runSource(echo: boolean)  {
+    function runCodeWithPath(args: any[]) {
         const wad = window.activeTextEditor.document;
         wad.save();
-        let rPath: string = ToRStringLiteral(wad.fileName, '"');
-        let encodingParam = config.get<string>("source.encoding");
-        encodingParam = `encoding = "${encodingParam}"`;
-        rPath = [rPath, encodingParam].join(", ");
-        if (echo) {
-            rPath = [rPath, "echo = TRUE"].join(", ");
-        }
-        chooseTerminalAndSendText(`source(${rPath})`);
+        const rPath: string = ToRStringLiteral(wad.fileName, '"');
+        chooseTerminalAndSendText(`${args[0]}${rPath}${args[1]}`);
     }
 
-    function knitRmd(echo: boolean, outputFormat: string)  {
-        const wad: TextDocument = window.activeTextEditor.document;
-        wad.save();
-        let rPath = ToRStringLiteral(wad.fileName, '"');
+    function runSource(echo: boolean) {
         let encodingParam = config.get<string>("source.encoding");
-        encodingParam = `encoding = "${encodingParam}"`;
-        rPath = [rPath, encodingParam].join(", ");
+        encodingParam = `, encoding = "${encodingParam}")`;
         if (echo) {
-            rPath = [rPath, "echo = TRUE"].join(", ");
+            encodingParam = [", echo = TRUE", encodingParam].join("");
         }
-        if (outputFormat === undefined) {
-            chooseTerminalAndSendText(`rmarkdown::render(${rPath})`);
-        } else {
-            chooseTerminalAndSendText(`rmarkdown::render(${rPath}, "${outputFormat}")`);
-        }
+        runCodeWithPath(["source(", encodingParam]);
     }
 
     async function runSelection() {
@@ -117,11 +103,15 @@ export function activate(context: ExtensionContext) {
         commands.registerCommand("r.head", () => runSelectionOrWord(["head"])),
         commands.registerCommand("r.thead", () => runSelectionOrWord(["t", "head"])),
         commands.registerCommand("r.names", () => runSelectionOrWord(["names"])),
+        commands.registerCommand("r.runCodeWithPath", (args: any[]) => runCodeWithPath(args)),
         commands.registerCommand("r.runSource", () => { runSource(false); }),
-        commands.registerCommand("r.knitRmd", () => { knitRmd(false, undefined); }),
-        commands.registerCommand("r.knitRmdToPdf", () => { knitRmd(false, "pdf_document"); }),
-        commands.registerCommand("r.knitRmdToHtml", () => { knitRmd(false, "html_document"); }),
-        commands.registerCommand("r.knitRmdToAll", () => { knitRmd(false, "all"); }),
+        commands.registerCommand("r.knitRmd", () => { runCodeWithPath(["rmarkdown::render(", ")"]); }),
+        commands.registerCommand("r.knitRmdToPdf", () => {
+            runCodeWithPath(["rmarkdown::render(", ", \"pdf_document\")"]); }),
+        commands.registerCommand("r.knitRmdToHtml", () => {
+            runCodeWithPath(["rmarkdown::render(", ", \"html_document\")"]); }),
+        commands.registerCommand("r.knitRmdToAll", () => {
+            runCodeWithPath(["rmarkdown::render(", ", \"all\")"]); }),
         commands.registerCommand("r.createRTerm", createRTerm),
         commands.registerCommand("r.runSourcewithEcho", () => { runSource(true); }),
         commands.registerCommand("r.runSelection", runSelection),


andycraig:
Setting this to WIP to think about whether it’s better for `args` to be an array or a dictionary.

MilesMcBain:
I was having a crack at this too! 

I went with:

```r
    function doCallWithActiveDocumentPath([fn, pathArgName, otherArgs]: [string, string, object]) {
        const wad: TextDocument = window.activeTextEditor.document;
        let rPath = ToRStringLiteral(wad.fileName, '"');
        let rPathArg = pathArgName.concat(" = ", rPath);
        let otherArgsList = Object.entries(otherArgs).
            map(([key, value]) =>
                String.prototype.concat(key.toString(), " = ", value.toString())).
            join(", ");

        let rCall = fn.concat("(", rPathArg, ", ", otherArgsList, ")");
        chooseTerminalAndSendText(rCall);
    }
```

which allowed keybinding config like this test case to work:

```
    {
        "description": "knit to html",
        "key": "ctrl+i",
        "command": "r.doCallWithActiveDocumentPath",
        "when": "editorTextFocus",
        "args": [
            "rmarkdown::render",
            "input",
            {
                "output_format": "rmarkdown::html_document()",
                "output_dir": "\"./\"",
                "clean": "TRUE",
                "output_file": "\"do_call_test.html\""
            }
        ]
    }
```

andycraig:
Thank you for having a think about this too!

I like the aesthetics of your approach. It reminds me of R’s `do.call`, which given the name is presumably what you were thinking of :)

I’d like to have the same API for this and the command that calls R code on words or selections. This makes it a little bit more complicated as that needs to be able to support:

1. Nested functions, e.g., `t(head(SELECTION))`, `f(g(), SELECTION)`
2. Surrounding the selection with

1. Nothing
2. Quotes
3. Backticks
4. Potentially others such as brackets or rlang’s {{ }}

Thoughts?

While I think of it, VSCode supports multiple selections. This MIGHT (might!) be something we want to support here.

MilesMcBain:
Yes! I immediately started thinking along similar lines. 

I stared on a selection or word version but got a a bit hung up on what it would mean if I had to use the `async` keyword. Using the machinery already setup means I have to use it, but it seems not to matter?

Here is a version that is similar to above that works with multiple selections:

```typescript
    async function doCallWithWordOrSelection([fn, objectArgName, otherArgs]: [string, string, object]) {
        const text = getWordOrSelection();
        //text could be a list of objects from multiple selections, build all versions of its arg
        let rObjArgs = text.map(([objectText]) =>
            String.prototype.concat(objectArgName, " = ", objectText))
        let otherArgsList = Object.entries(otherArgs).
            map(([key, value]) =>
                String.prototype.concat(key.toString(), " = ", value.toString())).
            join(", ");

        // make a call for each original selection (if multiple)
        let rCalls = rObjArgs.map(([objArg]) =>
            String.prototype.concat(fn, "(", objArg, ", ", otherArgsList, ")"));
        const callableTerminal = await chooseTerminal();

        runTextInTerm(callableTerminal, rCalls)
    }
```

You raise some good points about where this could go.

I don't think I would miss nested function calls. I could easily drop `thead <- function(x) t(head(x))` into my .Rprofile and call that using my keybinding. 

Another option is use a personal package. I do this kind of thing in Emacs where I call functions from my personal package to drive various automations. `mmmsic::thead()` would not be out of place in there.

But it is nice to be able to provide cool automations out of box for the user. Maybe the vscode R extension needs it's own package to distribute this stuff?

The text types is an interesting one. Unquoted or quoted with `"` would probably cover 90-something% of cases by my estimation.  I wonder if it is worth doing more?

Do you have a compelling usecase?

MilesMcBain:
Ugh. Scratch what I said about that working with multuple selections. It's not working at all atm.

MilesMcBain:
Ooookay. This kind of works:

```typescript
    async function doCallWithWordOrSelection([fn, objectArgName, otherArgs]: [string, string, object]) {
        const text = getWordOrSelection();
        //text could be a list of objects from multiple selections, build all versions of its arg
        let rObjArgs = text.map((objectText) =>
            String.prototype.concat(objectArgName, " = ", objectText))
        let otherArgsList = Object.entries(otherArgs).
            map(([key, value]) =>
                String.prototype.concat(key.toString(), " = ", value.toString())).
            join(", ");

        // make a call for each original selection (if multiple)
        let rCalls = rObjArgs.map((objArg) =>
            String.prototype.concat(fn, "(", objArg, ", ", otherArgsList, ")"));
        const callableTerminal = await chooseTerminal();

        runTextInTerm(callableTerminal, rCalls)
    }
```

I misread the code of `runTextInTerm` which seemed at first to be allowing for multiple selections. I see now it was allowing for multiple lines. So the above will invoke a function call on each line of the supplied text separately. I'm not sure that's expected?

andycraig:
I'm not much help with `async` I'm afraid, my experience with it is limited to this project and for that I just take existing functions as examples and then resolve compiler errors as they appear. The 'tslint' VSCode extension might help with that stuff?

With `t(head())`, there's an existing command `r.thead` in this extension that calls `t(head(SELECTION))`. I'd prefer to retain that, and have it and this new feature use the same machinery.

With quoting etc., I agree that unquoted or quoted with `"` would cover probably cover 90% of the cases.

The version I have in this PR at the moment does allow this stuff (on filepaths, but the same would apply for a command for selections). To take a extreme example, you could create a shortcut that calls `t(head(df["SELECTION"]))` by specifying a keybinding with

```
args = [ "t(head(df[\"", "\"]))" ]
```

This is a very contrived rather than compelling usecase, but since the flexibility is already there it would be nice to retain that.

> I misread the code of `runTextInTerm` which seemed at first to be allowing for multiple selections. I see now it was allowing for multiple lines. So the above will invoke a function call on each line of the supplied text separately. I'm not sure that's expected?

I don't think there's any support in vscode-R at the moment for multiple selections. `runTextInTerm` should just run once on the whole selection.

MilesMcBain:
Okay here's where my thinking is at. A function that could take an function call data object OR an array of such objects that allows config like this to work:

```json
    {
        "description": "knit to selection html",
        "key": "ctrl+shift+j",
        "command": "r.doCallWithWordOrSelection",
        "when": "editorTextFocus",
        "args": {
            "_fn": "rmarkdown::render",
            "_interpolateArg": "input",
            "output_format": "rmarkdown::html_document()",
            "output_dir": "\"./\"",
            "clean": "TRUE",
            "output_file": "\"do_call_test.html\""
        }
    },
    {
        "description": "thead",
        "key": "ctrl+shift+k",
        "command": "r.doCallWithWordOrSelection",
        "when": "editorTextFocus",
        "args": [
            {
                "_fn": "t",
                "_interpolateArg": "x",
            },
            {
                "_fn": "head"
            }
        ]
    }
```

It has special properties denoted with `_` - which is nice because these aren't legal R parameter names. At presenet there are just two such properties:

* `_fn`: mandatory, the function to call
* `_interpolateArg`: optional, the arg to be interpolated with the current word or selection

This means `_interpolateArg` can be used at any and multiple places in the function call chain. Allowing `f(g(), _interpolateArg)` and `f(g(_interpolateArg))` and `f(g(_interpolateArg), _interpolateArg)`.

One current restriction is that when chaining functions, the inner function always becomes the first argument to the outer function - i.e how `%>%` works.

Here is an implementation:

```typescipt
    async function doCallWithWordOrSelection(callInfo: [object] | object) {
        // callInfo is an array of function call objects that have:
        // _fn: a function name - must have
        // _interpolateArg - optional
        // optional other keys which become other args to fn
        // calls are wrapped in reverse order
        const text = getWordOrSelection().join("\n") //this could be split over lines we don't wan't that

        let makeCallData = (callObject) => {
            if (!callObject.hasOwnProperty("_fn") ||
                !(typeof (callObject._fn) == "string")) {
                throw "doCallWithWordOrSelection needs a _fn property stringfor each call info object."
            }
            let fn: string = callObject._fn
            let args: string[] = [];
            if (callObject.hasOwnProperty('_interpolateArg')) {
                args.push(callObject._interpolateArg.concat(" = ", text))
            }
            delete callObject._fn
            delete callObject._interpolateArg
            Object.entries(callObject).
                forEach(([key, value]) =>
                    args.push(String.prototype.concat(key.toString(), " = ", value.toString())))

            return ({ fn: fn, args: args.join(", ") })
        }
        let callInfoArray = isArray(callInfo) ? callInfo : [callInfo]
        let callData = callInfoArray.map(makeCallData);
        let innerMostCallData = callData.shift();
        let innerMostCall = innerMostCallData.fn.concat("(", innerMostCallData.args, ")")
        let nestedCalls = callData.reduce((inner, current) =>
            current.fn.concat("(", inner, ", ", current.args, ")"), innerMostCall)
        const callableTerminal = await chooseTerminal();

        runTextInTerm(callableTerminal, [nestedCalls])
    }
```

MilesMcBain:
One slightly annoying feature is the necessity to always name the arg to be interpolated.Like `x` in the `thead` example. There are some functions I use all the time that I have no idea what the main arg is named.

MilesMcBain:
Quote could be introduced with another special property like `_quote`?

andycraig:
Thanks for having a go at making it more flexible. The ability to have the selection appear multiple times is a nice plus.

In terms of function placement (the inner function becoming the first argument to the second), the values in the dictionary can themselves be dictionaries, so there’s the option of making it explicit like that. That is, the nesting of dictionaries could define the nesting of functions.

BTW the version in the PR does everything here except allow repeated use of the word/selection/filename. Is that the main thing you’re trying to work around? Is it the aesthetics of the way it’s called? I’ve become a bit unclear if we’re brainstorming generally or trying to work around a specific problem.

Whatever API we decide on it will definitely be implementable, so don’t feel you have to have an implementation for each proposal. Although if it’s as much for the fun of coding I totally understand that :)

MilesMcBain:
Oh I'm just brainstorming! You said you like the object/call aesthetic so I was trying to get to something like that that fulfills the requirements. I'm not really sure what's possible with typescript/VSCode so the I end up implementing to see it'll work.

The nested idea is worth exploring although I feel like heavily nested JSON becomes a bit unwieldy to configure.

andycraig:
Ah, excellent :)

> The nested idea is worth exploring although I feel like heavily nested JSON becomes a bit unwieldy to configure.

Yeah, after I mentioned that I did think of [Greenspun’s tenth rule](https://en.m.wikipedia.org/wiki/Greenspun%27s_tenth_rule).

I’m going to be pretty busy for the next few days, I reckon I’ll come back to this this time next week.

andycraig:
Things have quietened down again so I'm back. I had another thought about this. A way to incorporate repeated arguments and multiple selections into the style in this PR might be like this:

```
args = [ "function_a(", 1, ",  function_b(", 2, ", function_c(", 1, ")))" ]
```

In this case, the `1`s would be the contents of the first selection, and the `2` would be the contents of the second selection. (It might be that `1`, `2` etc. need to be strings rather than integers.)

It would be nice to start using these features so I'd like to propose the following:
1. Use this version of this PR
1. For the selection version, use the similar approach I suggested, that is:
    ```
    args = [ "t(head(df[\"", "\"]))" ]
    ```
1. BUT, put `experimental` at the start of the command names so it's clear that the APIs might change. (This PR would need a new commit to make the command `experimentalRunCodeWithPath`.) Then we're making it clear that they're not fixed, and we can explore ideas as we use them over the next couple of months or so.

renkun-ken:
Is there a reason we don't consider something like the following:

```json
[
  {
    "key": "ctrl+alt+shift+1",
    "command": "r.runCodeWithPath",
    "when": "editorTextFocus",
    "args": "rmarkdown::run('$1')"
  }
]
```

```json
[
  {
    "key": "ctrl+alt+shift+1",
    "command": "r.runCodeWithPath",
    "when": "editorTextFocus",
    "args": [ "source('$1')", "foo::bar('$1', '$2')" ]
  }
]
```

andycraig:
@renkun-ken I think that’s an option too, although we’d want to make sure it was possible to use escape characters for when the user wants the string ‘$1’ to appear as-is.

@renkun-ken @MilesMcBain Provided the command is clearly marked as experimental, I’m happy with whichever version the implementer would like to use.

MilesMcBain:
I quite like @renkun-ken's suggestions. I don't think it can clash with R expressions, and it should be possible to quote things anyway you like, e.g.:

  {
    "key": "ctrl+alt+shift+1",
    "command": "r.runCodeWithPath",
    "when": "editorTextFocus",
    "args": [ "source('$1')", "foo::bar($1, \"$2\")" ]
  }
```
vs

```
  {
    "key": "ctrl+alt+shift+1",
    "command": "r.runCodeWithPath",
    "when": "editorTextFocus",
    "args": [ "source('$1')", "foo::bar(`$1`, !!$2)" ]
  }
```
etc since the substitution target could just be `\$[0-9]`.

I've actually never used mutliple selections in any kind of R automation, but my instinct is I would be more likely to do "run this command (or series of commands) over all of these selections" rather than run this series of commands with all these selections interpolated in some kind of bespoke way.

andycraig:
@MilesMcBain Agree, code being run on each selection of a multiple selection would also be a reasonable approach.

MilesMcBain:
Hey @andycraig I'm so keen for this I made a new PR along the lines of @renkun-ken's last suggestion: https://github.com/Ikuyadeu/vscode-R/pull/237

andycraig:
@MilesMcBain Fantastic, thank you very much! I’ll have a look at it this weekend.

andycraig:
@MilesMcBain I said I would look at this this weekend but I don’t think there’s any need to as @renkun-ken has already reviewed it. Thank you both!

Closing in favour of #237

PR #211, opened 2020-01-29
Avoid duplicate handling of response update

### LINKED ISSUE #205: Duplicate handling of response update
When viewing data frames or help files, I am getting double tabs open of the same thing.

renkun-ken:
Closes #205 

Some software may watch a folder and modifies the file on file change in the folder, which causes a new line in `response.log` to be handled multiple times.

One example is OneDrive software under both Windows and macOS. It is constantly syncing a OneDrive folder. If a workspace folder in the OneDrive folder is open in VSCode, any update of `response.log` will result in duplicate handling, e.g. `View(mtcars)` creates multiple tabs of the same data.

To avoid this, we record the number of lines of `response.log` on each scan of the file. We only handle the last line when the number of lines ~~increase~~ changes. As a result, if a new line is appended to `response.log`, and `updateResponse` is called multiple times due to the syncing may cause the file to be somehow updated multiple times, only the first time it is handled will take effect.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -17,6 +17,7 @@ let PID: string;
 let tempDir: string;
 let plotDir: string;
 let resDir: string;
+let responseLineCount: number;
 const sessionDir = path.join(".vscode", "vscode-R");
 
 export function deploySessionWatcher(extensionPath: string) {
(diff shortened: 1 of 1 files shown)


renkun-ken:
It seems we only need the number of lines to change rather than increase to trigger handling response update. When user removes historic lines, the number of lines decreases. And appending new lines should also work then.

Ikuyadeu:
@renkun-ken Thank you for your changes.

From the next pull request, could you wait for the others review(me or @andycraig )?
In this time, I will clean up the whole of the project for future contributors.
The current code style is not fit the tslint style(white-space, console...) and each hunks or functions are too big.

renkun-ken:
@Ikuyadeu Thanks for your reminder!

PR #234, opened 2020-02-18
Change .Platform$GUI to vscode on session start

renkun-ken:
This PR replaces `.Platform$GUI` to `vscode` for user to check. By default, `.Platform$GUI` is `X11` under linux and macOS, which is not the case when `init.R` is sourced and all viewer are handled by vscode-R.

This is also similar with RStudio which replaces `.Platform$GUI` with `RStudio`.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -232,6 +232,10 @@ if (interactive() && Sys.getenv("TERM_PROGRAM") == "vscode") {
         }, "base")
         rebind("View", dataview, "utils")
 
+        platform <- .Platform
+        platform$GUI <- "vscode"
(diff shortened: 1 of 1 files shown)


andycraig:
@renkun-ken This looks correct to me. Is there anything in particular I should test?

renkun-ken:
Just a minor change. The only thing to test is whether `init.R` could be successfully sourced and whether `.Platform$GUI` is successfully changed to `vscode`.

I tested it on macOS and Linux and it works as expected.

andycraig:
Great, I will verify on Windows later today.

Issue #199, opened 2020-01-22
configurable run command in R terminal?


MilesMcBain:
**Is your feature request related to a problem? Please describe.**

The R extension has keybindings for some key R functions like `nrow`, `devtools::test()` etc. It would be nice to be able to configure our own keybindings that run code in the active R terminal.

**Describe the solution you'd like**

A command added to the extension like this:

```typescript
    async function runCommandInRTerm({ text }: { text: string }) {
        const callableTerminal = await chooseTerminal();
        if (callableTerminal === undefined) {
            return;
        }
        callableTerminal.sendText(text);
    }
```

Allows keybinding config like this to work:

```json
[
  {
    "description": "run drake::r_make()",
    "key": "ctrl+;",
    "command": "r.runCommandInRTerm",
    "when": "editorTextFocus",
    "args": {
      "text": "drake::r_make()\n"
    }
  }
]
```

**Describe alternatives you've considered**

Adding commands directly to the extension via PRs - seems unlikely to scale well and create unnecessary maintenance overhead for niche use cases.

andycraig:
Hi @MilesMcBain, there’s a function `chooseTerminalAndSendText` that does this, but at the moment it’s not exposed in the `"commands"` section of `package.json` so it can’t be directly accessed by the user.

I think the necessary change should be only a few lines, so I’ll try to do a PR tonight. (Or feel free to do one earlier if you’d like!)

MilesMcBain:
I tried writing a function that was just a rename of `chooseTerminalAndSendText` but I ran into issues passing `args` to commands in keybindings.

See how my version takes `{ text }: { text: string }`?. This was the only way I could get it to work. Otherwise it would silently fail before executing the command. I'm a total VSCode/Typescript noob though so I could easily believe I was missing something.

andycraig:
Ah, you were ahead of me :) Let me have a look and see if I can figure out the `args` issue.

andycraig:
It seems to work if `chooseTerminalAndSendText` is exposed in `package.json` provided only the value of `args` is specified:

```
[
  {
    "description": "run drake::r_make()",
    "key": "ctrl+;",
    "command": "r.chooseTerminalAndSendText",
    "when": "editorTextFocus && editorLangId == 'r'",
    "args": "drake::r_make()"
  }
]
```

I should be able to do a PR for this on Sunday, but feel free to do one in the meantime if you want.

MilesMcBain:
I'll have a go. 

[...]

PR #237, opened 2020-02-20
Add configurable command runner functions

MilesMcBain:

Addresses feature request in #199 and #183 using @renkun-ken's suggestions from #203.

I am proposing to add three functions that allow the configuration of custom keybindings:

* `runCommand`
* `runCommandWithSelectionOrWord`
* `runCommandWithEditorPath`

These allow keybindings to be configured like this:

```
[
    {
        "description": "run drake::r_make()",
        "key": "ctrl+;",
        "command": "r.runCommand",
        "when": "editorTextFocus",
        "args": "drake::r_make()"
    },
    {
        "description": "load drake target at cursor",
        "key": "ctrl+shift+;",
        "command": "r.runCommandWithSelectionOrWord",
        "when": "editorTextFocus",
        "args": "drake::loadd($1)"
    },
    {
        "description": "knit to html",
        "key": "ctrl+i",
        "command": "r.runCommandWithEditorPath",
        "when": "editorTextFocus",
        "args": "rmarkdown::render($1, output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE, output_file = \"run_command_test.html\")"
    },
    {
        "description": "thead",
        "key": "ctrl+shift+k",
        "command": "r.runCommandWithSelectionOrWord",
        "when": "editorTextFocus",
        "args": "t(head($1))"
    }
]
```

`$1` is replaced with the selection, word at cursor, or file path as appropriate for the command. Perhaps there is a better placeholder since maybe `$1` implies a `$2` is possible when at present it is not. 

One advantage of `$1` is that R names cannot begin with a number, and so the chance of a clash with object$name style expressions is very low.

In #203 we discussed multiple selections and the possibility of running the command over each selection when there are multiple. I have looked into this and I can see they should be available in the `window.activeTextEditor.selections` array. However there is a certain amount of fiddlyness with this since each selection may be either a true selection or a zero-length one (a cursor).

Personally I can't see myself using the multiple selection feature in this way, so I propose waiting on that implementation until a compelling usecase arrives.

I would like to put together some documentation as part of this PR. Should this go in the README?

**(If you have)Screenshot**

Sorry no.

You can try out the configuration examples above. You don't really need the referenced packages like `drake`. You only need to be confident the R code that gets sent to the console looks as it should with the `$1` interpolated with the selection, word, or path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Here are some example entries from `keybindings.json`:
         "key": "ctrl+i",
         "command": "r.runCommandWithEditorPath",
         "when": "editorTextFocus",
-        "args": "rmarkdown::render('$$', output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
+        "args": "rmarkdown::render(\"$$\", output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
     },
 ...
 ]
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ Requires [R](https://www.r-project.org/).
 
 * Package development short cut (`Load All`, `Test Package`, `Install Package`, `Build Package` and `Document`)
 
+* Bind keys to custom R commands using command runner functions (`r.runCommand`, `r.runCommandWithEditorPath`, `r.runCommandWithSelectionOrWord`)
+
 ## Requirements
 
 * R base from <https://www.r-project.org/>
@@ -138,6 +140,43 @@ attach to currently active session.
 
 *The R terminal used in the screenshot is [radian](https://github.com/randy3k/radian) which is cross-platform and
 supports syntax highlighting, auto-completion and many other features.*
+
+## Creating keybindings for R commands
+
+There are 3 ways you can use extension functions to create keybindings that run R commands in the terminal:
+
+1. `r.runCommand` to make a keybinding to run any R expression. 
+2. `r.runCommandWithEditorPath` to create a keybinding for an R expression where the placeholder value `$$` is interpolated with the current file path.
+3. `runCommandWithSelectionOrWord` to create a keybinding for an R expression where `$$` is interpolated with the current selection or the current word the cursor is on.
+
+Here are some example entries from `keybindings.json`:
+
+```
+[
+    {
+        "description": "run drake::r_make()",
+        "key": "ctrl+;",
+        "command": "r.runCommand",
+        "when": "editorTextFocus",
+        "args": "drake::r_make()"
+    },
+    {
+        "description": "load drake target at cursor",
+        "key": "ctrl+shift+;",
+        "command": "r.runCommandWithSelectionOrWord",
+        "when": "editorTextFocus",
+        "args": "drake::loadd($$)"
+    },
+    {
+        "description": "knit to html",
+        "key": "ctrl+i",
+        "command": "r.runCommandWithEditorPath",
+        "when": "editorTextFocus",
+        "args": "rmarkdown::render($$, output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
+    },
+...
+]
+```
 
 ## TODO
 
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Here are some example entries from `keybindings.json`:
         "key": "ctrl+i",
         "command": "r.runCommandWithEditorPath",
         "when": "editorTextFocus",
-        "args": "rmarkdown::render($$, output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
+        "args": "rmarkdown::render('$$', output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
     },
 ...
 ]
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -96,7 +96,7 @@ export function activate(context: ExtensionContext) {
     async function runCommandWithSelectionOrWord(rCommand: string) {
         const text = getWordOrSelection().join("\n");
         const callableTerminal = await chooseTerminal();
-        let call = rCommand.replace("$1", text);
+        let call = rCommand.replace("$$", text);
         runTextInTerm(callableTerminal, [call]);
     }
 
@@ -104,8 +104,8 @@ export function activate(context: ExtensionContext) {
         const wad: TextDocument = window.activeTextEditor.document;
         wad.save();
         const callableTerminal = await chooseTerminal();
-        let rPath = ToRStringLiteral(wad.fileName, '"');
-        let call = rCommand.replace("$1",rPath);
+        let rPath = ToRStringLiteral(wad.fileName, "");
+        let call = rCommand.replace("$$",rPath);
         runTextInTerm(callableTerminal, [call]);
     }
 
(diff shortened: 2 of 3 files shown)


andycraig:
Docs in the README would be great. Thank you!

src/extension.ts
```diff
@@ -93,6 +93,27 @@ export function activate(context: ExtensionContext) {
...
+        const callableTerminal = await chooseTerminal();
+        let call = rCommand.replace("$1", text);
+        runTextInTerm(callableTerminal, [call]);
+    }
+
+    async function runCommandWithEditorPath(rCommand: string) {
+        const wad: TextDocument = window.activeTextEditor.document;
+        wad.save();
+        const callableTerminal = await chooseTerminal();
+        let rPath = ToRStringLiteral(wad.fileName, '"');
```
renkun-ken: I suggest we use empty quote here (e.g. `ToRStringLiteral(wad.fileName, "")`) to be consistent with `runCommandWithSelectionOrWord` which does not quote the selection or word, and this also allows the use case of the following:

```json
{
        "description": "write file",
        "key": "ctrl+shift+w",
        "command": "r.runCommandWithEditorPath",
        "when": "editorTextFocus",
        "args": "writeLines('title: $1', '$1')"
}
```

renkun-ken:
Also `$1` can be used in the following case:

```r
stringi::stri_replace_all_regex("123abc456", "\\d+([a-z]+)\\d+", "$1")
```

We may need an esacpe just like `sprintf("%%s")` returns literal `%s`.

MilesMcBain:
Yeah escaping is not that hard, but I'd prefer to avoid if possible so the as users we don't have to forget about it and get burned by it. I always forget to `{{` in `glue` for example.

What about switching the replacement target to `$$`? It looks a little unconventional but it seems even less likely to cause clashes.

src/extension.ts
MilesMcBain: I was 50-50 on this. Consistency was a strong pull, but it seemed like since it would be the full path with separators and extension, you're very unlikely to use it other than as a string literal. But yeah happy to change.

renkun-ken:
LGTM

MilesMcBain:
Cool just working on the README now.

README.md
```diff
@@ -139,6 +141,43 @@ attach to currently active session.
...
+        "command": "r.runCommandWithSelectionOrWord",
+        "when": "editorTextFocus",
+        "args": "drake::loadd($$)"
+    },
+    {
+        "description": "knit to html",
+        "key": "ctrl+i",
+        "command": "r.runCommandWithEditorPath",
+        "when": "editorTextFocus",
+        "args": "rmarkdown::render($$, output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
```
renkun-ken: Now that the path is not quoted, shoud we use `"args": "rmarkdown::render('$$')"` or `"args": "rmarkdown::render(\"$$\")"`  instead?

MilesMcBain: YES. Sorry!

renkun-ken:
LGTM

renkun-ken:
Another question, how do these commands handle untitled documents? Recently, we've fixed how languageserver handles untitled documents whose URI looks like `untitled:untitled-n`.

For example, what's the behavior of `r.runCommandWithEditorPath`with an untitled document with R language selected?

MilesMcBain:
Good Question!

When called on an untitled docment `r.runCommandWithEditorPath` causes 2 things to happen
1. The unititled name is passed as the path to the command e.g.

```r
rmarkdown::render('Untitled-1', output_format = rmarkdown::html_document(), output_dir = ".", clean = TRUE, output_file = "do_call_test.html")
Error in abs_path(input) : The file 'Untitled-1' does not exist.
In addition: Warning message:
In normalizePath(path, winslash = winslash, mustWork = mustWork) :
  path[1]="Untitled-1": No such file or directory
```

2. A modal dialogue is opened to save the file.

2. arrises because of:
the call to `wad.save()` in the code. It is unfortunate that this is asynchronous. This is where we hit the limits of my VSCode/typescript knowledge... is it possible to make `wad.save()` a synchronous call that blocks on the modal dialogue?

renkun-ken:
I'm not sure if we could check if the document is saved? If so, does it make sense that if the document is untitled, then we notify user the command can only be called with document with valid path?

One thing also comes to my mind that if the document has changes that are not saved, should we save the document first and then run the command? Or otherwise user is actually running command with the unsaved version of the document, which may cause more confusion about what's happening?

MilesMcBain:
I was going to try to force a save of the document if it is untitled but I ran into an issue where I lost a handle to the saved version. This is a known issue apparently: https://github.com/Microsoft/vscode/issues/29156

So it seems refusing to run if the document is unititled is the only sensible option.

If the document is 'dirty' (has unsaved changes) I think automatically saving the document for the user and then running the command makes sense.

MilesMcBain:
Okay so behavoiur for `r.runCommandWithEditorPath` is now:

1. If document is untitled - bail with error.
2. If document has unsaved changes - attempt a save and bail with error if save fails for any reason.

src/extension.ts
```diff
@@ -93,6 +93,42 @@ export function activate(context: ExtensionContext) {
         runTextInTerm(callableTerminal, wrappedText);
     }
 
+    async function runCommandWithSelectionOrWord(rCommand: string) {
+        const text = getWordOrSelection().join("\n");
+        const callableTerminal = await chooseTerminal();
+        let call = rCommand.replace("$$", text);
```
renkun-ken: `.replace()` seems only do it once. If user supply something like 

```json
{
        "description": "run print current word",
        "key": "ctrl+shift+1",
        "command": "r.runCommandWithSelectionOrWord",
        "when": "editorTextFocus",
        "args": "{ print($$); saveRDS($$, 'data/$$.rds'); }"
}
```

Only the first `$$` is replaced but user may probably expect that all `$$` should be replaced?

MilesMcBain: Oh dear. Fixed.

renkun-ken:
I tested all implemented features in this PR and everything works nicely now.

Thanks!

README.md
```diff
@@ -139,6 +141,43 @@ attach to currently active session.
...
+        "command": "r.runCommandWithSelectionOrWord",
+        "when": "editorTextFocus",
+        "args": "drake::loadd($$)"
+    },
+    {
+        "description": "knit to html",
+        "key": "ctrl+i",
+        "command": "r.runCommandWithEditorPath",
+        "when": "editorTextFocus",
+        "args": "rmarkdown::render('$$', output_format = rmarkdown::html_document(), output_dir = \".\", clean = TRUE)"
```
renkun-ken: We may better use both double quotes or both single quotes in one command?

MilesMcBain: I think it looks nicer than escaped double quotes if that's what you mean? It's how I would put it in my config file even if I don't normally use single quotes that often in R.

MilesMcBain: But I suppose if you show some escaping it might make it clear this is necessary... okay from a instructable perspective it makes sense.

renkun-ken: Yes, single quotes look better here and do you mind making `output_dir = \".\"` use single quotes too to look more consistent?

MilesMcBain: Also I just realised what you meant I think... I had a mix of singles AND escaped doubles.

renkun-ken: Yes, user needs to know `\"` is required to use double quotes here.

PR #239, opened 2020-02-22
Check untitled document and save result before running command

renkun-ken:

This PR applies what has been described and implemented in https://github.com/Ikuyadeu/vscode-R/pull/237#issuecomment-589952776 to `runSource()` and `knitRmd()`. If the document is untitled or if the document could not be saved successfully, then error messages are shown and the commands cannot be run.

1. Create an untitled document
2. Switch the language of the document from Plain Text to R
3. Write some random code
4. Run `R: Run Source` and an error message shows for untitled document and the command is not run.
5. Save the document to `test.R` and use `chmod u-w test.R` to make the file read-only.
6. Change the file and make it dirty.
7. Run `R: Run Source` and an error message shows for unsuccessful saving document. Also, a prompt will ask whether to overwrite the read-only file and to attempt making it writable.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -44,7 +44,7 @@ export function activate(context: ExtensionContext) {
             return false;
         }
 
-        const isSaved: boolean = document.isDirty ? await document.save() : true;
+        const isSaved: boolean = document.isDirty ? (await document.save()) : true;
         if (!isSaved) {
             window.showErrorMessage("Cannot run R command: document could not be saved.");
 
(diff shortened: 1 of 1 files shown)


renkun-ken:
@Ikuyadeu This is a follow up of #237 to handle untitled or unsaved cases. Do you think it might be better we review and merge this before a new release?

Ikuyadeu:
@renkun-ken Ok, this PR should be merged and release again soon.

PR #252, opened 2020-03-02
Use Windows registry to find R path

### LINKED ISSUE #244: Get R term path from registry in Windows
Currently, Windows users have to manually set the R term path before creating an R terminal.

But it might be true that many beginner users don't know where the R binary is located or may get confused by `x86`, `x64` stuff.

If Windows users install R with default settings, then the installation will write R path to registry. Therefore, it seems to make sense to try reading from registry first to determine R path, just like what is done in vscode-r-lsp at <https://github.com/REditorSupport/vscode-r-lsp/blob/master/src/extension.ts#L11>. We may borrow that function here to make it easier for Windows users.

renkun-ken:

Closes #244 

If a Windows user installs R with default installation settings, the R path will be added to registry. So that user does not have to specify R executable in this case.

To test it, set `rterm.windows` to empty (now it's empty by default), run `Create R Terminal` and see if it could properly find R path and create a terminal with it.

Another thing to test is that if multiple versions are installed, the registry will contain multiple entries of R paths. Although the registry entry will have multiple subentries in this case, the value of the key should be the last installed version.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -33,7 +33,7 @@ Yes / No
 
 ```json
 // R.exe path for windows
-"r.rterm.windows": "C:\\Program Files\\R\\R-3.4.4\\bin\\x64\\R.exe",
+"r.rterm.windows": "",
 
 // R path for Mac OS X
 "r.rterm.mac": "/usr/local/bin/R",
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -20,7 +20,7 @@ Yes / No
 
 ```json
 // R.exe path for windows
-"r.rterm.windows": "C:\\Program Files\\R\\R-3.4.4\\bin\\x64\\R.exe",
+"r.rterm.windows": "",
 
 // R path for Mac OS X
 "r.rterm.mac": "/usr/local/bin/R",
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@ Requires [R](https://www.r-project.org/).
 
 ## Usage
 
-* For Windows, set config `r.rterm.windows` to your `R.exe` Path like `"C:\\Program Files\\R\\R-3.3.4\\bin\\x64\\R.exe"`;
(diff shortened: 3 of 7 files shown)


renkun-ken:
I tested this on my Windows 10 VM and it works as expected.

image

To test, please download the build at https://github.com/Ikuyadeu/vscode-R/actions/runs/49221093 or see the artifacts in the [details](https://github.com/Ikuyadeu/vscode-R/pull/252/checks?check_run_id=484012051) of main / build (pull_request).

andycraig:
@renkun-ken I tried testing this just now on Windows 10, but no success I'm afraid.

Testing using VSIX file:
```
Cannot find R client.  Please check R path in preferences and reload.
```

Just to check, I also tested by running your `win-find-rpath` branch in debug mode:
```
Activating extension 'Ikuyadeu.r' failed: Cannot find module 'winreg'.
command 'r.createRTerm' not found
```

Not sure if it helps, but this is what my registry looks like:

reg

Full key is like `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\R-core\R\3.4.1`

ksq:
I test with the VSIX at https://github.com/Ikuyadeu/vscode-R/actions/runs/49233584 and it works perfectly on my Windows 10.

My registry looks like 

image

renkun-ken:
@andycraig We use `R`'s `InstallPath` key rather than any sub-entries such as `3.4.4` or `3.6.3`.
Would you like to take a look at the keys `R` has?

andycraig:
@renkun-ken Looks like that's the issue in my case. `InstallPath` seems to be under `3.4.1` rather than `R`:

reg1

reg2

renkun-ken:
@andycraig I see. I'm not sure if newer versions of R will write to both `R` and `R/3.x.y`, or maybe it could also be caused by uninstalling the last installed version.

We are using almost the same code as in vscode-r-lsp and its default value. Therefore we are consistent with it too. If user does not have the R executable path correctly written in `R/InstallPath`, then user should specify path in VSCode settings. Maybe we should make it more clear in the documentation or error message?

andycraig:
@ksq Thank you for testing!

@renkun-ken Yeah, the version of R I have installed on Windows is old so that could well be it.

It worked for @ksq, and for me it displayed the appropriate error message. The README is very clear as well. I am happy with it as-is. Please merge it if you have no additional changes.

Thank you for this PR! It will make it simpler for new users.

PR #268, opened 2020-03-29
Make rebind also work with attached packages

### LINKED ISSUE #263: Some extensions should be installed for the View function?
The view() is a great function. The data frame or help document was not be displayed in the VSCode IDE when the view function was turned on. What kind of extensions should be installed in the VSCode or R. The error message as follow.

_> View(mtcars)
Error in .External2(C_dataviewer, x, title) : unable to start data viewer
In addition: Warning message:
In View(mtcars) : unable to open display_

I use VSCode ssh to my Ubuntu server (ver 16.04). The Microsoft R Open 3.5 was installed on the Ubuntu server.

renkun-ken:

Closes #263 

Previously, if a user attaches `utils` for some reason before calling `source("~/.vscode-R/init.R")`, then `utils::View` is replaced but `View` is still not. As a result, `View` cannot not trigger web view and user has use `utils::View` as a walk-around.

This PR makes `rebind` in `init.R` more robust so that it also works in this case too.

1. Insert `library(utils)` before sourcing `~/.vscode-R/init.R` in `~/.Rprofile`
2. Start a new R session
3. Run `View(mtcars)` and it should work.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -217,10 +217,21 @@ if (interactive() && Sys.getenv("TERM_PROGRAM") == "vscode") {
         }
 
         rebind <- function(sym, value, ns) {
-          ns <- getNamespace(ns)
-          unlockBinding(sym, ns)
-          on.exit(lockBinding(sym, ns))
-          assign(sym, value, envir = ns)
+          if (is.character(ns)) {
+            Recall(sym, value, getNamespace(ns))
+            pkg <- paste0("package:", ns)
+            if (pkg %in% search()) {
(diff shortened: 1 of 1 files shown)


andycraig:
Hi @renkun-ken, I will review these two PRs this weekend.

andycraig:
@renkun-ken I tested this on Windows 10 and it works. If you are happy with it, please merge.

renkun-ken:
Thanks for testing!

PR #274, opened 2020-03-31
Make plot update smarter using magic null dev size

### LINKED ISSUE #271: knitr::knit a document with plot code opens existing session plot file
**Describe the bug**

When `r.sessionWatcher` is on, when user calls `knitr::knit()` with a Rmd document with plot commands, the existing plot file stored by R session watcher will be open after the function completes.

**To Reproduce**
1. Create a Rmd document (e.g. `doc1.Rmd`) with the following content:

````
# hello

Hello, world!

```{r}
plot(rnorm(100))
```
````

2. Call `knitr::knit("doc1.Rmd")`
3. When the function call completes, no plot file appears.
4. Call `plot(rnorm(10))` and the plot file is open. Close this file.
5. Call `knitr::knit("doc1.Rmd")` again and the previous plot file is open too.

### PR DESCRIPTION
**Pull request by renkun-ken**

Closes #271 

This PR makes plot update smarter using a magic null dev size so that we can check if the current device is the `NULL` device *we* create as the default device to output graphics. I don't see if there's any better way to help us identify the device we create since user has very limited capability to set metadata and limited access to the information of a device.

Now plot file is open only when graphics functions are called with the `NULL` pdf device we create. So `knitr::knit()` no longer triggers opening existing plot file, and user created device such `png` or other `pdf` will not set `plot_updated = TRUE` so VSCode won't copy the graphics from the current device.

**(If you have)Screenshot**

* Basic plot

```r
plot(rnorm(100))
abline(h = 0, col = "red")
# close the plot file
png("test.png") # should not open plot file
plot(rnorm(100)) # should not open plot file
dev.off() # should not open plot file
abline(v = 0, col = "blue") # should open the plot file
```

* `knitr::knit()`

1. Create a Rmd document (e.g. `doc1.Rmd`) with the following content:

````
# hello

Hello, world!

```{r}
plot(rnorm(100))
```
````

2. Call `knitr::knit("doc1.Rmd")`
3. When the function call completes, no plot file appears.
4. Call `plot(rnorm(10))` and the plot file is open. Close this file.
5. Call `knitr::knit("doc1.Rmd")` again and the previous plot file is not open.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -18,17 +18,25 @@ if (interactive() && Sys.getenv("TERM_PROGRAM") == "vscode") {
         plot_file <- file.path(dir_session, "plot.png")
         plot_history_file <- NULL
         plot_updated <- FALSE
+        null_dev_id <- c(pdf = 2L)
+        null_dev_size <- c(7 + pi, 7 + pi)
+
+        check_null_dev <- function() {
+          identical(dev.cur(), null_dev_id) && identical(dev.size(), null_dev_size)
+        }
 
         new_plot <- function() {
-          plot_history_file <<- file.path(dir_plot_history,
-            format(Sys.time(), "%Y%m%d-%H%M%OS6.png"))
-          plot_updated <<- TRUE
+          if (check_null_dev()) {
+            plot_history_file <<- file.path(dir_plot_history,
+              format(Sys.time(), "%Y%m%d-%H%M%OS6.png"))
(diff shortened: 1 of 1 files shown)


andycraig:
@renkun-ken I tested this on Windows 10 and it works. If you are happy with it, please merge.

renkun-ken:
Thanks for testing!

PR #278, opened 2020-04-09
Fix R code formatting according to linting results

renkun-ken:
This PR fixes the `init.R` code formatting according to linting results and also adds `.lintr` so that unneeded linters are disabled.

### DIFF
--- a/R/.lintr
+++ b/R/.lintr
@@ -1,0 +1,1 @@
+linter: with_defaults(cyclocomp_linter = NULL)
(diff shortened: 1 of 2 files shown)


Ikuyadeu:
@renkun-ken Good! Thank you for your formatting.

renkun-ken:
Just notice that the lintr does not seem to respect the newly added `R/.lintr` file while linting `./R` directory. I'll figure it out later.

PR #282, opened 2020-04-15
Improve .Rprofile

renkun-ken:
This PR reworks `.Rprofile` used to create R terminal so that it could mimic creating R session by user in terminal:

The rule is:

1. If `R_PROFILE_USER` is non-empty then source it if exists. (if the file does not exist, do nothing and it does not continue to source other profiles)
2. If `R_PROFILE_USER` is empty,
2.1 If project profile `.Rprofile` exists, source it, otherwise
2.2 If user profile `~/.Rprofile` exists, source it.

### DIFF
--- a/R/.Rprofile
+++ b/R/.Rprofile
@@ -10,8 +10,7 @@ local({
 
   try_source(Sys.getenv("R_PROFILE_USER_OLD")) ||
     try_source(".Rprofile") ||
-    try_source(file.path("~", ".Rprofile")) ||
-    try_source(file.path(R.home(), "etc", "Rprofile.site"))
+    try_source(file.path("~", ".Rprofile"))
 })
 
 if (is.null(getOption("vscodeR"))) {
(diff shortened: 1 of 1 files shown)


andycraig:
This looks to me like a refactoring/tidying with no behaviour change. If so, I'm happy for it to be merged (although I haven't tested it). If there is a behaviour change that I've missed, could you let me know? Thank you.

renkun-ken:
@andycraig There's indeed a subtle behavior change to be consistent with starting R in terminal:

If user defined `R_PROFILE_USER` is non-empty but the file does not exist, R will silently ignore this but does not continue to source any other `.Rprofile`.

andycraig:
@renkun-ken Thank you! I see it now. That is subtle - well done for identifying that behaviour in R! I am happy with this. Please merge it when you are ready.

renkun-ken:
Thanks for your reviewing!

PR #290, opened 2020-04-25
Add command Run from Beginning to Line

### LINKED ISSUE #288: Keyboard shortcut for "run from line 1 to cursor line"
No

I'd like to be able to use the shortcut **Ctrl+Alt+B** to send the code from line 1 to the line where the cursor is, to the console/terminal. The cursor position should remain unchanged. (The latter shortcut exists in RStudio.)

The alternative is to select the code manually using **Shift+PageUp** followed by **Ctrl+Enter** to send the selection. This is inconvenient because then I need to go back and find the line where I was using **PageDown** and arrow keys.

**Additional context**
For example, if in the attached screenshot the cursor is on line 8 column 22, then after pressing **Ctrl+Alt+B** I would like to see lines 1-8 inclusive executed, while the cursor remains on line 8 column 22.

Thanks you very much for any help & thanks for building this excellent extension!

Screenshot from 2020-04-22 16-11-53

andycraig:
Closes #288 

Adds command `Run from Beginning to Line`. Binds it to <kbd>Ctrl+Alt+B</kbd> (not used by VS Code).

If there is a selection, it uses the line at the END of the selection (like RStudio).

Create `temp.R` with text

```
x <- 1
y <- 2
z <- 3
```

Place cursor anywhere on line `y <- 2`. Run command `Run from Beginning to Line`. Verify that text sent to console is

```
x <- 1
y <- 2
```

Do same thing with selection including both line `x <- 1` and `x <- 2`. Result should be same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -328,6 +328,11 @@
         "title": "Run Command in Terminal",
         "category": "R",
         "command": "r.runCommand"
+      },
+      {
+        "title": "Run from Beginning to Line",
+        "category": "R",
+        "command": "r.runFromBeginningToLine"
       }
     ],
     "keybindings": [
@@ -389,6 +394,12 @@
         "command": "r.runSourcewithEcho",
         "key": "shift+Ctrl+enter",
         "mac": "shift+cmd+enter",
+        "when": "editorTextFocus && editorLangId == 'r'"
+      },
+      {
+        "command": "r.runFromBeginningToLine",
+        "key": "Ctrl+alt+b",
+        "mac": "cmd+alt+b",
         "when": "editorTextFocus && editorLangId == 'r'"
       }
     ],
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -137,6 +137,19 @@ export function activate(context: ExtensionContext) {
         runTextInTerm(callableTerminal, rCommand);
     }
 
+    async function runFromBeginningToLine() {
+        const callableTerminal = await chooseTerminal(true);
+        if (callableTerminal === undefined) {
+            return;
+        }
+        const endLine = window.activeTextEditor.selection.end.line;
+        const charactersOnLine = window.activeTextEditor.document.lineAt(endLine).text.length;
+        const endPos = new Position(endLine, charactersOnLine);
+        const range = new Range(new Position(0, 0), endPos);
+        const text = window.activeTextEditor.document.getText(range);
+        runTextInTerm(callableTerminal, text);
+    }
+
     languages.registerCompletionItemProvider('r', {
         provideCompletionItems(document: TextDocument, position: Position) {
             if (document.lineAt(position).text
@@ -171,6 +184,7 @@ export function activate(context: ExtensionContext) {
         commands.registerCommand('r.runSourcewithEcho', () => { runSource(true); }),
         commands.registerCommand('r.runSelection', runSelection),
         commands.registerCommand('r.runSelectionInActiveTerm', runSelectionInActiveTerm),
+        commands.registerCommand('r.runFromBeginningToLine', runFromBeginningToLine),
         commands.registerCommand('r.createGitignore', createGitignore),
         commands.registerCommand('r.previewDataframe', previewDataframe),
         commands.registerCommand('r.previewEnvironment', previewEnvironment),


renkun-ken:
I test this with multiple cases where `|` denotes the cursor at which the command is run.

1. This works as expected.

```r
x <- 1
y <- 2|
z <- 3
```

2. The expression is broken:

```r
m <- local({
  x <- rnorm(100)
  y <-| rnorm(100)
})
```

```
> m <- local({ 
      x <- rnorm(100) 
  })                                                                                                                                       

>   y <- rnorm(100)                                                                                                                        

> 
```

2. Not working:

```r
m <- local({
  x <- rnorm(100)
  y <- rnorm(100)
})|
```

```
> m <- local({ 
      x <- rnorm(100) 
  })                                                                                                                                       

>   y <- rnorm(100)                                                                                                                        

> })                                                                                                                                       
Error: unexpected '}' in "}"
```

Looks like it not works with simple expressions but if the cursor is at any location inside a multi-line expression, it does not work correctly.

andycraig:
@renkun-ken Thank you for testing! I've tried to duplicate your examples but I'm getting different results. With example 2 I get

```
> m <- local({
+   x <- rnorm(100)
+   y <- rnorm(100)
+
```

and with the third example I get

```
> m <- local({
+   x <- rnorm(100)
+   y <- rnorm(100)
+ })
```

These are consistent with RStudio and I'd say they're okay. I get the same results (with slight formatting differences) in R and radian, with and without bracketed paste. In your case it looks like there is some autocompletion of `})` happening that is producing strange output. Do you get the same results in R with bracketed paste off?

renkun-ken:
Just noticed that it is caused by the default radian setting `radian.auto_match = TRUE`.

When I turn it off in options, the command works as described but `radian.auto_indentation = TRUE` makes the indentation looks messy.

I guess we can't do much about this but the default `auto_match = TRUE` behavior may make the expression sent to terminal incorrect in the cursor is in the middle of a multi-line expression. Maybe it is also caused by how the input characters are sent to the terminal.

andycraig:
My earlier results were on Windows 10. Just now I tried on Ubuntu. With radian version 0.4.0 I got the same results as before (although the indentation was messed up), but when I updated to radian version 0.5.3 I got the same results as you. My version of radian on Windows is probably a bit old too so I'll try it again later after updating.

andycraig:
Looks like default of `radian.auto_match` used to be `FALSE`.

renkun-ken:
It seems less ideal if the behavior relies on the which terminal is used and whether `radian.auto_match` is turned on.

Do you think it might make more sense if we make it something like `Run from beginning to current expression` so that the text from the beginning of document to end of `extendSelection` is to be sent to terminal.

andycraig:
I tried it just now on Windows with radian 0.5.3, and I still get the behaviour I was originally getting (no completion of brackets). So it looks like the brackets are auto-completed on Linux/Mac but not Windows. Maybe radian uses the presence of bracketed paste ASCII characters to decide whether to autocomplete brackets or not.

> Do you think it might make more sense if we make it something like Run from beginning to current expression so that the text from the beginning of document to end of extendSelection is to be sent to terminal.

I was thinking about this as well. I'm not sure yet. Auto-completion of brackets while typing makes a lot of sense but auto-completing brackets when chunks are sent to the console is a bit surprising to me. I want to make sure it's intended.

andycraig:
Hi @randy3k, we are finding that when we send incomplete bracketed multi-line blocks of code to radian that the brackets are auto-completed. E.g., sending *just the first three lines* of this

```
m <- local({
  x <- rnorm(100)
  y <- rnorm(100)
})
```

results in this output in radian:

```
> m <- local({ 
      x <- rnorm(100) 
  })                                                                                                                                       

>   y <- rnorm(100) 
```

This happens on Linux/Mac but not Windows, so my guess is it only happens when there are bracketed paste ASCII characters. [This comment](https://github.com/randy3k/radian/issues/29#issuecomment-350503671) makes it seem like this might not be the intended behaviour, so I wanted to check with you. Thanks!

randy3k:
what is `incomplete bracketed multi-line blocks of code`? I beleive bracketed code should always be in one piece.

randy3k:
> Just noticed that it is caused by the default radian setting `radian.auto_match = TRUE`.
> 
> When I turn it off in options, the command works as described but `radian.auto_indentation = TRUE` makes the indentation looks messy.
> 
> I guess we can't do much about this but the default `auto_match = TRUE` behavior may make the expression sent to terminal incorrect in the cursor is in the middle of a multi-line expression. Maybe it is also caused by how the input characters are sent to the terminal.

Auto matching is not enabled in bracketed paste mode, it may explain why.

andycraig:
@randy3k Thanks for the quick response!

> what is incomplete bracketed multi-line blocks of code? I beleive bracketed code should always be in one piece.

Sorry, I'm using 'bracketed' in multiple ways. The example is sending this code to radian:

```
m <- local({
  x <- rnorm(100)
  y <- rnorm(100)
```

It is 'incomplete bracketed' in the sense that the `({` on line 1 have no match in the code sent.

The output is 

```
> m <- local({ 
      x <- rnorm(100) 
  })                                                                                                                                       

>   y <- rnorm(100) 
```

so it looks like radian is auto-matching the `({` on line 1.

> Auto matching is not enabled in bracketed paste mode, it may explain why.

Bracketed paste is on in this example. So, it sounds like radian auto-matching should NOT be happening.

@renkun-ken @randy3k I'm starting to think that the bracketed paste may not be happening correctly in this case. That is, that the bracketed paste ASCII characters may be in slightly the wrong place. I think this is supported by the output being split across two radian prompts, which shouldn't happen with bracketed paste mode. I think there may be a preexisting problem with vscode-R's code for bracketed paste. I will set this to WIP for now.

andycraig:
I have updated this now that #305 is merged.

@renkun-ken Now when I run your example

```r
m <- local({
  x <- rnorm(100)
  y <-| rnorm(100)
})
```

in Linux with radian and bracketed paste I get the expected outcome:

```
r$> m <- local({ 
      x <- rnorm(100) 
      y <- rnorm(100)
```

renkun-ken:
I also test this on macOS and this works well now. Thanks for the good work! Please feel free to merge if nothing to add.

andycraig:
@renkun-ken Thank you for all your multiple rounds of testing on this one! I'll merge now.

PR #301, opened 2020-05-02
Change so setting changes take effect immediately

### LINKED ISSUE #299: Have setting changes apply immediately
When changing settings like `r.bracketedPaste`, a reload is necessary for the new setting to take effect. I'm sure I've seen other extensions for which setting changes apply immediately.

andycraig:
Closes #299 

Currently, changing a vscode-R setting requires a reload of the extension before the setting change takes effect. This PR makes setting changes take effect immediately.

Create a file `temp.R` with content

```
echo "Hello World"
```

Enable setting `R: Always Use Active Terminal`.

Open a terminal. Do NOT start R.

<kbd>Ctrl+Enter</kbd> in `temp.R`.

Observe that `Hello World` is sent to the terminal.

Disable setting `R: Always Use Active Terminal`.

<kbd>Ctrl+Enter</kbd> in `temp.R`.

Observe that a new R session starts in the terminal (with an error because `echo` is not valid R code).

Previously, `R: Always Use Active Terminal` would not have taken effect until the extension was reloaded (e.g., by restarting VS Code).

### DIFF
--- a/package.json
+++ b/package.json
@@ -413,17 +413,17 @@
         "r.rterm.windows": {
           "type": "string",
           "default": "",
-          "description": "R.exe path for windows"
+          "description": "R.exe path for windows."
         },
         "r.rterm.mac": {
           "type": "string",
           "default": "/usr/local/bin/R",
-          "description": "R path for Mac OS X"
+          "description": "R path for Mac OS X."
         },
         "r.rterm.linux": {
           "type": "string",
           "default": "/usr/bin/R",
-          "description": "R path for Linux"
+          "description": "R path for Linux."
         },
         "r.rterm.option": {
           "type": "array",
@@ -431,7 +431,7 @@
             "--no-save",
             "--no-restore"
           ],
-          "description": "R command line options",
+          "description": "R command line options.",
           "items": {
             "type": "string"
           }
@@ -439,7 +439,7 @@
         "r.source.encoding": {
           "type": "string",
           "default": "UTF-8",
-          "description": "An optional encoding to pass to R when executing the file, i.e. 'source(FILE, encoding=ENCODING)'"
+          "description": "An optional encoding to pass to R when executing the file, i.e. 'source(FILE, encoding=ENCODING)'."
         },
         "r.source.focus": {
           "type": "string",
@@ -448,22 +448,22 @@
             "editor",
             "terminal"
           ],
-          "description": "Keeping focus when running"
+          "description": "Keeping focus when running."
         },
         "r.alwaysUseActiveTerminal": {
           "type": "boolean",
           "default": false,
-          "description": "Use active terminal for all commands, rather than creating a new R terminal"
+          "description": "Use active terminal for all commands, rather than creating a new R terminal."
         },
         "r.bracketedPaste": {
           "type": "boolean",
           "default": false,
-          "description": "Use bracketed paste mode when sending code to console. Enable for Radian console"
+          "description": "Use bracketed paste mode when sending code to console. Enable for Radian console."
         },
         "r.sessionWatcher": {
           "type": "boolean",
           "default": false,
-          "description": "Enable R session watcher (experimental). Restart required to take effect"
+          "description": "Enable R session watcher (experimental). Restart required to take effect."
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -463,7 +463,7 @@
         "r.sessionWatcher": {
           "type": "boolean",
           "default": false,
-          "description": "Enable R session watcher (experimental)"
+          "description": "Enable R session watcher (experimental). Restart required to take effect"
         }
       }
     }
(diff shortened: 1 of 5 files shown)


andycraig:
I committed with unsaved changes to `session.ts`. I will force-push the changes to that.

src/extension.ts
```diff
@@ -188,7 +188,7 @@ export function activate(context: ExtensionContext) {
         window.onDidCloseTerminal(deleteTerminal),
     );
 
-    if (config.get<boolean>('sessionWatcher')) {
+    if (workspace.getConfiguration('r').get<boolean>('sessionWatcher')) {
```
renkun-ken, line 191: The session watcher initialization here only occurs when extension is activated. If user enables it for the first time, it won't work immediately.

andycraig: Thanks for checking! You're right. It looks like the session watcher *would* need `onDidChangeConfiguration` in order for the user to activate it mid-session.

> Instead, we could add something like Restart required to take effect to the description of r.sessionWatcher to make it clear.

I think this is a good approach for now. I will add a commit for this.

package.json
```diff
@@ -463,7 +463,7 @@
         "r.sessionWatcher": {
           "type": "boolean",
           "default": false,
-          "description": "Enable R session watcher (experimental)"
+          "description": "Enable R session watcher (experimental). Restart required to take effect"
```
renkun-ken, line 466: Looks good but do we need `.` in the end?

andycraig: All vscode-R's setting descriptions end without a `.` including the multi-sentence ones (although I think I wrote the only other multi-sentence one). Looking through the settings provided by VS Code itself it seems that the standard style is to end all setting descriptions with a `.`. I'm happy to change *all* our setting descriptions to end in `.` for consistency. I'll add a commit to do that.

Ikuyadeu: Currently, we do not have `.`.
 But, other major extensions like Python and ESlint use a `.`.
it is a good time to use popular format.

andycraig: @Ikuyadeu Thanks for the thumbs up on the style change!

renkun-ken: I take a look at vscode settings description and it looks like most have ending `.`, and the rule could be as simple as:

1. If the description is only one sentence, vscode settings descriptions uses ending `.` but some extension do not.
2. If the description is multiple sentences, a ending `.` seems always used.

To be consistent with vscode and most popular extensions, we could always use `.`.

renkun-ken:
Another question is that it seems the `config` variable in `utils.R` (https://github.com/Ikuyadeu/vscode-R/pull/301/files#diff-8cd4968b81985f0efc2053eabc37db6dR7) seems no longer used any more. Should we clean it up and the imports?

andycraig:
@renkun-ken Good catch! Will do.

andycraig:
Hmm, we could retain the intention of `config` by changing it to a function instead of a variable.

src/util.ts
```diff
@@ -4,11 +4,10 @@ import { existsSync } from 'fs-extra';
 import path = require('path');
 import { window, workspace } from 'vscode';
 import winreg = require('winreg');
-export let config = workspace.getConfiguration('r');
```
Ikuyadeu, line 7: how is just change to
```ts
export const config = (function() {
    return workspace.getConfiguration('r');   
});
```
It will reduce whole of changes.

andycraig: @Ikuyadeu Exactly, that's what I was thinking of.

andycraig: @Ikuyadeu 
```typescript
export const config = (function() {
    return workspace.getConfiguration('r');   
});
```
is better than
```typescript
export function config() {
    return workspace.getConfiguration('r');
}
```
?

renkun-ken: Yes this looks better.

Ikuyadeu: I agree with you.

andycraig: Okay, I will change to this.

```typescript
export function config() {
    return workspace.getConfiguration('r');
}
```

The commits are becoming a bit messy. I would like to tidy up using REBASE and FORCE PUSH. @Ikuyadeu @renkun-ken Do you approve rebase and force push here?

Ikuyadeu: In this case, `Squash and Merge` is better, it will combine your commits to the one.
After your change, I will merge it.

andycraig: @Ikuyadeu Okay, no rebase. Thanks in advance for squash and merge.

I will add the commit for `config()`.

renkun-ken: I'm totally fine with both.

andycraig:
Diff is much smaller now.

Ikuyadeu:
Great!

andycraig:
@Ikuyadeu @renkun-ken Thanks both!

PR #303, opened 2020-05-02
Refine .Rprofile to remove unnecessary printing in R startup message.

renkun-ken:

An uncessary `[1] TRUE` is printed at startup if user runs `R: Create R terminal`, which is caused by `local(...)` in `R/.Rprofile` returning a logical value visibly.

**(If you have)Screenshot**

image

1. Create R terminal
2. `[1] TRUE` should not appear in the startup message.


renkun-ken:
Thanks!

PR #320, opened 2020-05-09
Add launch tests to launch.json

renkun-ken:
This PR adds a launch configuration to run extension tests according to https://code.visualstudio.com/api/working-with-extensions/testing-extension#debugging-the-tests.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
             "args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
             "stopOnEntry": false,
             "sourceMaps": true,
-            "outFiles": [ "${workspaceRoot}/out/src/**/*.js" ],
+            "outFiles": [ "${workspaceRoot}/out/**/*.js" ],
             "preLaunchTask": "npm"
         },
         {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,12 +12,10 @@
       "args": [
         "--extensionDevelopmentPath=${workspaceRoot}"
       ],
-      "stopOnEntry": false,
-      "sourceMaps": true,
       "outFiles": [
         "${workspaceRoot}/out/**/*.js"
       ],
-      "preLaunchTask": "build"
+      "preLaunchTask": "npm: compile"
     },
     {
       "name": "Extension Tests",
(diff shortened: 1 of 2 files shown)


andycraig:
Thank you for trying to solve this! I tried it but I get the same error I was getting with my earlier attempt:

```
Error: Error: Cannot find module '/home/ac/projects/r/vscode-R-2/vscode-R/out/test/suite/index'
Require stack:
- /usr/share/code/resources/app/out/vs/loader.js
- /usr/share/code/resources/app/out/bootstrap-amd.js
- /usr/share/code/resources/app/out/bootstrap-fork.js
	at h._doHandleExtensionTests (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:827:661)
	at processTicksAndRejections (internal/process/task_queues.js:85:5)
extensionHostProcess.js:826
```

The problem I have is that it doesn't do TypeScript to JavaScript compilation step. If the `out` directory already exists (e.g., from running `npm run test` in the terminal) it looks like it works, but it's running the tests from the existing version of `out` without creating it/updating it. Does it work for you if you delete `out` before running Extension Tests?

renkun-ken:
@andycraig Thanks for testing!

I update the `tasks.json` to the latest format and add a `pretest` task to generate `out` folder. The `Extension Tests` runs `pretest` before launching so that the `out` folder is generated before running tests. The tasks should work normally now.

andycraig:
So that's what was needed! This works for me on Ubuntu. I tested 'Launch Extension' and 'Extension Tests' in Ubuntu and confirmed that both use the current code. I will test on Windows a little bit later and will post the results.

renkun-ken:
The `tasks.json` and `launch.json` are edited to be consistent with the hello-world example introduced at https://code.visualstudio.com/api/get-started/your-first-extension or https://github.com/Microsoft/vscode-extension-samples/tree/master/helloworld-sample.

However, the breakpoints here do not work somehow. I'm trying to figure out why.

andycraig:
On Windows, 6c8077352e5b1016e5c23f7aee8e041c500f5dac (previous commit) works for me but the current commit gives me the 'Cannot find module' error. On Ubuntu, both work. 

Breakpoints don't work for me but they haven't been working for me for a while now.

renkun-ken:
@andycraig Does the latest commit work for you on Windows?

It works for me on both Ubuntu and macOS. The breakpoints also work well on Ubuntu but not on macOS. No clue at the moment.

andycraig:
The latest commit works for me on Windows. And breakpoints work now too! Thank you!

renkun-ken:
Thanks for testing!

PR #325, opened 2020-05-15
Add r.runSelectionRetainCursor

### LINKED ISSUE #324: Run code block but leave cursor in existing position (Rstudio feature)
Rstudio have this feature where you can run the current code "block" but leaving the cursor in its existing position. You already have the ab

ility to run code "blocks" (which is awesome) but would be also great to have this second option available as well!  

Apologies if it already exists and I missed it, gif to help clarify what I mean:

renkun-ken:

Closes #324 

This PR adds a new command `r.runSelectionRetainCursor` that runs selection but does not move cursor.

Run command `r.runSelectionRetainCursor` when cursor is at the first line of the following code:

```r
m <- rnorm(100)
n <- rnorm(100)
```

then the first line should be sent to terminal but the cursor does not move.

It should also work with multi-line expression. Run when cursor is at first of last line:

```r
m <- local({
  x <- rnorm(100)
  y <- rnorm(100)
})
```

Both should send the entire expression but cursor should not move.

### DIFF
--- a/package.json
+++ b/package.json
@@ -333,6 +333,11 @@
         "title": "Run from Beginning to Line",
         "category": "R",
         "command": "r.runFromBeginningToLine"
+      },
+      {
+        "title": "Run Selection/Line (Retain Cursor)",
+        "category": "R",
+        "command": "r.runSelectionRetainCursor"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -348,6 +353,12 @@
         "mac": "cmd+enter",
         "when": "editorTextFocus && editorLangId == 'rmd'"
       },
+      {
+        "command": "r.runSelectionRetainCursor",
+        "key": "alt+enter",
```
renkun-ken, line 358: I set up the key binding here but I'm not sure why it does not come to be my default key bindings when I run the extension.

andycraig: For me alt+enter worked without needing to change anything. Maybe you have something else bound to alt+enter that has higher priority?

renkun-ken:
Thanks for testing!

Issue #327, opened 2020-05-17
Roxygen Comment autocompletes quote symbol


gowerc:
When creating a roxygen comment it auto completes the single quote i.e. typing `#'`  results in the editor being populated with `#''`.   The roxygen comment is just `#'` however so the single quote really shouldn't auto complete in this scenario.

andycraig:
Fixed by #328. Thank you @gowerc!

PR #328, opened 2020-05-17
supress auto-opening quote in roxygen comment

gowerc:
#327 

**(If you have)Screenshot**

Gif to show change still allows single quotes to auto complete when not following a # symbol

Couldn't find any related tests connect to auto-bracketing but please let me know if there are some that I've missed.

Couldn't see anyway of adding context to the auto-bracketing file so decided to make a new auto-bracket specifically for roxygen comments which simply doesn't have any closing tag. Please let me know if there is a better way of doing this


renkun-ken:
This works nicely. @andycraig Do you think if there anything more should be added?

andycraig:
@gowerc Thank you!

@gowerc @renkun-ken I don't know of any way to automate a test for this either so I don't think anything else needs to be added.

andycraig:
LGTM. Thank you @gowerc!

Issue #334, opened 2020-05-23
Current langauge file uses cpp in Rmd whereas it should be Rcpp


gowerc:
From the rmarkdown documentation the supported language is rcpp not cpp 

https://bookdown.org/yihui/rmarkdown/language-engines.html

This is further reinforced by the highlighting supplied by Rstudio for Rmd files:
![image](https://user-images.githubusercontent.com/8447209/82735569-94d63c00-9d1a-11ea-8312-a83f08221f45.png)

The current syntax highlighting in the language file here is set to look for cpp and doesn't current work on rcpp.

![image](https://user-images.githubusercontent.com/8447209/82735593-bb947280-9d1a-11ea-976f-fff4a46a291c.png)

PR #333, opened 2020-05-23
Convert language files to Json

gowerc:
**What problem did you solve?**

Converted language files from XML to json - this should make them much easier to read and edit for other related PRs

**(If you have)Screenshot**

Screen shots of the various file formats effect to show the syntax highlight is still being applied

Rmd:
image

Rd:
image

Being honest I'm not entirely sure what the Rcpp definition is adding so I wasn't sure how to test if that is still working or not !

The only content I've changed in these files is in the Rcpp one in which there appeared to be a duplicate block which I deleted which for reference you can see [here](https://github.com/Ikuyadeu/vscode-R/blob/master/syntax/Rcpp.tmLanguage#L34) and [here](https://github.com/Ikuyadeu/vscode-R/blob/master/syntax/Rcpp.tmLanguage#L44)

PR #335, opened 2020-05-23
Fix rcpp rmd bug

gowerc:
This PR is dependent / branched off #333 

#334

**(If you have)Screenshot**

image

Screenshot shows that syntax highlighting now works for rcpp blocks and not cpp blocks (as per the supported languages spec (here)[https://bookdown.org/yihui/rmarkdown/language-engines.html] but also shows the syntax highlighting for the R block is maintained.

There was a slight complication in that the syntax highlighting for the R block was being falsely matched and applied to the Rcpp block (as it starts with R) so added some stricter regex to the R block to prevent this false match.


renkun-ken:
For `{rcpp}` chunk, it works as expected. However, according to the https://bookdown.org/yihui/rmarkdown/language-engines.html, `{Rcpp}` should work too but it does not:

image

while both `{r}` and `{R}` chunks work:

image

gowerc:
@renkun-ken  - O good spot, will update

Though interestingly you've shown there that the syntax highlighting is maintained for ```r   (rather than ```{r}) however this isn't valid for knitr.  Am wondering if we should stop the syntax highlighting here to give a visual cue to users that they haven't formatted it properly ?  (can make this a separate issue if you'd prefer)

gowerc:
Following up on this some more  apparently `rcpp` isn't valid, it will only render `Rcpp`

image

renkun-ken:
Interesting finding. `{rcpp}` actually is not working? Then I guess we should not provide syntax highlighting for it as it does not work.

gowerc:
Just withdrawing the PR whilst I still iterate on it + other markdown changes

Issue #332, opened 2020-05-22
Feature: Expose send text delay constant as a parameter


jrounds:
Is it possible to expose this as a parameter:
https://github.com/Ikuyadeu/vscode-R/blob/master/src/rTerminal.ts#L119

The current one feels a little slow.  Is that as fast as it goes without error?

andycraig:
Hi @jrounds, thank you for the suggestion. This could be exposed as a parameter. Because of changes to VS Code itself, I think sending text to VS Code's terminal has slowed down slightly since that line was added. For many users, the value could probably be safely decreased.

In the mean time, for much faster sending of text to the console we recommend [radian](https://github.com/randy3k/radian).

PR #336, opened 2020-05-23
expose send text delay as a parameter

gowerc:

#332 

* Open R program
* write R code
* Open R terminal
* send text
* change option for delay text to 2000
* send text gain

It should now take 2 seconds before it sends

Not sure if this needs any more documentation or a better explanation

### DIFF
--- a/package.json
+++ b/package.json
@@ -486,6 +486,11 @@
           "type": "boolean",
           "default": false,
           "description": "Enable R session watcher (experimental). Restart required to take effect."
+        },
+        "r.rtermSendDelay": {
+          "type" : "integer",
+          "default" : 8,
+          "description": "Delay before sending text to rterm"
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -490,7 +490,7 @@
         "r.rtermSendDelay": {
           "type" : "integer",
           "default" : 8,
-          "description": "Delay before sending text to rterm"
+          "description": "Delay in milliseconds before sending each line to rterm (only applies if r.bracketedPaste is false)"
         }
       }
     }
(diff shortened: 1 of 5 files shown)


renkun-ken:
Thanks! This works as expected.

package.json
```diff
@@ -486,6 +486,11 @@
           "type": "boolean",
           "default": false,
           "description": "Enable R session watcher (experimental). Restart required to take effect."
+        },
+        "r.rtermSendDelay": {
+          "type" : "integer",
+          "default" : 8,
+          "description": "Delay before sending text to rterm"
```
renkun-ken, line 493: Should we add the info that this only works with `r.bracketedPaste` disabled and it is in millisecond?

src/rTerminal.ts
```diff
@@ -116,7 +116,7 @@ export async function runTextInTerm(term: Terminal, text: string) {
         term.sendText(text);
     } else {
         for (const line of text.split('\n')) {
-            await delay(8); // Increase delay if RTerm can't handle speed.
+            await delay(config().get('rtermSendDelay')); // Increase delay if RTerm can't handle speed.
```
renkun-ken, line 119: I'm not sure if it is better that we get the value of `rtermSendDelay` before the `for` loop and apply the same number in each iteration, or constantly fetch the value so that it is immediately effective even when user is in the middle of sending a big chunk of code?

package.json
gowerc: Changed the description to:

```
Delay in milliseconds before sending each line to rterm (only applies if r.bracketedPaste is false)
```

Which hopefully should make it super clear what the parameter affects (and when)

src/rTerminal.ts
gowerc: Good point, I think to me it would be unintuitive for the delay speed to change after having submitted the code. I think it also makes sense to not do repeated calls to get the config variable so I have pulled it out to its own variable prior to the for loop

renkun-ken:
LGTM

Issue #337, opened 2020-05-23
Switch to R terminal


gowerc:
If alwaysUseActiveTerminal is False and the current active terminal isn't named ['R', 'R Interactive'] then nothing happens (not even so much as a warning).

Would it not make more sense to switch to the first terminal named ['R', 'R Interactive'] ? At the very least perhaps a message to suggest to the user to open a terminal (or I guess could implement the suggestion in #291 )

(for reference this is how Rstudio handles it if you have the terminal open and you try to run R code, it will flick to the R console)

andycraig:
> If alwaysUseActiveTerminal is False and the current active terminal isn't named ['R', 'R Interactive'] then nothing happens (not even so much as a warning).

Hi @gowerc, in this case if I use `R: Run Selection/Line` a new R terminal called 'R Interactive' is created.

Is your suggestion that, if there is already a terminal called 'R' or 'R Interactive' that is not the active terminal, then it should switch to that terminal instead of creating a new terminal?

renkun-ken:
I think this could be related to #200. `R: Create R terminal` allows user to create multiple terminals. Each time the active terminal is set to the last created one. If the last one is closed, then active terminal is invalid and previously created terminals are no longer tracked.

renkun-ken:
In the case where multiple terminals are created as I mentioned, should we find the terminal with wanted name and set it to be the active terminal?

andycraig:
@renkun-ken That sounds like a good approach. I would suggest we find the most recent (highest numbered) terminal with wanted name.

gowerc:
> Hi @gowerc, in this case if I use `R: Run Selection/Line` a new R terminal called 'R Interactive' is created.

O weird I wasn't seeing this behaviour, will double check to make sure I've set mine up correctly

> Is your suggestion that, if there is already a terminal called 'R' or 'R Interactive' that is not the active terminal, then it should switch to that terminal instead of creating a new terminal?

Yes that is right.   For example I quite often have 3 or more terminals open (R / Bash / Python / etc) and have to manually flick between them,  having the submit code automatically switch to the existing R console would be a nice usability improvement :)

gowerc:
> In the case where multiple terminals are created as I mentioned, should we find the terminal with wanted name and set it to be the active terminal?

@renkun-ken  - When you say most recent, do you mean the most recently created? or the most recently used ?

renkun-ken:
[...]

PR #338, opened 2020-05-23
Added functionality to switch to an existing R terminal

gowerc:

#337  - Though feel free to reject if you don't like the proposal :) 

**(If you have)Screenshot**

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -65,15 +65,15 @@ export async function chooseTerminal(active: boolean = false) {
     }
 
     if (window.terminals.length > 0) {
-        const rTermNameOpinions = ['R', 'R Interactive'];
+        const rTermNameOptions = ['R', 'R Interactive'];
         if (window.activeTerminal !== undefined) {
             const activeTerminalName = window.activeTerminal.name;
-            if (rTermNameOpinions.includes(activeTerminalName)) {
+            if (rTermNameOptions.includes(activeTerminalName)) {
                 return window.activeTerminal;
             }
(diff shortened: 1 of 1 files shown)


src/rTerminal.ts
```diff
@@ -71,6 +71,13 @@ export async function chooseTerminal(active: boolean = false) {
             if (rTermNameOpinions.includes(activeTerminalName)) {
```
renkun-ken, line 71: Is it `rTermNameOpinions` or `rTermNameOptions`? Looks like a typo introduced a while ago? Or I'm missing something?

andycraig: Definitely a typo. (Possibly by me.)

gowerc:
@andycraig  & @renkun-ken 

Added a commit to fix the typo you mentioned.
Also added a commit to try to select the most "recent" R terminal. (more details in #337 )

renkun-ken:
What happens if there are multiple open terminals but none is named as `['R', 'R Interactive']`? It would be nice to show a message or create a new terminal in this case?

gowerc:
@renkun-ken  -  If there aren't any terminals called R then rTerm is left undefined `undefined` and gets picked up by the block [here](https://github.com/Ikuyadeu/vscode-R/blob/ffeab34ec5bf7fe7131fb75aae6d18ce44ebf087/src/rTerminal.ts#L98) and thus attempts to create a new terminal (printing an error message if its unable to do so).

renkun-ken:
Thanks, just forgot the case has been handled previously.

LGTM.

PR #344, opened 2020-05-27
Define lint in package.json and use it in GitHub Actions

renkun-ken:


package.json
```diff
@@ -500,7 +500,8 @@
     "compile": "webpack --mode none",
     "watch": "webpack --mode none --watch",
     "pretest": "tsc -p ./",
-    "test": "node ./out/test/runTest.js"
+    "test": "node ./out/test/runTest.js",
+    "lint": "eslint src --ext ts"
```
Ikuyadeu, line 504: How is
```
 "lint": "eslint --config .eslintrc.js ./src/*.ts"
```
it will use `.eslintrc.js`

renkun-ken: eslint by default uses `.eslintrc.*` if exists. It's perfectly fine if we make it explicit.

Ikuyadeu: Great! I got it.

PR #350, opened 2020-05-28
Fix View empty environment

### LINKED ISSUE #349: View(globalenv()) not working with empty globalenv()
**Describe the bug**

If the global environment is empty, then `View(globalenv())` does not work correctly as it will open a new editor with `NULL` text rather than a WebView.

**To Reproduce**
1. Start an R session with empty global environment.
2. `View(globalenv())`.
3. A new document is open.

renkun-ken:

Close #349

Empty environment is handled to produce an empty table for View.

1. Start an R session with empty global environment
2. `View(globalenv())`
3. An empty table should be presented in a WebView.


renkun-ken:
Thanks for testing!

Issue #306, opened 2020-05-03
Remove Run Selection/Line in Active Terminal


andycraig:
The command `Run Selection/Line in Active Terminal` was added before the setting `Always Use Active Terminal`. `Always Use Active Terminal` has been available for a while now and my guess is that nobody is using `Run Selection/Line in Active Terminal` anymore. If it was removed, I think that a lot of functions could be simplified by avoiding this 'pattern':

```typescript
const callableTerminal = await chooseTerminal();
if (callableTerminal === undefined) {
    return;
}
runSelectionInTerm(callableTerminal);
```

A couple of options:

1. Remove `Run Selection/Line in Active Terminal` and see if anyone complains;
2. Deprecate it by showing a message when it is used, like 'This command will be removed. Enable setting "Always Use Active Terminal" (with option to not show message again). Then remove it after a couple of months.

andycraig:
Showing a warning message with option to not show message again requires adding a setting like`deprecateSendActiveDontShowAgain`.

andycraig:
Option 3: Remove the command’s functionality and just show a message when it is used, like 'This command has been removed. Enable setting "Always Use Active Terminal"'. Then, remove it completely after another couple of months.

That way it’s clear to the user, and we wouldn’t have to add a setting that would only be used temporarily.

Unless anyone has any objections, let’s go with Option 3.

renkun-ken:
Option 3 LGTM

PR #351, opened 2020-05-30
Change runSelectionInActiveTerm effect to warning

andycraig:
Part of #306 

Replaces effect of `R: Run Selection/Line in Active Terminal` with warning message, in preparation for its eventual removal.

Use command `R: Run Selection/Line in Active Terminal`. Observe code is NOT sent to terminal. Observe warning message 'This command has been removed. Enable setting "Always Use Active Terminal".'

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -97,11 +97,7 @@ export function activate(context: ExtensionContext) {
     }
 
     async function runSelectionInActiveTerm() {
-        const callableTerminal = await chooseTerminal(true);
-        if (callableTerminal === undefined) {
-            return;
-        }
-        runSelectionInTerm(callableTerminal, true);
+        window.showWarningMessage('This command has been removed. Enable setting "Always Use Active Terminal".');
     }
 
     async function runSelectionRetainCursor() {


renkun-ken:
LGTM. Please feel free to merge if there's nothing to be added.

src/extension.ts
```diff
@@ -97,11 +97,7 @@ export function activate(context: ExtensionContext) {
     }
 
     async function runSelectionInActiveTerm() {
-        const callableTerminal = await chooseTerminal(true);
-        if (callableTerminal === undefined) {
-            return;
-        }
-        runSelectionInTerm(callableTerminal, true);
+        window.showWarningMessage('This command has been removed. Enable setting "Always Use Active Terminal".');
```
renkun-ken, line 100: In my understanding, a warning is a message that complains but still works. Here do we still want the command to work? If it is not supposed to work any more, should we switch to an error instead?

andycraig: I thought about that too. We don't want the command to work. My logic was that an error makes it look like something has gone wrong, whereas this is more like an alert.

renkun-ken: Sounds fine.

andycraig:
@renkun-ken Thank you for testing and reviewing! I will merge.

andycraig:
@renkun-ken And thank you for merging! 😄

PR #353, opened 2020-06-02
Improve getBrowserHtml

renkun-ken:
As suggested at https://github.com/nx10/httpgd/issues/1#issuecomment-637503382, the browser WebView code is improved to better present embedded iframe. The original code might produce a slight margin on iframe so that a scroll bar appears on the webpage when the page is 100% in the webview while the new code make the webpage better fill in the WebView without a margin.

There are several ways to test this:

* R html help

```r
?get
```

* Shiny example

```r
shiny::runExample("01_hello")
```

* httpgd live page

```r
httpgd::httpgd()
utils::browseURL(httpgd::httpgdURL())

plot(rnorm(100), main = "title")
abline(h = 0, col = "red")
points(rnorm(100), col = "blue")

library(ggplot2)

ggplot(mpg, aes(displ, hwy, colour = class)) +
  geom_point()

par(mfrow = c(2, 2))
m <- lm(mpg ~ cyl, data = mtcars)
plot(m)
```

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -166,11 +166,16 @@ function getBrowserHtml(url: string) {
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <style>
+    html, body {
+        height: 100%;
+        padding: 0;
+        overflow: hidden;
+    }
+  </style>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


renkun-ken:
The problem reported in https://github.com/microsoft/vscode/issues/99208 seems only to exist on macOS. I use Ubuntu at work and there's no such problem.

renkun-ken:
Thanks for testing!

PR #357, opened 2020-06-08
Remove single quote from doesLineEndInOperator

### LINKED ISSUE #356: extendSelection not woking properly with single quote
Send the following code to terminal.

```r
'a'
"hello"
```

Put cursor on the first line, send to terminal, and then it sends both lines to the terminal.

It looks like `extendSelection` does not work properly with single quote.

renkun-ken:

Close #356 

Single quote should not be treated as an operator that continues the code line. Therefore, I removed it from the search pattern in `doesLineEndInOperator`.

New test cases are added. The single quote test case fails with current master and passes with this PR.

**(If you have)Screenshot**

Sending the first line of the following code to terminal no longer sends both lines but only the first line.

```r
'hello'
"hello"
```

### DIFF
--- a/src/lineCache.ts
+++ b/src/lineCache.ts
@@ -47,7 +47,7 @@ function cleanLine(text: string) {
 }
 
 function doesLineEndInOperator(text: string) {
-    const endingOperatorIndex = text.search(/(,|\+|!|\$|\^|&|\*|-|=|:|\'|~|\||\/|\?|%.*%)(\s*|\s*\#.*)$/);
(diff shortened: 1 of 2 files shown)


renkun-ken:
Just forgot to ask, @Ikuyadeu Is there a reason that the single quote appears in the `endingOperatorIndex`?

Ikuyadeu:
@renkun-ken That looks my old miss, I just covered all of the symbol characters.

Ikuyadeu:
Thank you for the fix!
Also, I'm fixing other regular expressions on the #361 like
https://github.com/Ikuyadeu/vscode-R/pull/361/files#diff-abd35639e41649929c4e0dc674ab014eL50

PR #374, opened 2020-06-26
On Mac & Linux, rely on PATH being set up

reactormonk:
Hardcoding the path should only be necessary in case the default configuration via `PATH` fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -457,12 +457,12 @@
         "r.rterm.windows": {
           "type": "string",
           "default": "",
-          "description": "R.exe path for windows."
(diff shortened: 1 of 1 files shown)


renkun-ken:
LGTM

renkun-ken:
I'm wondering if the approach used at https://github.com/REditorSupport/vscode-r-lsp/blob/master/src/extension.ts#L11 has the same behavior?

reactormonk:
Almost. I think you'd also have to change it to detect the executable flag.

PR #390, opened 2020-08-11
Fix previewDataframe for 2+ letter variables

### LINKED ISSUE #389: This does not appear to be a dataframe
I have just completed the install of ikuyadeu.r including radian etc as directed. I'm running vscode on Mojave 10.14.6 and have installed the grapecity.gc-excelviewer extension. However, I cannot view a dataframe. I am trying to view the "cars" dataframe but get an error in the bottom right saying that "This does not appear to be a dataframe". Any help would be appreciated, thanks!

_Originally posted by @tbkw in https://github.com/Ikuyadeu/vscode-R/issues/96#issuecomment-671453426_

andycraig:
Fixes #389 

Command `R: Preview Dataframe` does not work if data frame name is 2+ letters. E.g., `R: Preview Dataframe` works if data frame name is `a`, but not if data frame name is `cars`.

The reason is that `getWordOrSelection()` changed from returning an array of strings to returning a single string, but its use in this function was not updated.

Create a file called `temp.R` containing just the text `cars`. Place cursor on `cars` and use `R: Preview Dataframe`. Observe that preview appears as expected.

Change the text to `cars cars`. Select BOTH WORDS and use `R: Preview Dataframe`. Observer the message 'This does not appear to be a dataframe', as expected (because the selection contains a space).

### DIFF
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -30,10 +30,9 @@ export async function previewDataframe() {
         return undefined;
     }
 
-    const selectedTextArray = getWordOrSelection();
-    const dataframeName = selectedTextArray[0];
+    const dataframeName = getWordOrSelection();
 
-    if (selectedTextArray.length !== 1 || !checkForSpecialCharacters(dataframeName)) {
+    if (!checkForSpecialCharacters(dataframeName)) {
         window.showInformationMessage('This does not appear to be a dataframe.');
 
         return false;


renkun-ken:
LGTM. Feel free to merge.

andycraig:
Thank you for the review!

PR #391, opened 2020-08-11
Add warning to .Rprofil

assignUser:
I used usethis::edit_r_profil() from within vsc which opens this .Rprofil and not ~/.Rprofil.
I did not notice the way this works (did not find any mention of this in the docs either), my changes where lost.
This short warning should help anyone else doing the same :)

p.s. Thanks for this great extension! 😍


renkun-ken:
Thanks for reporting!

I think a better fix for `usethis::edit_r_profil()` could be to revert `R_PROFILE_USER` so that `~/.vscode-R/.Rprofile` is invisible to user in this case.

assignUser:
Yeah that is a good solution, as it restores the functionality of `usethis::edit_r_profil()`. Thanks for the quick reply!

PR #392, opened 2020-08-12
Restore R_PROFILE_USER

renkun-ken:
Close #391 

This PR restores `R_PROFILE_USER` if specified so that `~/.vscode-R/.Rprofile` is hidden on user side when user gets `R_PROFILE_USER` as if it is originally specified.

When user calls `usethis::edit_r_profile()`, if user edits `~/.vscode-R/.Rprofile`, it would be confusing and of no use since the file is replaced each time the extension is activated. Restoring the original `R_PROFILE_USER` might make more sense.

### DIFF
--- a/R/.Rprofile
+++ b/R/.Rprofile
@@ -9,6 +9,11 @@ local({
   }
 
   r_profile <- Sys.getenv("R_PROFILE_USER_OLD")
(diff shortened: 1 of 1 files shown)


andycraig (approved):
LGTM.

Issue #378, opened 2020-07-09
Support VSCode Notebook API


renkun-ken:
We could explore the [VSCode Notebook API](https://code.visualstudio.com/api/extension-guides/notebook) and see if we could build something out from it.

DonJayamanne:
I would like to contribute to this extension and get r vscode working with the Jupyter extension notebook (ipynb).
Basically, the Jupyter extension only deals with opening/saving ipynb files and r extension will be responsible for rich support, execution, language capabilities, etc.

Please advice if such a PR would be welcome and of there's any prior work that I can contribute to or pickup.

@Ikuyadeu @renkun-ken /cc

Ikuyadeu:
@DonJayamanne  Thank you for your suggestion. Of course, we welcome your PR.

Moving notebook functionality to the Jypyter extension helps simplify the role of this extension.
As with Python extensions, we only need to add your plugin to the dependencies.

As an additional future, I hope we can also move the R markdown execution to that extension. (See # 465)
https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells

DonJayamanne:
Based on what I can tell, this extension doesn't really start any R kernel.
It just sends R code to the terminal, is that correct.
If that's the case, then I don't think we need to do anything special, as the existing Jupyter R kernel will work with the Jupyter extension in VS Code without any work.

I assumed this extension had its own R kernel (manually spinning up an R process & the like), however that doesn't seem to be the case.

renkun-ken:
@DonJayamanne #394 is an experimental PR to implement the Notebook API. Only in this PR we implement a simple R kernel that evaluates code from notebook cells.

DonJayamanne:
Thanks.

I've had a look at the code & I'm struggling to see the need for spawning a kernel manually.
Please coul dyou try the Jupyter extension and let me know whats missing.
I'm suggesting this because I don't see the need to create a kernel & manually launch it as done in https://github.com/Ikuyadeu/vscode-R/pull/394.
The only benefit is, a user doesn't need to install a kernel spec. I believe we can get around this by having the R extension register an R kernel in a known location (as done here https://github.com/DonJayamanne/typescript-notebook) & Jupyter extension will pick this and spawn the kernel & communicate with it using the Jupyter protocol.

Thanks. Please could you try using the Native Notebooks in Jupyter extension & let me know what is missing:
* Install VSC Code Insiders (Not Stable)
[...]

PR #394, opened 2020-08-12
WIP: R Notebook

renkun-ken:
This is a very early, proof-of-concept demo of R Notebook using the [VSCode Notebook API](https://code.visualstudio.com/api/extension-guides/notebook) as suggested by #378.

image

Suggestions are welcome!

### DIFF
--- a/R/client.R
+++ b/R/client.R
@@ -1,0 +1,8 @@
+con <- socketConnection(host = "127.0.0.1", port = 8780, blocking = TRUE,
+  server = FALSE, open = "r+", timeout = 86400)
+on.exit(close(con))
+request <- list(client = "hello", time = format(Sys.time()), expr = "1+1")
+json <- jsonlite::toJSON(request, auto_unbox = TRUE)
+writeLines(json, con)
+response <- readLines(con)
+response
--- a/R/client.R
+++ b/R/client.R
@@ -1,7 +1,7 @@
-con <- socketConnection(host = "127.0.0.1", port = 8780, blocking = TRUE,
-  server = FALSE, open = "r+", timeout = 86400)
+con <- socketConnection(host = "127.0.0.1", port = 8781, blocking = TRUE,
+  server = FALSE, open = "w", timeout = 86400)
 on.exit(close(con))
-request <- list(client = "hello", time = format(Sys.time()), expr = "1+1")
+request <- list(time = format(Sys.time()), expr = "1+1")
 json <- jsonlite::toJSON(request, auto_unbox = TRUE)
 writeLines(json, con)
 response <- readLines(con)
--- a/R/client.R
+++ b/R/client.R
@@ -1,8 +1,8 @@
-con <- socketConnection(host = "127.0.0.1", port = 8781, blocking = TRUE,
-  server = FALSE, open = "w", timeout = 86400)
-on.exit(close(con))
+con <- socketConnection(host = "127.0.0.1", port = 8708,
+  blocking = TRUE, server = FALSE, open = "r+")
 request <- list(time = format(Sys.time()), expr = "1+1")
 json <- jsonlite::toJSON(request, auto_unbox = TRUE)
 writeLines(json, con)
-response <- readLines(con)
+response <- readLines(con, n = 1)
 response
+close(con)
--- a/R/notebook.R
+++ b/R/notebook.R
@@ -7,8 +7,9 @@ local({
     eval(expr, env)
   }
 
-  print(ls.str(env))
+  ls.str(env)
   while (TRUE) {
+    cat("Waiting\n")
     con <- socketConnection(host = "127.0.0.1", port = env$port,
       blocking = TRUE, server = TRUE,
       open = "r+")
@@ -22,7 +23,7 @@ local({
       type = if (inherits(res, "try-error")) "error" else "output",
       result = utils::capture.output(print(res))
     )
-    result <- jsonlite::toJSON(str, force = TRUE)
+    result <- jsonlite::toJSON(str, auto_unbox = TRUE, force = TRUE)
     writeLines(result, con)
     close(con)
   }
--- a/R/notebook.R
+++ b/R/notebook.R
@@ -9,26 +9,21 @@ local({
 
   print(ls.str(env))
   while (TRUE) {
-    con <- try(socketConnection(host = "127.0.0.1", port = env$port,
+    con <- socketConnection(host = "127.0.0.1", port = env$port,
       blocking = TRUE, server = TRUE,
-      open = "r+", timeout = 86400L), silent = TRUE)
-    if (inherits(con, "try-error")) {
-      message(con, "\n")
-    } else {
-      tryCatch({
-        request <- jsonlite::fromJSON(readLines(con, n = 1))
-        cat(sprintf("[%s] %s\n",
-          request$time, request$expr))
-        expr <- parse(text = request$expr)
-        res <- try(eval(expr, globalenv()), silent = TRUE)
-        str <- list(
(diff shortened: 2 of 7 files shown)


andycraig:
Wow, this looks amazing!

I tried to run it in Windows but didn't have much luck so I'm posting the details.

**Version without YAML header**

Content of R Markdown document:

    ```{r}
    print("hello")
    ```

I run command `Notebook: Execute Notebook`. Nothing changes in the notebook itself. Here is the output from the debug console:

```
R stderr (24952): Loading required namespace: jsonlite

extension.js:6336
R stdout (24952): Waiting

extension.js:6340
connected to server!
extension.js:6360
Send: {"time":1597228000521,"expr":""}

extension.js:6365
R stdout (24952): [1597228000521] 

extension.js:6340
{"type":"output","result":"NULL"}

extension.js:6374
disconnected from server
extension.js:6369
R stdout (24952): Waiting
```

**Version with YAML header**

Content of R Markdown document:

    ---
    title: test
    ---
    
    ```{r}
    print("hello")
    ```

When I do `Notebook: Execute Notebook` I get this output in the debug console:

```
R stderr (16720): Loading required namespace: jsonlite

extension.js:6336
R stdout (16720): Waiting

extension.js:6340
R stderr (16720): Error in socketConnection(host = "127.0.0.1", port = env$port, blocking = TRUE,  : 
  cannot open the connection
Calls: local ... eval.parent -> eval -> eval -> eval -> eval -> socketConnection
In addition: Warning message:

extension.js:6336
R stderr (16720): In socketConnection(host = "127.0.0.1", port = env$port, blocking = TRUE,  :
  problem in listening on this socket
Execution halted

extension.js:6336
R exited with code 1
```

The cell displays the message `connect ECONNREFUSED 127.0.0.1:64110`. Also the header shows 'title'.

header

Hope that helps, and let me know if it's because my setup is wrong or similar.

This is very exciting feature, I'm really looking forward to using it!

andycraig:
The debug output for the version with the YAML header is probably because I was running it in the same session. If I run it in a fresh session, here's the debug console output I get:

```
R stderr (23076): Loading required namespace: jsonlite

extension.js:6336
R stdout (23076): Waiting

extension.js:6340
connected to server!
extension.js:6360
Send: {"time":1597229316281,"expr":"---\ntitle: test\n---"}

extension.js:6365
R stdout (23076): [1597229316281] ---
title: test
---

extension.js:6340
R stderr (23076): Error in parse(text = request$expr) : <text>:4:0: unexpected end of input
2: title: test
3: ---
  ^
Calls: local ... eval.parent -> eval -> eval -> eval -> eval -> parse
Execution halted

extension.js:6336
R exited with code 1
```

renkun-ken:
The demo is probably too early for any meaningful test. Nothing works at the moment. 🤦

andycraig:
Ah, okay! Still very exciting 😃

Ikuyadeu:
Great works!
It will be a connector with python users and R users.

renkun-ken:
https://github.com/julia-vscode/julia-vscode/pull/980 has a good discussion of the API and implementation details we might also benefit from.

renkun-ken:
@jrieken, I encounter another problem: When I delete any cell in the notebook, the language server received a shutdown signal, and the language server is stopped. I'm wondering if it is intended behavior by sending shutdown request to language server on cell delete? Or am I missing something?

renkun-ken:
I guess it is caused by https://github.com/REditorSupport/vscode-r-lsp/pull/59/files#diff-45327f86d4438556066de133327f4ca2R215-L183 where I want to stop the langauge server for single file outside workspace. A notebook is a collection of `TextDocument`. Removing a cell closes the `TextDocment` and then stops the language server.

renkun-ken:
Fixed via https://github.com/REditorSupport/vscode-r-lsp/pull/59/commits/1a159d4c33c4c36231bfb7f642cef82e7cc52357.

jrieken:
Yeah, whenever a cell is removed we emit a `onDidCloseTextDocument`-event. Your fix look OK, tho it builds on the "internal" cell uri structure. We have also proposed `TextDocument#notebook` (see https://github.com/microsoft/vscode/issues/102091) which would allow you to check on the notebook of a document directly - assuming a document references a notebook. We should maybe also add `NotebookDocument#isClosed` which would then allow you to have `document.notebook?.isClosed`

renkun-ken:
@jrieken Thanks!

renkun-ken:
I re-implement the `RKernel` and `notebook.R` using `callr::r_session` to support cancelling execution of cells.

@jrieken I've got a few questions before I move on.

1. Regarding `cancelAllCellsExecution`. It seems `Notebook: Cancel Notebook Execution` command does not call `cancelAllCellsExecution`. I'm not sure how I could get it called?
2. Regarding *Display htmlwidgets in cell output*. An htmlwidget produced in R is an HTML file that uses a number of local js and css files with relative paths. The source code looks like:

image

which looks like the following in web browser

image

I'm wondering if it is already supported in Notebook to show such HTML content properly in cell output? I tried replacing the resource paths to something like `vscode-webview-resource://` (https://github.com/microsoft/vscode/issues/106516) but it doesn't work.

renkun-ken:
A note: `R_BROWSER` could be used to allow `callr` session to launch browser as discussed at https://github.com/r-lib/callr/issues/177.

markbaas:
@renkun-ken I'm playing around a bit with your code. For starters I rebased it to master: https://github.com/markbaas/vscode-R/tree/notebook-mark, perhaps you want to merge that part already ;)

markbaas:
@renkun-ken supports the view now. Did the trick with an iframe.
html output is also supported. But i'm rewriting this. Any chance you can a have a look at my progress?

renkun-ken:
Good to see you continue working on the notebook feature! I'll take a look when possible.

markbaas:
@renkun-ken @jrieken 

- [x] proper implementation of table output with a custom renderer
- [x] proper implementation of the viewer output through a custom renderer + iframe

prent
prent
also supports raw data
prent

markbaas:
> [julia-vscode/julia-vscode#980](https://github.com/julia-vscode/julia-vscode/pull/980) has a good discussion of the API and implementation details we might also benefit from.

About saving output I see we have a few options. 
1. Like the html_notebook output saving output in a rendered html file. But this would divert from the vscode structure.
2. Save output in a separate json file. Benefit here is that we can just put in the raw response that we get from the notebook kernel. Probably it would also be beneficial to keep an .Rdata file with the current workspace, so you can continue working where you left off.

@renkun-ken what approach would be preferred?

renkun-ken:
Thanks for the work, @markbaas!

>Save output in a separate json file. Benefit here is that we can just put in the raw response that we get from the notebook kernel. Probably it would also be beneficial to keep an .Rdata file with the current workspace, so you can continue working where you left off.

I think this makes good sense and might better fit more scenarios.

markbaas:
@jrieken 

why is the are the `outputs` and `metadata` properties deprecated for reading as well. I understand this is the case for writing, but for reading its a convenient way to get the metadata for the specific cell and I see no other way.

```		/** @deprecated use WorkspaceEdit.replaceCellOutput */
		outputs: CellOutput[];
		// readonly outputs2: NotebookCellOutput[];
		/** @deprecated use WorkspaceEdit.replaceCellMetadata */
		metadata: NotebookCellMetadata;
```

markbaas:
@renkun-ken Added support for restoring saved cells... only that it doesn't work. I have no clue what the problem is. Perhaps a bug in the api.

Tal500:
Looks awesome! Any update on this?

Issue #405, opened 2020-08-31
Insert Roxygen Skeleton


assignUser:
In Rstudio you can insert a roxygen skeleton with the correct param tags via keybind (when the cursor is within a function definition) to create e.g. this:
```r
#' Title
#'
#' @param x 
#'
#' @return
#' @export
#'
#' @examples
exported <- function(x) {
    x + 1
}
```

It would be a nice qol feature for package developers.
I might look into this but I have yet to make the first steps with vsc extension development/ts and wanted to put the idea out here :)

andycraig:
Hi @assignUser, thank you for the suggestion. This seems like a good fit for a snippet. Users can add their own snippets. Also, snippets can be added to the extension without needing to write TypeScript.

Here is a guide on how to create them: https://code.visualstudio.com/docs/editor/userdefinedsnippets#_create-your-own-snippets

assignUser:
@andycraig Hey thanks for teh quick reply. I will look into it but I am not sure that snippets are flexible enough to completely emulate the rstudio function which adds params depending on the function signature. But it is a good quick fix for now I guess 👍

andycraig:
@assignUser Oh, I see what you mean. Yes, that would probably require some TypeScript rather than just a snippet.

If you decide you want to try implementing this, we can provide advice etc. as needed.

renkun-ken:
I guess this could be made much easier to implement if we define some customized requests in language server so that editor clients could query the information of the parsed data of the document at a document point (row, column). Otherwise, detecting the function arguments could be non-trivial in many cases when cursor is inside the function body, e.g., complex function argument default value, cursor at closure in function, cursor at string literal in function.

Having access to the language client will make it much easier to implement customized language-related features such as this. Unfortunately, vscode-r-lsp and vscode-R are separate projects. I'm not sure if there's a way to access the language client created in vscode-r-lsp unless we merge the two projects (#98) previously done at https://github.com/renkun-ken/vscode-R/tree/lsp.

MilesMcBain:
This would come very cheaply with #302. That way we could take advantage or R code that already exists to do this.

I know of code in {sinew} and {fnmate}, there's probably a few others.

renkun-ken:
@MilesMcBain I didn't know there's already RStudio plugins doing all these. Thanks for letting us know!

MilesMcBain:
[...]

Issue #302, opened 2020-05-02
Support RStudio plugins


andycraig:
RStudio plugins communicate with RStudio via the `rstudioapi` package. It looks like this communicates with RStudio via a file-based communication protocol, similar to the session watcher. See `callRemote` in https://github.com/rstudio/rstudioapi/blob/ac27784fda4ed5a8f7e8d0c8159f841aac22010b/R/remote.R

It seems like we could theoretically support RStudio plugins by running them in a separate R session and supporting the file-based communication protocol. We would have to implement in TypeScript the `rstudioapi` functions like:

- `getActiveDocumentContext`
- `insertText`
- `setDocumentContext`

An obstacle is that `rstudioapi` does some checking of RStudio versions etc. so vscode-R might have to identify itself as RStudio at a couple of points. That seems weird and might have side effects. There could even be some copyright/trademark issue, in which case we would DEFINITELY not want to implement this.

Even if we cleared those obstacles, this would of course be a lot of work, and would place further demand on the system by adding another R session. So even if it's possible it's not necessary something we want.

renkun-ken:
Thanks for raising this. I thought about this too but didn't have time to investigate the underlying protocol between RStudio and the plugins.

We may raise an issue at RStudio and hear from them about whether the protocol and rstudioapi could be used outside RStudio and let other editors leverage the same protocol?

andycraig:
> We may raise an issue at RStudio and hear from them about whether the protocol and rstudioapi could be used outside RStudio and let other editors leverage the same protocol?

This sounds like a good idea to me. If you would like to raise the issue, please do.

MilesMcBain:
I had this idea myself today!

I think something analogous to the `rstudioapi` is highly desirable, elsewise the burden on the developer for developing things like `datapasta` and `fnmate` to VSCode is very high. I would have to create my own VSCode extensions or get them merged here to port them.

So I don't know if collaborating with RStudio is absolutely necessary if that is the blocker. VSCode could provide adapter functions that mock the `rstudioapi` by overwriting the functions in the `rstudioapi` namespace at run time. So we could literally reassign the likes of `rstudioapi::isAvailable` and `rstudioapi::getActiveDocumentContext` to functions that talk instead to VSCode via the session watcher protocol (need to build in a response protocol). For example using `assignInNamespace()`.

It could be done with the other init.R stuff by setting an onLoad hook for `rstudioapi`, e.g. 

```
[...]

PR #406, opened 2020-09-01
Check url in browser

renkun-ken:
**What problem did you solve?**

Closes #371 

**(If you do not have screenshot) How can I check this pull request?**

Following localhost http content it should work:

```r
?get
shiny::runExample("01_hello")
```

External http content should open in external web browser due to WebView restrictions on external web content:

```r
browseURL("https://google.com")
```

Local HTML file should open in external web browser due to WebView restrictions on resource location:

```r
tmpfile <- tempfile(fileext = ".html")
writeLines(c("<html>hello</html>"), tmpfile) 
browseURL(tmpfile)
```

Local non-HTML file should open in vscode editor as text document:

```r
tmpfile <- tempfile(fileext = ".txt")
writeLines("hello", tmpfile)
browseURL(tmpfile)
```

PR #408, opened 2020-09-08
WIP: RStudio Addin Support

MilesMcBain:
Addressing issue #302. Support for addins in VSCode via `rstudioapi` emulation.

# Capability Added

This PR would add an `rstudioapi` emulation layer to the extension. This would allow R functions that call the `rstudioapi` to be run, and behave as expected from within VSCode. Typically the `rstudioapi` is used by user-created RStudio addins to add functionality to the RStudio IDE.

One such example is the package [`{datapasta}`](https://github.com/milesmcbain/datapasta) which, among many things, formats data.frames on the clipboard as `tibble::tribble()` calls and inserts them into the active source editor or terminal. This relies on only a small subset of `rstudioapi` functionality for getting document context and inserting text.

It would also be possible to use existing RStudio addins to bridge functionality gaps. E.g. Users have requested a Roxygen skeleton generator in #405. With `rstudioapi` emulation in play we could recommend they make bindings (or ship them) to addin functions in [`{docthis}`](https://github.com/mdlincoln/docthis) or [`{sinew}`](https://github.com/yonicd/sinew).

Implemented functions:
*  ` getActiveDocumentContext `
 *  ` getSourceEditorContext `
 *  ` isAvailable `
 *  ` verifyAvailable `
 *  ` insertText `
 *  ` modifyRange `
 *  ` readPreference `
 *  ` readRStudioPreference `
 *  ` hasFun `
 *  ` findFun `
 *  ` showDialog `
 *  ` navigateToFile `
 *  ` setSelectionRanges `
 *  ` setCursorPosition `
 *  `setDocumentContents`
 *  ` documentSave `
 *  ` documentId `
 *  ` documentPath `
 *  ` documentSaveAll `
 *  ` documentNew `
 *  ` getActiveProject `
 *  `restartSession`
* `viewer`

# Testing It

This is challenging since most packages that do test rstudioapi stuff end up mocking the api anyway. I have sought feedback from R users on Twitter as to popular addins. We could make a short list and manually confirm the addins work. The list would contain (WIP):
* `{reprex}`
* `{datapasta}`
* `{fnmate}`
* `{equisse}`
* `{remedy}`
* `{citr}`
* `{blogdown}`
* `{AlignAssign}`
* `{prefixer}`


renkun-ken:
Just did minor fixes to comfort eslint.

It would be very nice if you put some minimal examples about what will be achieved in this PR and how we could test against what has been done.

Looks very promising! Cheers!

renkun-ken:
Thanks for the nice work!

I play with `reprex::reprex_addin()` and choose clipboard and it works nicely. But `reprex::reprex_selection()` does not seem to work:

```
> reprex::reprex_selection()                                                                             
Error in verifyAvailable() : 
  argument "version_needed" is missing, with no default
Backtrace:
1: verifyAvailable()
2: callFun(fn)
3: getDocumentContext("getSourceEditorContext")
4: rstudioapi::getSourceEditorContext()
5: rstudio_context()
6: rstudioapi::primary_selection(context)
7: rstudio_selection()
8: reprex(input = rstudio_selection(), venue = venue)
9: reprex::reprex_selection()
```

MilesMcBain:
Oh sorry that list of add-ons was aspirational. As you can see from the trace there's a fair few functions that need implementation yet.

I only have those 4 listed above. This is going to take a while!

MilesMcBain:
Also that trace highlights a problem with my `verifyAvailable()`. It needs to default the version arg to `NULL`. Thanks!

MilesMcBain:
It's milestone time! @andycraig @renkun-ken  

datapasta_vscode

Currently only works with the GitHub dev version of `datapasta` because I was making an undocumented call to the RStudio runtime which has since been replaced by a proper `rstudioapi` function. It was a one line change. Not bad! 

I think what I have in the PR so far will covers all the addins I've written. Not sure about the others on that list. Still a ways to go probably.

MilesMcBain:
Working through the eslint stuff now

andycraig:
@MilesMcBain Great stuff, that screencap is a joy to watch!

I'm totally okay with only supporting a subset of plugins initially. Would it be possible/practical to give the user a message if a plugin tries to use a function that isn't supported yet?

When you want some testing done let us know!

MilesMcBain:
Yep I think a 'not implemented' message is a good way to go. Shouldn't be hard to do.

From here:

  * I realised there's a couple of functions still needed for 100% `datapasta` compatibility which I may as well add. This is to do with recent stuff that was PR'd to make it work with RStudio server. From my reading, if I add these functions it should work with Remote VSCode sessions as well.
 * I'll run through that shortlist of addins and make a table of what needs to be added to support what. We can decide what is worth including in an initial release from there.
 * It would be good to have some kind of front end "addins menu" similar to RStudio. I'm not yet familiar with this part of the VSCode API but it seems doable based on other extensions I have used that do this. I'd be interested in opinions as to how this should work.
  - RStudio refreshes the menu options every time you install a package or restart RStudio, but I think it probably does it in background threads since it involves grepping through your entire library for `./inst/addins.dcf` files.
  - It might be easiest to have some kind of "search for addins" command that synchronously updates a list of installed addins held in the ~/.vscode-R directory.

renkun-ken:
To get the list of addins and their bindings, I tried the following code:

```r
pkgs <- .packages(all.available = TRUE)
addin_files <- vapply(pkgs, function(pkg) {
  system.file("rstudio/addins.dcf", package = pkg)
}, character(1L))
addin_files <- addin_files[file.exists(addin_files)]
descs <- lapply(addin_files, function(file) {
  desc <- read.dcf(file)
  as.data.frame(desc)
})
jsonlite::toJSON(descs)
```

which outputs all addins information from the library so that vscode-R could read it easily:

```json
{
  "blogdown": [
    {
      "Name": "Serve Site",
      "Description": "Run blogdown::serve_site() to live preview a website locally.",
      "Binding": "serve_site",
      "Interactive": "true"
    },
    {
      "Name": "New Post",
      "Description": "Create a new post with blogdown::new_post().",
      "Binding": "new_post_addin",
      "Interactive": "true"
    },
    {
      "Name": "Update Metadata",
      "Description": "Update the title, author, date, categories, and tags of the current blog post.",
      "Binding": "update_meta_addin",
      "Interactive": "true"
    },
    {
      "Name": "Insert Image",
      "Description": "Insert an external image into a blog post.",
      "Binding": "insert_image_addin",
      "Interactive": "true"
    },
    {
      "Name": "Touch File",
      "Description": "Change the timestamp of the current file in the editor.",
      "Binding": "touch_file_rstudio",
      "Interactive": "false"
    },
    {
      "Name": "Quote Poem",
      "Description": "Add > to the beginning of selected paragraphs and two trailing spaces to selected lines.",
      "Binding": "quote_poem_addin",
      "Interactive": "false"
    }
  ],
  "bookdown": [
    {
      "Name": "Preview Book",
      "Description": "Run bookdown::serve_book() to live preview a book.",
      "Binding": "serve_book",
      "Interactive": "true"
    },
    {
      "Name": "Input LaTeX Math",
      "Description": "Input math expressions via the MathQuill library.",
      "Binding": "mathquill",
      "Interactive": "true"
    }
  ],
  "clipr": [
    {
      "Name": "Value to clipboard",
      "Description": "Copies the results of a selected expression to the system clipboard",
      "Binding": "clipr_result",
      "Interactive": "false"
    },
    {
      "Name": "Output to clipboard",
      "Description": "Copies the console output of a selected expression to the system clipboard",
      "Binding": "clipr_output",
      "Interactive": "false"
    }
  ],
  "covr": [
    {
      "Name": "Calculate package test coverage",
      "Description": "Calculates the package test coverage and opens a report, using `covr::report()`",
      "Binding": "addin_report",
      "Interactive": "false"
    }
  ],
  "datapasta": [
    {
      "Name": "Paste as tribble",
      "Description": "Pastes a table from the clipboard to the editor as a tribble definition",
      "Binding": "tribble_paste",
      "Interactive": "false"
    },
    {
      "Name": "Paste as vector",
      "Description": "Pastes data from the clipboard to the editor as character vector, formatted horizontally on a single line.",
      "Binding": "vector_paste",
      "Interactive": "false"
    },
    {
      "Name": "Paste as vector (vertical)",
      "Description": "Pastes data from the clipboard to the editor as character vector, formatted vertically, one element per line.",
      "Binding": "vector_paste_vertical",
      "Interactive": "false"
    },
    {
      "Name": "Paste as data.frame",
      "Description": "Pastes a table from the clipboard to the editor as a data.frame definition",
      "Binding": "df_paste",
      "Interactive": "false"
    },
    {
      "Name": "Paste as data.table",
      "Description": "Pastes a table from the clipboard to the editor as a data.table definition",
      "Binding": "dt_paste",
      "Interactive": "false"
    },
    {
      "Name": "Fiddle Selection",
      "Description": "Take a selection and fiddle it to something better (maybe).",
      "Binding": "zzz_rs_dfiddle",
      "Interactive": "false"
    },
    {
      "Name": "Toggle Vector Quotes",
      "Description": "Toggle quotes in a vector defintion",
      "Binding": "zzz_rs_toggle_quotes",
      "Interactive": "false"
    }
  ],
  "devtools": [
    {
      "Name": "Run a test file",
      "Description": "Run the current test file, using `devtools::test_file()`.",
      "Binding": "test_file",
      "Interactive": "false"
    },
    {
      "Name": "Report test coverage for a file",
      "Description": "Calculate and report test coverage for the current test file, using `devtools::test_coverage_file()`.",
      "Binding": "test_coverage_file",
      "Interactive": "false"
    },
    {
      "Name": "Report test coverage for a package",
      "Description": "Calculate and report the test coverage for the current package, using `devtools::test_coverage()`.",
      "Binding": "test_coverage",
      "Interactive": "false"
    },
    {
      "Name": "Document a package",
      "Description": "A wrapper for `roxygen`'s `roxygen2::roxygenize()`",
      "Binding": "document",
      "Interactive": "false"
    }
  ],
  "lintr": [
    {
      "Name": "Lint current file",
      "Description": "Runs lintr::lint on the current file",
      "Binding": "addin_lint",
      "Interative": "false"
    },
    {
      "Name": "Lint current package",
      "Description": "Runs lintr::lint_package",
      "Binding": "addin_lint_package",
      "Interative": "false"
    }
  ],
  "reprex": [
    {
      "Name": "Render reprex...",
      "Description": "Run `reprex::reprex()` to prepare a reproducible example for sharing.",
      "Binding": "reprex_addin",
      "Interactive": "true"
    },
    {
      "Name": "Reprex selection",
      "Description": "Prepare reprex from current selection",
      "Binding": "reprex_selection",
      "Interactive": "false"
    }
  ],
  "styler": [
    {
      "Name": "Set style",
      "Description": "Prompt for and set the style transformers used by all styler addins",
      "Binding": "set_style_transformers",
      "Interactive": "true"
    },
    {
      "Name": "Style selection",
      "Description": "Pretty-print selection",
      "Binding": "style_selection",
      "Interactive": "true"
    },
    {
      "Name": "Style active file",
      "Description": "Pretty-print active file",
      "Binding": "style_active_file",
      "Interactive": "true"
    },
    {
      "Name": "Style active package",
      "Description": "Pretty-print active package",
      "Binding": "style_active_pkg",
      "Interactive": "true"
    }
  ]
}
```

We could add it to session watcher or addins init script so that the addins could be written to a JSON file in `~/.vscode-R` so vscode-R could know the addins on each session startup. We could add some binding (probably hacking?) to package install functions to refresh the list.

andycraig:
> It would be good to have some kind of front end "addins menu" similar to RStudio. I'm not yet familiar with this part of the VSCode API but it seems doable based on other extensions I have used that do this.

The QuickPick API might be the way to go here. It's not used in vscode-R at the moment but here's an example extension that does use it: https://github.com/microsoft/vscode-extension-samples/tree/master/quickinput-sample

MilesMcBain:
Thankyou both @renkun-ken and @andycraig, between the two of you, most of the research for the menu feature is done!

shirdekel:
Thanks for this! I'm keen to test it out ({fnmate} has become pretty essential in my workflow). I'm new to VS Code, however, so what is the best way to get this version?

renkun-ken:
@shirdekel You could go to https://github.com/Ikuyadeu/vscode-R/actions, choose the latest build from this PR, download the `vscode-R.vsix` artifact, install it in vscode, and play with this feature and see if it works for you.

shirdekel:
Thanks @renkun-ken! I managed to install, but unfortunately I got the following when running `fnmate::rs_fnmate()` (from a keybinding):

```
Error in request_response("active_editor_context") : 
  Did not receive a response from VSCode-R API within 30 seconds.
```

Is this something you think I can diagnose further, or just is a matter of waiting for compatibility?

MilesMcBain:
Thanks for jumping in and testing @shirdekel! I just tested `{fnmate}` and it works locally for me. Does `rstudioapi::getActiveDocumentContext()` return anything for you if you run it in VSCode?

MilesMcBain:
I just did a quick survey of functions in addin packages and compiled into the table in the header post of this thread.

The only one that would stump me is `rstudioapi::sourceMarkers`. It's an API for setting source annotations for linters and such. I had no idea it existed and I've not seen it widely used. 

It's clear this type of thing is possible in VSCode since the R package `{lintr}` is working with the extension and annotating R source. I'm struggling to find the functions for this in the VSCode API. I think I probably won't implement it but it would be nice to be able to research it a bit. Any ideas?

andycraig:
@MilesMcBain I haven't worked with them before but I think those might be Diagnostics. There's a little animation here that shows them: https://code.visualstudio.com/api/language-extensions/programmatic-language-features#provide-diagnostics

Here they are in the API docs: https://code.visualstudio.com/api/references/vscode-api#Diagnostic

renkun-ken:
In languageserver, we run lintr and publish diagnostics to open documents.

vscode-R could publish diagnotics in addition to lintr results too, just like it also provides session symbol completions in addition to languageserver completions.

shirdekel:
> Thanks for jumping in and testing @shirdekel! I just tested `{fnmate}` and it works locally for me. Does `rstudioapi::getActiveDocumentContext()` return anything for you if you run it in VSCode?

I got the same kind of error:

```
> rstudioapi::getActiveDocumentContext() 
Error in request_response("active_editor_context") : 
  Did not receive a response from VSCode-R API within 30 seconds.
```

shirdekel:
Update: it works! 

Can't be sure exactly what it was, but I just recently redid some of my setup (changed my `radian` binary location through a `python` reinstall). Sorry I can't be more helpful with the diagnosis, but at least it's working now :)

MilesMcBain:
>vscode-R could publish diagnotics in addition to lintr results too, just like it also provides session symbol completions in addition to languageserver completions

Thanks @renkun-ken that's helpful. So is it as simple as creating a `Diagnostic Collection` and using the `set` method to publish them in a document? If so that would make the two APIs very similar, and the implementation not that hard.

renkun-ken:
@MilesMcBain Yes, it works exactly as you describe. A problem with such is that when should we clear (or unset) these diagnostics? If we never clear it, I guess they will be there forever. Usually, diagnotics are updated on document change. I'm not sure if similar logic applies to addin-supplied diagnostics.

In languageserver, lintr is run and diagnotics are updated on document open and change, and cleared on document close.

MilesMcBain:
Ahhh riiight. Yes the `rstudioapi` doesn't give the user a callback, which is quite different. I'll to have a play and see how they're invalidated. If it's all handled on opaquely the RStudio end, that could be quite challenging.

MilesMcBain:
I had a play with `sourceMarkers` this evening. It's a bit less sophisticated than I thought. It's not annotations in the source it's just a panel of source links that have colour coded icons and messages:

image

It also has a fiddly API that I couldn't make work with data.frames, even though this is supposed to be supported, only lists worked for me. 

I am thinking pass on this for now.

MilesMcBain:
@andycraig @renkun-ken getting very warm now!

`{datapasta}`:
datapasta_vscode2

`{reprex}`:
reprex_vscode

Still some tinkering to do with the picker. The filtering based on what you type is a bit wonky.

Most things I try seem to work, except anything that tries to launch a Shiny App falls foul of this type of thing:

```
Error in normalizePath(url, "/", mustWork = TRUE) : 
  path[1]="http://127.0.0.1:4897": No such file or directory
```

I've seen this myself with `mapedit` and VSCode and I assume it's an issue with the VSCode viewer, and that it is known? Or is that not the case?

renkun-ken:
@MilesMcBain Looks quite smooth now!

The `normalizePath(url, "/", mustWork = TRUE)` error looks like a misuse of `viewer()` while it should be `browser()` since `viewer()` should only work for local files while `browser()` works with URL like `http://127.0.0.1:4897`.

MilesMcBain:
I think it's not the addin that's doing this, it's `{shiny}` itself, see:

https://github.com/rstudio/shiny/blob/master/R/viewer.R#L23

called by `shiny::runGadget` which seems to be what some addins are calling.

MilesMcBain:
I supposed we could shim this function too? 🤔

MilesMcBain:
On second thought your idea is better. Probably good practice to overwrite as few functions as possible.

renkun-ken:
Ok, it calls `viewer()` instead of `browser()` to show browser page. It could be easily handled on our side by checking input path in our `viewer()`.

MilesMcBain:
Cool. That should probably be a separate PR?

renkun-ken:
Would you mind if I push to your PR branch to fix this directly?

MilesMcBain:
Yeah that's also fine 👍  . Sorry there's all those lint errors. I'll clean that up soon. 😬

renkun-ken:
Looks like this PR this behind master and does not include #406. There's major overlap between browser and viewer check. I guess I should do a separate PR for unify these functions.

MilesMcBain:
If you want to quickly test it , you can try: `mapedit::editMap(mapview::mapview())`

renkun-ken:
`mapedit::editMap(mapview::mapview())` works via c143fcf now.

image

R/init.R
```diff
@@ -385,6 +395,54 @@ if (interactive() &&
         page_viewer = page_viewer
       )
 
+      #vsc-r-api
+      RESPONSE_TIMEOUT <- 30
+      response_lock_file <- file.path(dir_extension, "response.lock")
+      response_file <- file.path(dir_extension, "response.log")
+      addin_registry <- file.path(dir_extension, "addins.json")
```
renkun-ken, line 402: I guess we should write `addins.json` to session temp folder? Then we could allow different sessions to have different set of addins, which might be useful if the sessions are started with independent user libraries (e.g. `renv` is used). The command shows the addins of the attached session.

MilesMcBain: yep good idea

MilesMcBain:
I implemented the session addins.json. This highlighted a race condition issue, where the list of addin items for the menu may not be ready if init.R has not completed in time. I had to jack up the 'wam up' time here: https://github.com/Ikuyadeu/vscode-R/pull/408/files#diff-d33c533ae558f85846acbe177816358aR100. It would be nice if there were some way to wait until init.R is certainly complete? Maybe it could spin waiting for a status in a file?

Shiny apps working has highlighted another issue that needs some thought. When VSCode opens a viewer or browser that takes focus, for some reason the `window.activeTextEditor` gets set to `undefined`. Which means the addin callbacks can't use `window.activeTextEditor` to insert the results coming from the shiny app session.

Try `bookdown::mathquill()` for an example. If I set `vsc.browser` and `vsc.viewer` to `FALSE` things work as expected.

I think to make these apps work properly I need a handle to the text editor that was active before the view/browser panel is opened? Seems messy.

renkun-ken:
We know that `init.R` is complete when `.vsc.attach` is called and `attach` command is received. The addins list should be updated upon each attach, just like `globalenv.json` to update the session completion provider.

MilesMcBain:
> We know that init.R is complete when .vsc.attach is called and attach command is received. The addins list should be updated upon each attach, just like globalenv.json to update the session completion provider.

Ah perfect, thanks.

renkun-ken:
>Shiny apps working has highlighted another issue that needs some thought. When VSCode opens a viewer or browser that takes focus, for some reason the window.activeTextEditor gets set to undefined. Which means the addin callbacks can't use window.activeTextEditor to insert the results coming from the shiny app session.

I'm thinking about this too. Sounds like a tricky one.

R/init.R
```diff
@@ -385,6 +395,54 @@ if (interactive() &&
...
+        jsonlite::read_json(response_file)
+      }
+
+     source(file.path(dir_extension, "vsc_r_api_util.R"),
+       local = TRUE
+     )
+     source(file.path(dir_extension, "vsc_r_api.R"),
+       local = TRUE
+     )
+     update_addin_registry(addin_registry)
```
renkun-ken, line 438: Maybe we should call this in `attach()` function in `.vsc`? Then each attach call refreshes the addins so that user could re-attach after installing a new addin to get vscode-R noticed.

MilesMcBain:
> I'm thinking about this too. Sounds like a tricky one.

What about if the extension uses the onDidChangeActiveTextEditor event handler to set a global, which is the last activeTextEditor that is not undefined?

When the addin handlers reach for the activeTextEditor, they use that variable if window.activeTextEditor is `undefined`. I think this more or less makes `window.activeTextEditor` work like the documentation says:

>The currently active editor or undefined. The active editor is the one that currently has focus or, when none has focus, the one that has changed input most recently.

R/init.R
MilesMcBain: Yes and when I refactor this, I must:
* reset the 'warmup' back to 200
* make it so the addin picker can only run if a terminal has been attached.

MilesMcBain:
The tracking the last defined TextEditor idea seems to work okay. Shiny apps are go!

I am still having trouble figuring out a way to be able to launch the addin picker and have that automatically attach an R session, whilst ensuring that addins.json file exists when it is needed.

As far as I can see it really does come down to some kind of spinlock on the "attach" request coming from `.vsc.attach()`. But this spinlock needs to be here: https://github.com/Ikuyadeu/vscode-R/blob/master/src/rTerminal.ts#L100

Happy to hear other ideas. For now I have stopped the picker from launching if `rTerm` is `undefined`.

renkun-ken:
>The tracking the last defined TextEditor idea seems to work okay. Shiny apps are go!

Looks promising!

>I am still having trouble figuring out a way to be able to launch the addin picker and have that automatically attach an R session, whilst ensuring that addins.json file exists when it is needed.

Are you trying to do the following?

1. User calls addin picker
2. Check if there's an R session attached
3. If so, read the `addins.json` and show addin picker.
4. If not, try to attach an R session

If I'm understanding correctly, the following logic makes sense to me.

1. Use  calls addin picker
2. Check if there's an R session attached
3. If so, read the `addins.json` and show addin picker.
4. If not, check `r.alwaysUseActiveTerminal`.
5. If `r.alwaysUseActiveTerminal=false` (default), then run `r.createRTerm` before showing the addin picker
6. If `r.alwaysUseActiveTerminal=true`, then show a message that there's no attached R session to run addin picker.

MilesMcBain:
I think the problem is that if I `await r.createRTerm()` I am just awaiting a 200ms delay. In my local testing that was not enough time to generate the JSON file.

renkun-ken:
>As far as I can see it really does come down to some kind of spinlock on the "attach" request coming from `.vsc.attach()`

Now I could see what you need.

src/vsc_r_api.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  if (!config().get<boolean>('sessionWatcher')) {
+    throw ('{rstudioapi} emulation requires session watcher to be enabled in extension config.');
+  }
+  if (typeof rTerm === 'undefined') {
+    throw ('No active R terminal session, attach one to use RStudio addins.');
+  }
+
+  const activeRTerm = await chooseTerminal();
+
```
renkun-ken, line 163: I'm thinking that we might implement a little function of wait-for-attach and await it here.

The function could return a `Promise` where `sessionDir` is constantly checked or timeout.

MilesMcBain: That's pretty much exactly what I had in mind. 👍

renkun-ken: Currently, we don't reset variables like `pid` and `sessionDir` when session is terminated because if a session is killed we don't know that anyway. Resetting `sessionDir` before await attach seems necessary.

src/vsc_r_api.ts
```diff
@@ -116,6 +124,19 @@ export async function documentSave(id: string) {
   await targetDocument.save();
 }
 
+export async function documentSaveAll() {
+  await workspace.saveAll();
+}
+
+export function projectPath() {
+  const path = typeof workspace.workspaceFolders !== 'undefined' ? 
+    workspace.workspaceFolders[0] : undefined;
```
renkun-ken, line 133: We have support for multi-root workspace (i.e. working with multiple project folders).

A better approach here might be using `workspace.getWorkspaceFolder(document.uri)` where `document` is the last defined active document.

MilesMcBain:
That completes the what's possible with the emulation for the Document API, and I think that's all I plan to do.

Next for me is just working out what needs to get a 'not implemented' message, and some documentation.

MilesMcBain:
@renkun-ken I added this function to restart the R session, I was wondering if it should be exported as a extension command, since I know it's a popular feature with many people in RStudio?

https://github.com/Ikuyadeu/vscode-R/pull/408/files#diff-d33c533ae558f85846acbe177816358aR47

MilesMcBain:
I've added some basic instructions to the README, I was thinking it might be a good idea to make a wiki page that can be linked from there that describes the scope of the addin support and a list of known compatible addins that I've tested.

MilesMcBain:
To fix:
  * the memoised AddinItems array needs to be disposed along with the rTerminal, so that it is actually regenerated on subsequent attaches.
  * In testing `{remedy}` I found there's some fun to be had supporting positions that contain `Inf`.

MilesMcBain:
Here's my proposal for a wiki page:

# RStudio Addin Support 

The VSCode-R extension is compatible with a subset of RStudio Addins via an
`{rstudioapi}` emulation layer. Nearly all of the document inspection and
manipulation API is supported, although with some notable concessions:

  * VSCode always directs addin text output destined for the "active document" to
    the active or last active text editor, and never the active terminal/console.

There are two main ways to launch an addin:

  1. via the addin picker, launched from the command palette as "R: Launch
    RStudio Addin", or via a keybinding to `r.lauchAddinPicker`. 
  2. via a direct
    keybinding to the addin R functions. These can be found by looking in the
    `inst/rstudion/addins.dcf` file of the addin-providing-package's source. For
    example to bind a key to Reprex the current selection, you would configure a
    keybinding like:
```json
    {
        "description": "reprex selection",
        "key": "ctrl+;",
        "command": "r.runCommand",
        "when": "editorTextFocus",
        "args": "reprex::reprex_selection"
    }
```

## Installing addins

Unlike RStudio the VSCode addin registry is regenrated every time a new R
session is attached, not when a new package is installed. You can use
`rstudioapi::restartSession()` to reattach the R session and repopulate
installed addins.

## Shiny Gadgets

Addins that launch Shiny gadgets and call supported functions can work,
although for best results, consider setting: `options(vsc.viewer = "Beside")`
to have the gadget open up alongside the active text editor.

## Known compatible packages

Packages listed here have been verified working in VSCode.

  * `{datapasta}` dev version
  * `{reprex}`
  * `{fnmate}`
  * `{equisse}`
  * `{remedy}`
  * `{AlignAssign}`
  * `{blogdown}`
    - Live preview in VSCode works but viewer colour theme interacts with site colour theme. Choose a light theme for light site preview.
  * `{docthis}`
  * `{sinew}`
    - The interactive oxygen generator polls the active document_context, which is glitchy but can work.
  * `{prefixer}`
  * `{gistfo}`

Pacakges that can be compatible with minimal changes:

  * `{citr}` e.g. [this fork](https://github.com/milesmcbain/citr)

## Implemented functions

`{rstudioapi}` Functions in this section have been reimplemented to work with VSCode:

  *  ` getActiveDocumentContext `
  *  ` getSourceEditorContext `
  *  ` isAvailable `
  *  ` verifyAvailable `
  *  ` insertText `
  *  ` modifyRange `
  *  ` readPreference `
  *  ` readRStudioPreference `
  *  ` hasFun `
  *  ` findFun `
  *  ` showDialog `
  *  ` navigateToFile `
  *  ` setSelectionRanges `
  *  ` setCursorPosition `
  *  ` documentSave `
  *  ` documentId `
  *  ` documentPath `
  *  ` documentSaveAll `
  *  ` documentNew `
  *  ` getActiveProject `
  *  `restartSession`
  *  `viewer`

## Unimplemented functions

Functions not above that need to interact with the IDE to work have been
reassigned to stop with an error:

`This {rstudioapi} function is not currently implemented for VSCode.`


# Developer notes

The best way to check if an IDE API function is supported in a way that is
supported by both RStudio and VSCode is to call `hasFun` or `findFun`. This
will work as expected for all implemented functions above.

`rstudioapi` offers the ability to assert certain version numbers of the IDE
whilst using these functions with a `version_needed` argument. This is not
supported by VSCode at present.

MilesMcBain:
@renkun-ken and @andycraig I think it's ready for review and wider testing now. 

The one outstanding issue is that launching the addin picker does not launch a terminal session, but it does error with an informative message, which I am happy with for now if you are. Either that or we have some kind of lock for attach, as discussed above.

There is also a one line conflict with master which I think should be straight forward to resolve.

Also one other thing I noticed is that browseURL behaves a bit differently on this fork, possibly due to the changes we made for shiny apps @renkun-ken. But what it does mean is that simple browse commands to external sites don't prompt to open externally, and instead open a blank WebViewPanel. You can try this yourself with `browseURL("https://githhub.com")`

MilesMcBain:
Also I realise a 1000 LOC PR is a bit on the extreme side :sweat_smile:  so I would be happy to do a code walk through via Live Share to explain how things hang together. Let me know if that's something that would help and we can find a time.

renkun-ken:
Thanks for your nice work! I'll test and review this PR these days.

>Also one other thing I noticed is that browseURL behaves a bit differently on this fork, possibly due to the changes we made for shiny apps @renkun-ken. But what it does mean is that simple browse commands to external sites don't prompt to open externally, and instead open a blank WebViewPanel.

This is resolved by #406 which is ahead of this PR. So there's nothing to do about it in this PR.

MilesMcBain:
My PR to `{citr}` to support VSCode was merged, so I moved it the supported list in the proposed wiki page above.

src/vsc_r_api.ts
```diff
@@ -0,0 +1,374 @@
...
+let addinQuickPicks: AddinItem[] = undefined;
+
+export async function getAddinPickerItems() {
+
+  if (typeof addinQuickPicks === 'undefined') {
+    const addins: any[] = await fs.readJSON(path.join(sessionDir, 'addins.json'));
+    const addinItems = addins.map((x) => {
+      return {
+        alwaysShow: true,
+        description: x.description,
```
renkun-ken, line 220: It might look better if we use the following:

```
description: `{${x.package}}`,
detail: x.description
```

and the addin picker will look like 

image

renkun-ken:
Also, I notice that `R: Launch RStudio Addin` command only works if the R session is created by `R: Create R terminal` command. However, my main use case is creating R session manually in a tmux window (as described at https://github.com/Ikuyadeu/vscode-R/wiki/R-Session-watcher#advanced-usage-for-self-managed-r-sessions). It would be nice if we support populating the addin list whenever `.vsc.attach()` is called from the session?

src/vsc_r_api.ts
MilesMcBain: Yup I agree.

MilesMcBain:
The addin item list is now purged whenever the request handler for the 'attach' request runs. This feels like a much better spot for it. Does this work with your setup @renkun-ken?

src/vsc_r_api.ts
```diff
@@ -0,0 +1,374 @@
...
+export function purgeAddinPickerItems() {
+  addinQuickPicks = undefined;
+}
+
+export async function launchAddinPicker() {
+
+  if (!config().get<boolean>('sessionWatcher')) {
+    throw ('{rstudioapi} emulation requires session watcher to be enabled in extension config.');
+  }
+  if (typeof rTerm === 'undefined') {
```
renkun-ken, line 242: The addin picker still does not work with self-managed R sessions. It seems the check of `rTerm` here only applies to R sessions created by `R: Create R terminal` command.

We might need to check `sessionDir` instead of `rTerm` to ensure there's an attached session. We might also need to unset `sessionDir` when `rTerm` is closed.

R/init.R
```diff
@@ -385,6 +396,54 @@ if (interactive() &&
         page_viewer = page_viewer
       )
 
+      #vsc-r-api
+      RESPONSE_TIMEOUT <- 30
```
renkun-ken, line 400: Does it look more consistent if we keep using the snake_case here?

Also, waiting for 30 seconds until timeout seems a bit too long? Could we use a more reasonable number which is small but sufficient to indicate there might be a problem, e.g. 5 seconds?

R/vsc_r_api.R
```diff
@@ -0,0 +1,334 @@
...
+}
+
+restart_r_session <- function() {
+    invisible(
+        request_response("restart_r")
+    )
+}
+
+vsc_viewer <- function(url, height = NULL) {
+    # cant bind to this directly because it's not created when the bidning is
```
renkun-ken, line 246: A typo here

R/vsc_r_api.R
```diff
@@ -0,0 +1,334 @@
...
+        "hasColourConsole",
+        "bugReport",
+        "buildToolsCheck",
+        "buildToolsInstall",
+        "buildToolsExec",
+        "dictionariesPath",
+        "userDictionariesPath",
+        "executeCommand",
+        "translateLocalUrl"
+    )
```
renkun-ken, line 334: The mapping and no mapping lists look a bit redundant which duplicate names with different naming conventions.

Does it look better if we dedicate a file in which the original names are used to define mapped functions, and we source the file to an environment on session attach, and replace all functions that exist in both this mapping environment and `rstudioapi` namespace with original names. Then there's no need to maintain such mapping lists.

For example, a dedicated file could look like

```r
getActiveDocumentContext <- function(...) ...
getSourceEditorContext <- function(...) ...

sourceMarkers <- not_yet_implemented
```

src/vsc_r_api.ts
renkun-ken: For example, check if sessionDir exists.

R/init.R
MilesMcBain: I am used to using caps for numeric constants but if that's not a style you use no worries. Yeah 30 seconds is definitely too long, 5 sounds good.

R/vsc_r_api.R
MilesMcBain: This could work although things get a bit tricky to make robust.

* I am not replacing all functions in the `rstudioapi` namespace since there are some that remain useful, like those that deal with constructing/testing positions and ranges. 
* The functions that are to be assigned into the namespace currently share an environment with a bunch of helper functions, those shouldn't be assigned into the `rstudioapi` namespace - most likely they wouldn't because they don't match with a function in the namespace, but there is a small chance they could? I certainly haven't checked all unexported methods.
* So to be robust to collision, we would have to prep two environments: A child environment to contain the methods to be iterated over for assignment into API namespace, and a parent environment, which is not iterated over, which contains the helper methods those API replacements need to function.

I didn't think too hard about it, but having the mappings seemed a nice simple way to ensure robustness.

I thought snake case for the VSC-R api was convenient, since although some functions are a 1:1 mapping with the `rstudioapi`, others are not, and it would save some confusion.

MilesMcBain: I guess an advantage of your proprosal would be the the api functions do not fill up the `.vsc` environment, making that a bit easier to use by inspection. Not sure how important that is?

MilesMcBain: Okay I've refactored along these lines in cd1fb08.

This looks a little spooky to me: https://github.com/Ikuyadeu/vscode-R/pull/408/files#diff-2711d8988b4edf6562a0668d7e873c13R435

But it gets the job done.

MilesMcBain: There are now only two potential collisions:

* .vsc_not_yet_implemented
* .vsc_document_context

and those prefixes should be good enough to guard against that?

src/vsc_r_api.ts
MilesMcBain: tried this in 9b99786, seems to work right for me, but I don't have your advanced setup. I'd be keen to see what that looks like btw!

R/init.R
```diff
@@ -423,7 +423,7 @@ if (interactive() &&
         request(command, ...)
         wait_start <- Sys.time()
         while (!get_response_lock()) {
-          if ((Sys.time() - wait_start) > RESPONSE_TIMEOUT)
+          if ((Sys.time() - wait_start) > response_timeout)
             stop("Did not receive a response from VSCode-R API within ",
                   RESPONSE_TIMEOUT, " seconds.")
```
renkun-ken, line 428: Missing this `RESPONSE_TIMEOUT`?

src/vsc_r_api.ts
```diff
@@ -239,7 +239,7 @@ export async function launchAddinPicker() {
   if (!config().get<boolean>('sessionWatcher')) {
     throw ('{rstudioapi} emulation requires session watcher to be enabled in extension config.');
   }
-  if (typeof rTerm === 'undefined') {
+  if (typeof sessionDir === 'undefined') {
```
renkun-ken, line 242: Maybe checking if `sessionDir` exists in filesystem is more robust? If self-managed terminal is killed, then the `sessionDir` still is defined but no longer exists.

R/init.R
MilesMcBain: :grimacing:

src/vsc_r_api.ts
MilesMcBain: So if a self-managed terminal is killed, the `deleteTerminal` handler is run, which runs `removeSessionFiles`?

Inside `removeSessionFiles` I added a line to unset the `sessionDir`, so I think that should be equivalent?

Unless the handler isn't run. If the self-managed session can terminate without notifying VSCode to call the handler, I can't see a way to know for sure a session is attached and maybe the design needs to be changed to support self-managed sessions.

renkun-ken: By "a self-managed terminal is killed", I mean an R session started manually in a vscode terminal (from a vscode terminal directly, or a tmux window in the vscode terminal). In either case, `deleteTerminal` won't be called since the vscode terminal is not closed but the R session is killed.

R/vsc_r_api.R
```diff
@@ -0,0 +1,290 @@
...
+}
+
+getSourceEditorContext <- getActiveDocumentContext
+
+verifyAvailable <- function(version_needed = NULL) {
+    if (is.null(version_needed)) TRUE else FALSE
+}
+
+isAvailable <- function(version_needed = NULL, child_ok) {
+    verify_available(version_needed)
```
renkun-ken, line 23: Looks like some old calls are not renamed?

R/vsc_r_api.R
```diff
@@ -0,0 +1,290 @@
...
+}
+
+readRStudioPreference <- readPreference
+
+hasFun <- function(name, version_needed = NULL, ...) {
+    if (!is.null(version_needed)) {
+        return(FALSE)
+    }
+
+    exists(x = name, envir = as.environment(rstudio_vsc_mapping), ...)
```
renkun-ken, line 96: Should find directly rather than in non-existing `rstudio_vsc_mapping`?

R/vsc_r_api.R
```diff
@@ -0,0 +1,290 @@
...
+}
+
+getFun <- function(name, version_needed = NULL, ...) {
+    if (!is.null(version_needed)) {
+        stop("VSCode does not support used of 'version_needed'.")
+    }
+
+    get(x = name, envir = as.environment(rstudio_vsc_mapping), ...)
```
renkun-ken, line 104: Same with https://github.com/Ikuyadeu/vscode-R/pull/408/files#r502975097

src/vsc_r_api.ts
renkun-ken: In other words, `deleteTerminal` is only called when user closes the VSCode terminal, rather than the R session in the terminal. If the R session is self-managed, then terminating the R session will make the VSCode terminal go back into e.g. `bash`. The R session is gone but the VSCode terminal is still there and `deleteTerminal` won't be called in this case.

renkun-ken: If the user-managed R session quits on R error or `q()`, then the session `tempdir()` which contains `sessionDir` will be removed. In this case, `sessionDir` is still defined but the directory it points to no longer exists.

Checking if `sessionDir` exists in filesystem will not work if the user-managed R session is killed or crashes on C level, then the `tempdir()` won't be removed.

renkun-ken:
I tried running "Style active file" from `{styler}` on an active document but the following error occurs:

```
> styler:::style_active_file()                                                                                                    
Using style transformers `styler::tidyverse_style()`
Error in rstudioapi::setCursorPosition(context$selection[[1]]$range) : 
  Expecting a document_position object. Got something else.
Backtrace:
1: stop("Expecting a document_position object. Got something else.")
2: rstudioapi::setCursorPosition(context$selection[[1]]$range)
3: styler:::style_active_file()
```

R/vsc_r_api.R
MilesMcBain: doh.... yeah so this was another useful feature of the explicit mapping. I was only reporting true for things that had a valid implementation

MilesMcBain: I think this is a useful feature. To get it back I'll need to move all the not implemented functions to a separate file which gets sourced into a different environment again. I am really questioning the value of this change now.

MilesMcBain:
> I tried running "Style active file" from {styler} on an active document but the following error occurs:

Oh. It seems that the `rstudioapi` supports either positions or ranges to either setter. Removing checks is easy.

R/vsc_r_api.R
renkun-ken: Yes, the explicit mapping could be useful here. But there's no need to make things complicated. The purpose of this is to make the code easier to maintain, therefore a simple fix for this could be

```r
.vsc.rstudioapi_env <- environment()

hasFun <- function(name, version_needed = NULL, ...) {
    if (!is.null(version_needed)) {
        return(FALSE)
    }

    obj <- .vsc.rstudioapi_env[[name]]
    is.function(obj) && !identical(obj, .vsc_not_yet_implemented)
}

getFun <- function(name, version_needed = NULL, ...) {
    if (!is.null(version_needed)) {
        stop("VSCode does not support used of 'version_needed'.")
    }

    get(name, .vsc.rstudioapi_env, mode = "function", inherits = FALSE)
}
```

MilesMcBain: Yeah okay that's not too bad. :+1:

MilesMcBain: `getFun` needs to call `hasFun` also, and error if `False`, since it's supposed to error if it cannot get the function.

renkun-ken: Yes, sorry for missing that `getFun` only works with mapped functions.

MilesMcBain: all good that comment was more of a reminder to myself. Looks good.

renkun-ken:
Error like the following occurs:

```
> styler:::style_active_file()                                                                                                                                                                            
Using style transformers `styler::tidyverse_style()`
Error in as.environment(match("tools:rstudio", search())) : 
  invalid 'pos' argument
Backtrace:
1: toolsEnv()
2: exists(toolsName("versionInfo"), envir = toolsEnv())
3: usingTools()
4: callFun("setSelectionRanges", position, id)
5: rstudioapi::setCursorPosition(context$selection[[1]]$range)
6: styler:::style_active_file()
```

R/init.R
```diff
@@ -385,6 +396,55 @@ if (interactive() &&
         page_viewer = page_viewer
       )
 
+      #vsc-r-api
+      response_timeout <- 5
+      response_lock_file <- file.path(dir_extension, "response.lock")
+      response_file <- file.path(dir_extension, "response.log")
```
renkun-ken, line 402: Putting the response file here might be problematic if user has multiple vscode instances, all sharing the same response file.

src/vsc_r_api.ts
MilesMcBain: Okay thanks for the explanation! I have changed the test to look for the session directory. It seems that's about the best we can do easily? 

The consequences do not seem serious anyway, since if you have had a session terminate at the C level, and then run the addin picker without reattaching a new session it will probably still work? Since it will call chooseTerminal which will not find an R terminal and will create one to run the addin. So worst case I think is that you are looking at an out of date addin list?

R/init.R
MilesMcBain: oh I thought we changed that to the session dir.... :thinking:

MilesMcBain: Oh dear: https://github.com/Ikuyadeu/vscode-R/pull/408/files#diff-f3fe9c4fe673e665c8df5914c3b9c9a0R65

half right.

renkun-ken: The file should be put to the reader, R session dir in this case, whereas the request file is read by vscode-R extension.

MilesMcBain: ...how was that even working... at all?

MilesMcBain: Oh wait I see. Watcher dir, not session dir?

MilesMcBain: So I think I have thought about this a while ago. It's kind of okay, since all responses are synchronous and addins are interactive dev tools?

Pretty much impossible for a human to launch two in a way that will create a problem?

MilesMcBain: Oh wait but I have seen an addin that polled the active document context in `{sinew}`. Boo.

MilesMcBain: So maybe this need some rejigging on the request side of things too. If `request.wd` was the session dir, we could know to write back to a response file in that session dir?

MilesMcBain: ha but of course that can't work for the 'attach' handler because we don't know our session dir.

So maybe we leave that as is, but all calls to `request_response` from the R side include the session dir? which the response uses to write back to?

MilesMcBain: Ok I think I have successfully moved the response files to the session dir. That was good actually things are a bit simpler now, and I think it shook out a bug with the file setup.

MilesMcBain:
Confirmed `{styler}` is working. But there's been enough rejigging going on I think I'll need to run right through all the supported addins again after we're done with that.

R/init.R
```diff
@@ -385,6 +396,55 @@ if (interactive() &&
...
+      get_response_lock <- function() {
+        lock_time_stamp <- get_response_timestamp()
+        if (isTRUE(lock_time_stamp != response_time_stamp)) {
+          response_time_stamp <<- lock_time_stamp
+          TRUE
+        } else FALSE
+      }
+
+      request_response <- function(command, ...) {
+        request(command, ..., sd__ = dir_session)
```
renkun-ken, line 423: I guess such quirky argument name `sd__` is used to avoid collision of request arguments. However, we already have `time`, `pid`, `wd`, `command` fields in a typical request, and fields like `id`, `file` are all common names to make it ambiguous to distinguish which are rstudioapi arguments. I think a better way to do this is to capture these arguments in `list(...)`, i.e.

```r
request(command, args = list(...), sd = dir_session)
```

MilesMcBain: Ha I was going to raise an issue to make `wd__` `pid__` etc... but this is better.

MilesMcBain: So your fix is pretty nimble and I do like it, but do you think it is strange that on the handler side in session.ts, for some requests you pull out the arguments with `request.thing`, and others it's `request.arg.thing`?

MilesMcBain: For example:

```
                case 'dataview': {
                    showDataView(request.source,
                        request.type, request.title, request.file, request.viewer);
                    break;
                }
```

MilesMcBain: should these be `request.arg.type`?

renkun-ken: In fact, I was thinking that the rstudioapi emulation functions should be something like sub-commands. The following call reflects my intention more clearly:

```r
request("rstudioapi", action = "document_save_all", args = list(...), sd = dir_session)
```

Existing commands such as `attach` and `dataview` have known defined arguments. It might make sense that `session.ts` should dispatch commands to the `rstudioapi` level, and those sub-commands should be dispatched in `vsc_r_api.ts`.

MilesMcBain: Yeah this makes sense, I was to thinking about doing this from the very start.

MilesMcBain: Okay that's done.

MilesMcBain:
I am thinking I will rename the files associated with this PR:
  * `vsc_r_api.ts` -> `rstudioapi.ts`
  * `vsc_r_api.R` -> `rstudioapi.R`
  * `vsc_r_api_util.R -> `rstudioapi_util.R`

They are very specific to the RStudio API.

MilesMcBain:
Now looking at the CI failures.

Looks like 1 test failure is occurring because of runTextInTerm changing on master. I could rebase but that would rewrite all my commit shas and make this thread a bit of a mess. I might leave that til the very last thing.

Not sure what to do about lintr, it's going to keep complaining because there's code in different files and it's not a package.

src/rstudioapi.ts
```diff
@@ -0,0 +1,434 @@
...
+                    await writeSuccessResponse(sd);
+                    break;
+                }
+                case 'document_save_all': {
+                    await documentSaveAll();
+                    await writeSuccessResponse(sd);
+                    break;
+                }
+                case 'get_project_path': {
+                    await writeResponse(await projectPath(), sd);
```
renkun-ken, line 57: `projectPath` does not return a `Promise`. Not sure why the linter does not complain.

renkun-ken:
>Looks like 1 test failure is occurring because of runTextInTerm changing on master. I could rebase but that would rewrite all my commit shas and make this thread a bit of a mess. I might leave that til the very last thing.

No need to rebase.

>Not sure what to do about lintr, it's going to keep complaining because there's code in different files and it's not a package.

We could disable `object_usage_linter` in future to remove such false alarms.

renkun-ken:
Would you like to add something to the README so that users will know this incredible new feature? For example, a new sub-section under [R Session watcher](https://github.com/Ikuyadeu/vscode-R#r-session-watcher-experimental) before [how to disable it](https://github.com/Ikuyadeu/vscode-R#how-to-disable-it)?

MilesMcBain:
I did add some stuff to the README, but the image proabably needs an update now. I'll move the section to where you suggest.

MilesMcBain:
There's a broken link in the README which is intended to link to that wiki page I put in this thread (some time ago now). I can't create wiki pages, so someone with the permissions will need to do that step please.

renkun-ken:
>There's a broken link in the README which is intended to link to that wiki page I put in this thread (some time ago now). I can't create wiki pages, so someone with the permissions will need to do that step please.

I create a wiki page for [RStudio addin support](https://github.com/Ikuyadeu/vscode-R/wiki/RStudio-addin-support). You could use this link now.

@andycraig @Ikuyadeu This PR looks good to me. Do you have any suggestions? Once this PR is merged, I'd suggest that we add @MilesMcBain as a collaborator to maintain and improve this nice feature and its documentation in wiki.

MilesMcBain:
Thanks @renkun-ken, though this is what I had in mind for the content of that wiki page: https://github.com/Ikuyadeu/vscode-R/pull/408#issuecomment-703251259

Just in case you didn't see it.

I am happy to stick around and maintain this work if that is okay by the others. 👍

renkun-ken:
@MilesMcBain Sorry didn't notice the comment. Wiki page updated.

src/rstudioapi.ts
MilesMcBain: my linter was actually flagging locally but not in the 'Problems' pane for some reason. Weird. Fixed!

renkun-ken (approved):
LGTM

Ikuyadeu:
Great works!! LGTM too.
It can be merged after all tests are passed.

Also, I send the request @MilesMcBain for the collaborator.
I hope that you're interested to attend with us.

renkun-ken:
The build fails with the following error:

```
ERROR in /github/workspace/src/rstudioapi.ts
./src/rstudioapi.ts
[tsl] ERROR in /github/workspace/src/rstudioapi.ts(320,38)
      TS2554: Expected 1 arguments, but got 2.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! r@1.4.6 vscode:prepublish: `webpack --mode production`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the r@1.4.6 vscode:prepublish script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
``` 

but I don't see anything incorrect in this call

```typescript
runTextInTerm(activeRTerm, addinSelection.package + ':::' + addinSelection.binding + '()');
```

and the build and lint produce no error locally.

@Ikuyadeu Do you have any idea why this could occur?

MilesMcBain:
This is what I was talking about earlier, on master the signature for runTextInTerm has changed?

MilesMcBain:
also I think I see a typo to fix!

renkun-ken:
>This is what I was talking about earlier, on master the signature for runTextInTerm has changed?

Yes, on master the function call only accepts the call string. No need to specify terminal.

I guess we need to rebase or merge master to incorporate this change accordingly.

Ikuyadeu:
@MilesMcBain This changes looks useful to apply that 
https://github.com/Ikuyadeu/vscode-R/commit/6c9c34f66c6938226e7b052be9994c2188eb6643

MilesMcBain:
@Ikuyadeu oh you think cherry pick that commit to this branch? That would make this thread less messy if it works. It probably will.

Ikuyadeu:
@MilesMcBain Sorry, my comment was misleading. It can be useful for just referring change.
I think rebasing or merging is a safe way like @renkun-ken suggesting.

MilesMcBain:
I tried the cherry pick idea and it seemed to not wreck anything locally. But I guess there is a danger that if he developed on top of that I will revert that work.

MilesMcBain:
Okay I will ditch that and rebase.

MilesMcBain:
Okay I think the rebase was successful. Sorry about the mess. This is only the 2nd time I've done that. Probably what I should have done was squashed everything into one new commit? I'll know for next time. :grimacing:

Ikuyadeu:
Perfect! I'll release a new version this week.

renkun-ken:
Cheers!

Ikuyadeu:
> Probably what I should have done was squashed everything into one new commit? 

I think, just change a line https://github.com/Ikuyadeu/vscode-R/pull/408/files#diff-8b19336fd34c052a83b852dc7aeac4b2a2b56f17c783c2480ca6948ae648330fR320 was the easiest change.
However, we can squash the whole of a pull request when commits are large.

andycraig:
@MilesMcBain Sorry, work has been super busy lately and I fell off the face of the planet for a bit.

Thank you very much for this huge piece of work. There was a lot more required for this than I'd assumed when I created that Issue about supporting these addins. Thank you so much for taking it on!

This is a massive new feature that people (including myself of course!) are really going to enjoy and appreciate.

Thank you again!

Issue #421, opened 2020-10-15
Issues with rstudioapi emulation


MilesMcBain:
A couple of things I noticed when the update landed on my work PC:

* `tidyverse` calls the rstudioapi to check the version which errors, stopping it from being attached.
* For some reason the R building of the addin list fails which interferes with `.vsc.attach`.

@Ikuyadeu @renkun-ken `tidyverse` is a bit of a showstopper. The second issue might be to do with my work PC library setup which is a bit unusual. I'll so some debugging on it now.

Is it possible to roll back to previous release?

dcangst:
I have the same issue. Any pointers on how to fix/downgrade?

(love the new features though! Thanks!)

MilesMcBain:
Yes sorry @dcangst, you can install a vsix file from a recently tagged release. Or if you want a very recent build you might try the one here:

https://github.com/Ikuyadeu/vscode-R/actions/runs/300521275

Instructions on manual vsix install:
https://code.visualstudio.com/docs/editor/extension-gallery#:~:text=Install%20from%20a%20VSIX%23,vsix%20file.

dcangst:
no worries! thanks for the quick reply!!

MilesMcBain:
I've figured out the second issue. This is to do with malformed addins.dcf files. In this case from `shrtcts` on my work machine. I can do some filtering to protect against it and perhaps put the whole parsing thing in a try-catch so that it doesn't stop the session being attached properly.

renkun-ken:
I guess we need to do a quick fix to opt-in this feature using `options(vsc.rstudioapi = TRUE)`.

By default, it is not enabled and rstudioapi-related code will not be sourced?

@MilesMcBain @Ikuyadeu What do you think?

MilesMcBain:
I think it's a good idea. :+1:

renkun-ken:
@MilesMcBain I don't quite get what's not working.

<img width="1185" alt="image" src="https://user-images.githubusercontent.com/4662568/96118469-1fb01100-0f1e-11eb-9822-661455c5db7d.png">

`library(tidyverse)` seems to work normally?

dcangst:
this is what I get when running the master branch using 'f5':
<img width="1311" alt="Screenshot 2020-10-15 at 13 43 51" src="https://user-images.githubusercontent.com/7902555/96118965-90d6d080-0eec-11eb-9d48-24e20a273191.png">

MilesMcBain:
[...]

PR #422, opened 2020-10-15
Fix issues in rstudioapi emulation

MilesMcBain:
Resolves the issues discussed in #421:

1. If an `addins.dcf` file from a package is malformed it could cause an error when trying to write the JSON to drive the addin picker. This error could cause the `vsc.attach` function to fail to run fully. This was not very common. I found only one example in `{shrcts}`. 

I added extra checks to look for malformed files, and to only select the first few columns, assuming those contain the information I want to display. The whole `update_addins_registry` call is wrapped in a try-catch so that if it throws an error it is converted to a message about the failure and the consequences. `vsc.attach` will complete otherwise normally.

2. On Windows `library(tidyverse)` fails due to the `cli` package trying to interrogate the version of RStudio if it is available. I have put in implementations for `getVersion` and `versionInfo` that should allow this to work in the short term.

The plan is to discuss this with RStudio to see if we can get something less fragile.

Finally, to use the RStudio API emulation the user must set `options(vsc.rstudioapi = TRUE)` in the .Rprofile. I have put documentation about this in the README and will add it to the wiki.

Sorry for the hassles!

### DIFF
--- a/R/rstudioapi_util.R
+++ b/R/rstudioapi_util.R
@@ -7,7 +7,9 @@ rstudioapi_enabled <- function() {
 }
 
 rstudioapi_patch_hook <- function(api_env) {
-    if (!rstudioapi_enabled()) return(NULL)
+    if (!rstudioapi_enabled()) {
+        return(NULL)
+    }
 
     patch_rstudioapi_fn <-
         function(old, new) {
@@ -197,6 +199,10 @@ normalise_text_arg <- function(text, location_length) {
 }
 
 update_addin_registry <- function(addin_registry) {
+    if (!rstudioapi_enabled()) {
+        return(NULL)
+    }
+
     pkgs <- .packages(all.available = TRUE)
     addin_files <- vapply(pkgs, function(pkg) {
         system.file("rstudio/addins.dcf", package = pkg)
@@ -213,21 +219,38 @@ update_addin_registry <- function(addin_registry) {
                         "interactive",
                         "package"
                     )
-                addin_description <-
-                    as.data.frame(read.dcf(package_dcf),
-                        stringsAsFactors = FALSE
+                description_result <-
+                    tryCatch(
+                        {
+                            addin_description <-
+                                as.data.frame(read.dcf(package_dcf),
+                                    stringsAsFactors = FALSE
+                                )
+
+                            if (ncol(addin_description) < 4) {
+                                NULL
+                            }
+                            ## if less than 4 columns it's malformed
+                            ## a NULL will be ignored in the rbind
+
+                            addin_description$package <- package
+                            names(addin_description) <- addin_description_names
+
+                            addin_description[, addin_description_names]
+                            ## this filters out any extra columns
+                        },
+                        error = function(cond) {
+                            message(
+                                "addins.dcf file for ", package,
+                                " could not be read from R library. ",
+                                "The RStudio addin picker will not contain it's addins"
+                            )
+
+                            NULL
+                        }
                     )
 
-                if (ncol(addin_description) < 4) {
-                    NULL
-                }
-                ## if less than 4 columns it's malformed
-                ## a NULL will be ignored in the rbind
-
-                addin_description$package <- package
-                names(addin_description) <- addin_description_names
-                addin_description[, addin_description_names]
-                ## this filters out any extra columns
+                description_result
             },
             names(addin_files),
             addin_files,
@@ -240,21 +263,6 @@ update_addin_registry <- function(addin_registry) {
         )
 
     jsonlite::write_json(addin_descriptions_flat, addin_registry, pretty = TRUE)
-}
-
-update_addin_registry_safely <- function(addin_registry) {
-    if (!rstudioapi_enabled()) return(NULL)
-
-    tryCatch(update_addin_registry(addin_registry),
-        error = function(cond) {
-            message(
-                "List of installed addins could not",
-                "be read from R library. ",
-                "Possibly due to malformed addins.dcf files. ",
-                "The RStudio addin picker will not function."
-            )
-        }
-    )
 }
 
 namespace_has <- function(obj, namespace) {
--- a/R/rstudioapi_util.R
+++ b/R/rstudioapi_util.R
@@ -2,7 +2,13 @@ rstudioapi_call <- function(action, ...) {
     request_response("rstudioapi", action = action, args = list(...))
 }
 
+rstudioapi_enabled <- function() {
+    isTRUE(getOption("vsc.rstudioapi"))
+}
+
 rstudioapi_patch_hook <- function(api_env) {
+    if (!rstudioapi_enabled()) return(NULL)
+
     patch_rstudioapi_fn <-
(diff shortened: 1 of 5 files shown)


R/rstudioapi_util.R
```diff
@@ -226,7 +242,22 @@ update_addin_registry <- function(addin_registry) {
     jsonlite::write_json(addin_descriptions_flat, addin_registry, pretty = TRUE)
 }
 
+update_addin_registry_safely <- function(addin_registry) {
+    if (!rstudioapi_enabled()) return(NULL)
+
+    tryCatch(update_addin_registry(addin_registry),
```
renkun-ken, line 248: I suggest that the tryCatch be used per `read.dcf` so that a single malformed file won't make the entire feature broken.

renkun-ken: Just notice that you check the number fields at https://github.com/Ikuyadeu/vscode-R/pull/422/files#diff-11107abea790174063b25c459ad87e095af4cab1fbf98bfe33338337901a5939R221-R223. If the malformed file won't cause `read.dcf` to stop, then a single malformed won't make the feature broken?

MilesMcBain: Yeah 
  * if it's malformed short in columns - return null and it will be filtered out by rbind
  * if it's malformed long - take the first 4 (and append 1)
  * if there's something that causes an error like it's NULL and I can't index it, error

MilesMcBain: I like your idea of moving the tryCatch per file read/parse

MilesMcBain: I am just worried they're could be some error case on the bind that I haven't thought of

MilesMcBain: It seems that even if all are NULL rbind and write_json will work. So let's do it your way, that's better because I can report the specific file.

R/rstudioapi_util.R
```diff
@@ -2,7 +2,13 @@ rstudioapi_call <- function(action, ...) {
     request_response("rstudioapi", action = action, args = list(...))
 }
 
+rstudioapi_enabled <- function() {
+    isTRUE(getOption("vsc.rstudioapi"))
```
renkun-ken, line 6: Does it make sense that the option check be moved to `init.R` to enclose https://github.com/Ikuyadeu/vscode-R/pull/422/files#diff-1e8c7b94d7dc2f0d70c4e457124889b5637ba42993015643e5d80be52e3ec912R432-R478 so that if `vsc.rstudioapi` is not `TRUE` then the entire rstudioapi-related code won't be excuted.

MilesMcBain: I think so. Still needed in the updateAddin registry call, because that's called in .vsc.attach or we can wrap around it in there?

MilesMcBain: Although the response mechanism does not have to only be used by the rstudioapi?

MilesMcBain: I guess we can move the guard if we want to use it for something else.

renkun-ken: `.vsc.attach` should check `rstudioapi_enabled()` before updating addin registry.

We could single out the response mechanism in the future if we use it to something else.

MilesMcBain: okay done

MilesMcBain: nope sorry missed a change!

MilesMcBain: Ok now done

renkun-ken (approved):
LGTM

renkun-ken:
@Ikuyadeu We might need a release once this PR is merged and notice users that `options(vsc.rstudioapi = TRUE)` should be used to enable rstudioapi emulation in the changelog.

When the feature become more stable, then we could make it `TRUE` by default.

renkun-ken:
Maybe off the topic but https://github.com/Ikuyadeu/vscode-R/blob/master/src/rstudioapi.ts#L306 could be removed.

MilesMcBain:
Still not sure if this is working properly on Windows. I just tried to install the vsix and I am getting issues. Not sure if I installed the right thing.

MilesMcBain:
Oh yeah looks like I installed an old one. Let me confirm.

MilesMcBain:
No. I'm seeing something very strange.

Using the vsix file built by the CI:
* On Linux, no issues.
* On my Windows work computer: 

```
[ins] r$> .vsc.attach()
Error in .vsc.attach() : object 'rstudioapi_util_env' not found
```

Which is totally weird, since I can do:

```
[ins] r$> vsc_attach_env <- environment(.vsc.attach)

[ins] r$> vsc_attach_env$rstudioapi_enabled()
[1] TRUE
```

But if no `rstudioapi` stuff is in that environment, and no lockfiles etc created on in tmp folder.

It's really spinning me out.

MilesMcBain:
OK OK OK. I got it. It was an issue with my .Rprofile which had some old stuff in there, manually sourcing init.R.

ALL GOOD. 😌

renkun-ken:
I tried the latest commit on macOS and Windows 10 VM and everything looks good so far.

almaceleste:
> `Error in .vsc.attach() : object 'rstudioapi_util_env' not found`
> I got it. It was an issue with my .Rprofile which had some old stuff in there, manually sourcing init.R.

@MilesMcBain, I have the same error message on my Ubuntu. 
could you explain in more detail, how you solve this problem?

MilesMcBain:
@almaceleste `.vsc.attach()` is called at the end of an init.R script that runs when an R terminal is created. I was getting this error because I had call to `vsc.attatch()` in my `.Rprofile`, which I think was a hangover from an old way that the extension used to work.

Check for any calls to this function in any of your session startup code and remove.

almaceleste:
@MilesMcBain, I have no `vsc.attach()` in my `.Rprofile`, only this code:  
```R
source(
    file.path(
        Sys.getenv(
            if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"
        ), ".vscode-R", "init.R"
    )
)
options(vsc.rstudioapi = TRUE)
```

renkun-ken:
@almaceleste `options(vsc.rstudioapi = TRUE)` should run before `source()`, or otherwise it won't work.

almaceleste:
@renkun-ken, thanx, it helps me. 
now I do not get errors when running `Create R terminal` and `Launch RStudio Addin` commands, and I get RStudio Addins list successfully. 
I think this should be described more obviously in the instruction and in the wiki

renkun-ken:
@almaceleste Good to know it works for you now.

>I think this should be described more obviously in the instruction and in the wiki

Definitely. @MilesMcBain Would you mind updating docs so that users know how to enable it exactly? (e.g. https://github.com/Ikuyadeu/vscode-R/pull/422#issuecomment-723541397 won't work)

MilesMcBain:
I added a bit extra to the wiki: https://github.com/Ikuyadeu/vscode-R/wiki/RStudio-addin-support#enabling-rstudio-addin-support

Issue #424, opened 2020-10-16
Change name of `browser` function in init.R?


MilesMcBain:
Somewhat surprisingly having a function called `browser` on the search path interferes with R's command line debugging tools.

`base:: browser()` can be used, but recover mode using `options(error = recover)` does not work - I find this one really handy for jumping to the source of errors in dense code like init.R.

I'd suggest renaming the internal `browser` for future debugging convenience. Thoughts?

renkun-ken:
Yes, it should be renamed. I'll do a PR for this.

PR #425, opened 2020-10-16
Rename init functions

renkun-ken:

As #424 points out, `init.R` uses `browser` as a local function name. Although it does not go into user's global environment, it does make it a less convenient for developers to use `browser()` directly to debug the code in the scope.

This PR renames `browser` as well as several other functions so that they are consistently named like `do_something`.

Attach a session and test if all related features still work:

```r
# show_dataview
View(mtcars)

# show_viewer
DT::datatable(mtcars)

# show_browser
shiny::runExample("01_hello")

# show_page_viewer
profvis::profvis({n <- 1e6; x <- rnorm(n); y <- rnorm(n); m <- lm(y ~ x)})
```

I've tested this on macOS and it works as expected.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -257,7 +257,8 @@ if (interactive() &&
           list(columns = columns, data = data)
         }
 
-        dataview <- function(x, title, viewer = getOption("vsc.view", "Two")) {
+        show_dataview <- function(x, title,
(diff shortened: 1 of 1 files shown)


MilesMcBain (approved):
Tested the examples and tested debugging using broswer and recover all good. New names make sense.

LGTM

PR #430, opened 2020-10-26
fix and enhance navigateToFile

MilesMcBain:
- fix bug in navigateToFile in rstudioapi emulation where positions were being converted to VSCode coordinates twice leading to incorrect position being navigated to (row and column -1).
- add 'reveal' call so that position row/column is scrolled to in file after opening.

check that in an attached VSCode R session `rstudioapi::navigateToFile(file, row, colum)` opens a text editor to path in `file` and scrolls to show the cursor at position line = `row`, column = `column`. Does not need to be an .R file.

For example in the VSCode-R project, with an attached R session, `rstudioapi::navigateToFile("src/rstudioapi.ts", 154, 23)` should navigate to the Typescript definition for `navigateToFile`.

### DIFF
--- a/src/rstudioapi.ts
+++ b/src/rstudioapi.ts
@@ -14,65 +14,65 @@ let lastActiveTextEditor: TextEditor;
 export async function dispatchRStudioAPICall(action: string, args: any, sd: string) {
 
   switch (action) {
-case 'active_editor_context': {
-                    await writeResponse(await activeEditorContext(), sd);
-                    break;
-                }
-                case 'insert_or_modify_text': {
-                    await insertOrModifyText(args.query, args.id);
-                    await writeSuccessResponse(sd);
-                    break;
-                }
-                case 'replace_text_in_current_selection': {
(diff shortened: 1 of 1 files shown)


src/rstudioapi.ts
```diff
@@ -417,7 +416,7 @@ async function reuseOrCreateEditor(targetDocument: TextDocument) {
 
 
   const matchingTextEditors = KnownEditors.filter((editor) =>
-      editor.document.uri.toString === targetDocument.uri.toString);
+    editor.document.uri.toString === targetDocument.uri.toString);
```
renkun-ken, line 419: Should be `editor.document.uri === targetDocument.uri`? or both `toString()`.

MilesMcBain: Oh wow. That's incredible.That was just a style change, but what a find.

MilesMcBain: I think toString works best as `uri` is listish data structure.

renkun-ken (approved):
Tested and it works well.

LGTM

Issue #432, opened 2020-10-29
Improve display of help pages


ManuelHentschel:
The integrated [viewer](https://github.com/Ikuyadeu/vscode-R/blob/63edf49d33cd2eb4f7a04f8b0c452a9e1c6ed463/src/session.ts#L207:L228) for e.g. help pages is already a great improvement over external browsers or simple text.

In [vscode-R-help](https://github.com/ManuelHentschel/vscode-R-help) I've tried to improve the help pages a bit further by reading the content of the html help files (found in the installation directory of packages), and directly rendering the html in a vscode.Webview. This makes it possible to modify the html to e.g.:
* apply custom css (match the color theme of the editor)
* apply syntax highlighting
* enable further customization (foldable sections, "click-to-run" examples, ...)

Instead of reading the html files directly, it would also be possible to start a normal http help server in R and forward all queries there to get the html (instead of doing so "indirectly" in an iframe).

Is this something you would consider implementing here?

I'd be happy to keep working on this, but could probably use some help, since I'm not very familiar with the codebase of this extension and don't have much experience with html/css either... :)

Ikuyadeu:
@ManuelHentschel Thank you for your great implementation idea!
It looks helpful to not only display, but it also makes it easy to maintain the viewer source code.
If you submit a pull request here, we will review and merge that.

ManuelHentschel:
The following improvement ideas came up while working on the help panel, but are not implemented yet:
- [ ] Include images etc. (e.g. the R logo on the main doc page)
- [ ] Handle links to PDF files (and other file types?)
- [ ] Improve the usability of `r.showHelp`: implement fuzzy matching, autocomplete, simulation of `?` in the command palette
- [ ] Implement `??`
- [ ] Adapt the style file to different themes, maybe let the user choose a style file

Anyone interested in working on this or further ideas are of course highly welcome! :)

PR #433, opened 2020-10-31
WIP: Help view

ManuelHentschel:
As mentioned in #432 I tried to integrate [vscode-R-help](https://github.com/ManuelHentschel/vscode-R-help) here.
The main functionality seems to be working, simply use `help(...)` in an R Terminal (opened by the extension!), or the commands `r.showHelp`, `r.showDoc` in VS Code's command palette.

[Here](https://github.com/ManuelHentschel/vscode-R/blob/c21e134dce6bf158d979246c6222e1157e7dbab0/src/extension.ts#L58) it can be configured, whether to look for help files using a custom function or to look for help files by querying an R instance running `help.start()`. These should mostly behave the same, but there are different problems/difficulties with each, which need to be figured out.

Remaining issues/questions include:
- [x] No clean overwrite of `help()`, since the tools from the extension are attached before `utils` (Implemented as proposed [here](https://github.com/Ikuyadeu/vscode-R/pull/433#issuecomment-721428581))
- [x] Reset scroll position of webview when opening a new link (should work now)
- [x] Repair webpack config (there were some problems on my machine, so I disabled it altogether)
- [x] Use extra config to specify path to normal R executable (in case radian is used in r.rterm.XXX)
- [x] Improve performance when there are many links on a page (since each link is modified individually when loading a page) (Should work ok now, only notable on _very_ long documents)
- [x] Preserve hyperlinks in (styled) code sections
- [x] Fix redirect-pages when using `'Rserver'` as helpProvider
- [x] Implement hyperlinks that point to a different section of the same document
- [x] Fix non-standard hyperlinks (there are some links in help files that are not relative to the current file, e.g. `base::colSums`->`complete.cases`)
- [x] Focus/show the help panel if its content changes while it's not focussed/visible. 
- [x] Improve display of non-html pages (package DESCRIPTION files and some files linked on `doc/html/index.html`)
- [x] Improve style of help pages:
    - [x] Adapt to rest of the editor (font family and size adapts, syntax highlighting color theme does not)
    - [x] Add back/forward buttons? (Available in the editor title menu, do not remember scroll position though)
- [x] Configure the helpview properly (automatically and/or by the user)
    - [x] Use font size, font family etc. from editor settings (happens automatically in some places)
    - [x] Choose position of webview
    - [x] Choose which helpProvider to use
- [x] Fix "Index" links that return "Server error: invalid response from R" (example case?)
- [ ] Adapt style file to theme, let user choose style file?
- [ ] Include other files/images (e.g. the R logo on the main doc page)
- [ ] Improve usability of `r.showHelp`: implement fuzzy matching, autocomplete, simulation of `?` from command palette?
- [ ] Implement `??`
- [ ] Handle links to PDF files (and other file types?)
- [ ] General testing and bug fixing
- [ ] ...

I just added the code to get things to work, feel free to rearrange/reformat/rename to better match the rest of the project.
If you have suggestions, questions, comments, or find bugs, please let me know. All feedback or code contributions are welcome :)


ManuelHentschel:
Regarding the implementation the custom help function in R I am not quite sure what the best solution is.

In the normal case there are a number of functions involved when calling e.g. `? print`: The first call is directly to 
``` utils::`?` ```, which then calls `help()` and returns an object of class `help_files_with_topic`.
In an interactive session, this object is then printed with `print.help_files_with_topic`, which actually does the work of starting a help server (`tools::startDynamicHelp`) and opening the corresponding URL (`browseURL`).

I think the cleanest approach would be to overwrite `print.help_files_with_topic`, since this would preserve the behaviour of `help` in cases where the user actually wants to store the return value to a variable and not print it, while still giving us full control over the actual help server (e.g. to make it independent of the current R session).

One problem with this approach might be that the vsc tools are loaded before the utils package and I don't know if loading the package manually (calling `library(utils)`?) to reorder the search path is a good idea. Do you have any advice/opinion on this?

**Comment by tdeenes** [comment]
Maybe loading the 'utils' namespace and overwriting `print.help_files_with_topic()` could work? It does not modify the search path...

In the terminal: `export R_DEFAULT_PACKAGES='base'`
In the R console:
```r
local({
  ns <- loadNamespace("utils")
  search()
  # [1] ".GlobalEnv"   "Autoloads"    "package:base"

  unlockBinding("print.help_files_with_topic", ns)
  assign(
    "print.help_files_with_topic",
    function(x, ...) print.default("Ooops"),
    envir = ns
  )
  lockBinding("print.help_files_with_topic", ns)
})

library(utils)
search()
# [1] ".GlobalEnv"    "package:utils" "Autoloads"     "package:base" 

?mean
# [1] "Ooops"
```

ManuelHentschel:
Thanks! That approach seems to work well.

I just realized that a very similar function is already used in [init.R](https://github.com/Ikuyadeu/vscode-R/blob/ce6316d1e327bd94405c5952738b959cefd0b3f3/R/init.R#L54:L70), I guess we could use that function then.

ManuelHentschel:
I tried finding a solution that does not involve modifying the package namespace of `util`, but that seems to be quite hard. The following code manages to attach the tools just below the global environment:
```r
.First.sys <- function(){
    base::.First.sys()
    attach(list(print.help_files_with_topic = function(...) print('Ooops')))
    rm('.First.sys', envir=globalenv())
}
```
And checking the print method for help files seems to look good as well:
```
> getS3method('print', 'help_files_with_topic')
function (...)
print("Ooops")
<environment: 0x000000001c575778>
```

But calling `? mean` or `print(help('mean'))` still opens the internal help server...

renkun-ken:
It looks like this help viewer starts a standalone R session to provide the help files?

If one uses `radian` instead of `R`, then the R session won't start properly. `r.rterm.*` could only be used as the executable path of the interactive terminal. Any service must use R path instead like how vscode-r-lsp starts R to run languageserver.

ManuelHentschel:
Yep, R is either called to continuously run a help server in the background, or "on demand" to retrieve the libPaths and parse .Rd files.

Would it make sense to merge the settings in vscode-r, vscode-r-lsp, and vscode-r-debugger for this? Otherwise the user would have to specify their r path in three different config entries if it's not in a standard place...

We could e.g. introduce a setting `r.rPath` which can be used by all three extensions.

tdeenes:
> I tried finding a solution that does not involve modifying the package namespace of `utils`, but that seems to be quite hard. The following code manages to attach the tools just below the global environment:
> 
> ```r
> .First.sys <- function(){
>     base::.First.sys()
>     attach(list(print.help_files_with_topic = function(...) print('Ooops')))
>     rm('.First.sys', envir=globalenv())
> }
> ```
> 
> And checking the print method for help files seems to look good as well:
> 
> ```
> > getS3method('print', 'help_files_with_topic')
> function (...)
> print("Ooops")
> <environment: 0x000000001c575778>
> ```
> 
> But calling `? mean` or `print(help('mean'))` still opens the internal help server...

This is because `?` hits `package:stats` on the search path, so any further lookup (e.g, that of 'print.help_files_with_topic') starts there. So the direction becomes `package:stats` -> `namespace:stats` -> `imports:stats` -> base -> global env -> your attached environment. 

See [here](http://blog.obeautifulcode.com/R/How-R-Searches-And-Finds-Stuff/) for an old but helpful explanation.

ManuelHentschel:
That would definitely make sense, if `print` was called from within `?` or `help`. The problem still occurrs though when manually assigning an object to the help-file class and calling print:
```
> getS3method('print', 'help_files_with_topic')
function (...)
print("Ooops")
<environment: 0x000000001c573b68>
> x <- 1
> class(x) <- 'help_files_with_topic'
> print(x)
Error in if (type == "html") tools::startDynamicHelp(NA) else NULL :
  argument is of length zero
```

ManuelHentschel:
I think I've finally found a way to make this work:
```r
.First.sys <- function(){

    base::.First.sys()

    attach(list(
        print.help_files_with_topic = function(...) print('Ooops')
    ))

    .S3method('print', 'help_files_with_topic', print.help_files_with_topic)
}
```

The last call overwrites `utils`' special position when calling print on help objects.
If this does not interfere with the rest of `init.R`, we could use this to attach all the tools below `.GlobalEnv`, or else just attach this overwrite in its own search entry...

tdeenes:
> The problem still occurrs though when manually assigning an object to the help-file class and calling print:

Yeah, this is due to some machinery with S3 method lookup. You are calling the generic first, not your particular custom method implementation.

tdeenes:
> ```r
> .S3method('print', 'help_files_with_topic', print.help_files_with_topic)
> ```

This is a nice trick, but will make vscode-R depend on fresh versions of R. See the official blog post [here](https://developer.r-project.org/Blog/public/2019/08/19/s3-method-lookup/) and the commit [here](https://github.com/r-devel/r-svn/commit/e191d542fc217e81890a6fbdec5d60ac3a7f8893).

ManuelHentschel:
> This is a nice trick, but will make vscode-R depend on fresh versions of R. See the official blog post [here](https://developer.r-project.org/Blog/public/2019/08/19/s3-method-lookup/) and the commit [here](https://github.com/r-devel/r-svn/commit/e191d542fc217e81890a6fbdec5d60ac3a7f8893).

Ok, thanks for catching that. `.S3method` seems to be nothing more than a wrapper for `registerS3method`, so we could just copy the function body to achieve the same thing. I've edited the code above and that seems to work on R 4.0.3 and R 3.5.3.

ManuelHentschel:
@Ikuyadeu @renkun-ken **@???**, when you get a chance, could you please take a look at the code and see if everything works and integrates ok with the rest of the extension?

I think the main functionality of the help panel should work ok now. The remaining entries on the checklist are mostly cosmetic in nature, non-critical problems, or just ideas for improvements, and I'm not sure yet when I'll get around to fixing them.

Regarding the points discussed above: I added an extra option to specify the R path in case `r.rterm.xxx` points to a radian installation and used the trick proposed [here](https://github.com/Ikuyadeu/vscode-R/pull/433#issuecomment-721428581) to overwrite the default `print.help_files_with_topic`.

renkun-ken:
Thanks for the work! I'll take a closer look and feedback soon.

renkun-ken:
@ManuelHentschel Could you write some brief instructions so that I know what could be tested at the moment and what are the expected results? I played with this PR and the 'R: Show Help' only works with default input parameters (package ? `doc`, function name ? `index.html`) and the page looks normal under dark theme and the hyperlinks seem to only work on the top level. I'm not sure if they are already implemented?

ManuelHentschel:
> @ManuelHentschel Could you write some brief instructions so that I know what could be tested at the moment and what are the expected results? I played with this PR and the 'R: Show Help' only works with default input parameters (package ? `doc`, function name ? `index.html`) and the page looks normal under dark theme and the hyperlinks seem to only work on the top level. I'm not sure if they are already implemented?

Sure. The primary way to use the help panel is intended to be:
1. Open an R terminal (`F1` -> `R: Create R terminal`)
2. Enter for example:
```
> ? mean
```
or
```
> help("print")
```
3. Expected result: 
image

---

The commands `R: Show help` is meant as a secondary way to access the R help

Working example inputs should be:
* package: `base`, function name: `print`
* package: `utils`, function name: `help`

helpPanel2

---

The command `R: show R docs` is not as practical, since it links to very large documents that are easier to work with in a proper browser. It can be used to view the same page that is shown when using `help.start()` in normal R:

helpPanel3

---

The expected behavior for the hyperlinks is:
* Links to different help pages: open the help page in the same webview as the current page
* Links to external websites: show a warning/message from vscode and open in an external browser if confirmed by the user
* Links to different parts of the same document (e.g. the ones at the top of "An Introduction to R"): these don't work and should do nothing when clicked

Navigating back/forward between help pages is currently only possible by using the extra buttons 3 & 4 on a mouse (the ones typically used by default to navigate back/forward in normal browsers). It should be possible to also implement these commands with different key combinations (used while in the webview), with buttons added to the help page html, or with vscode internal commands (that could also be assigned key bindings).

helpPanel4

---

R/init.R
```diff
@@ -508,16 +508,59 @@ if (interactive() &&
         )
       }
 
+      print.help_files_with_topic <- function(h, ...) {
+        if (length(h) >= 1 && is.character(h)) {
+          file <- h[1]
+          path <- dirname(file)
+          dirpath <- dirname(path)
+          pkgname <- basename(dirpath)
+          requestPath <- paste0(
```
renkun-ken, line 517: This `requestPath` does not work under macOS.

When I run `?mean` in R terminal under macOS, nothing happens, and the following error appears in the debug console:

image

R/init.R
```diff
@@ -508,16 +508,59 @@ if (interactive() &&
...
+      rm(".First.sys", envir = parent.env(environment()))
+    }
+
     .vsc.attach <- .vsc$attach
     .vsc.view <- .vsc$show_dataview
     .vsc.browser <- .vsc$show_browser
     .vsc.viewer <- .vsc$show_viewer
     .vsc.page_viewer <- .vsc$show_page_viewer
+    .vsc.help <- .vsc$help
+    assign("help", .vsc.help, envir = globalenv())
```
renkun-ken, line 561: Is it necessary to put a `help` function in the global environment to make it work? I'd suggest not introducing anything into the global environment.

ManuelHentschel: That's just something I forgot to remove after testing, thanks!

ManuelHentschel:
Looks like I underestimated the difference that different platforms make here. With `"r.helpPanel.helpProvider": "Rserver"` the help panel now works fine on my WSL setup.

R/init.R
ManuelHentschel: Did the recent pushes resolve this?

renkun-ken: Yes, it works quite nicely now!

R/init.R
```diff
@@ -508,16 +508,57 @@ if (interactive() &&
...
+          pkgname <- basename(dirpath)
+          requestPath <- paste0(
+            "/library/",
+            pkgname,
+            "/html/",
+            basename(file),
+            ".html"
+          )
+          request(command = "help", requestPath = requestPath)
+        }
```
renkun-ken, line 525: When the help document is empty (e.g. `?nothing`), nothing happens.

The original behavior is

```
> ?nothing                                    
No documentation for 'nothing' in specified packages and libraries:
you could try '??nothing'
```

Maybe we could do something similar?

html/theme.css
```diff
@@ -0,0 +1,89 @@
...
+.hljs-constructor,
+.hljs-module-access {
+    color: #000000
+} */
+
+/* .hljs {
+} */
+
+pre {
+    font-family: Consolas, 'Courier New', monospace;
```
renkun-ken, line 70: If the font falls back to `Courier New` (e.g. on macOS or Linux), the code does not look very clear:

`Courier New`:

image

`monospace`:

image

Monaco:

image

It looks like we could use the CSS variables to refer to vscode font settings:

```css
code {
    font-family: var(--vscode-editor-font-family);
    font-size: 14px;
}
```

renkun-ken: I tried the following:

```css
body {
    font-size: var(--vscode-editor-font-size);
}

pre {
    font-family: var(--vscode-editor-font-family);
}

code {
    font-family: var(--vscode-editor-font-family);
}
```

and it looks quite compatible with the editor and when font settings are changed,  the R help webview will automatically adjust too.

renkun-ken:
The webpages look nice and the hyperlinks work nicely too.

I could say it is already quite use-able now.

tdeenes:
In my instance the R Help is blank. What do I wrong?

tdeenes:
VSCode info:
Version: 1.51.0
Commit: fcac248b077b55bae4ba5bab613fd6e9156c2f0c
Date: 2020-11-05T18:16:10.374Z
Electron: 9.3.3
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Linux x64 5.3.0-42-generic

ManuelHentschel:
> In my instance the R Help is blank. What do I wrong?

Did you change any of the settings `r.helpPanel.XXX`?
In case you are using radian as `r.rterm.XXX`, please make sure to enter the path to a "vanilla" R executable in `r.helpPanel.rpath`.

html/theme.css
ManuelHentschel: This look like the best approach to match the style of the help view to the rest of the editor.
AFAIK there are some more variables available to adapt e.g. to the color theme being dark/bright/high-contrast, but no way to access the actual theme colors used for normal syntax highlighting.

tdeenes:
I set it to /usr/bin/R

image

tdeenes:
The panel opens, but it remains blank. No error is printed to the console or to the output.

ManuelHentschel:
How are you triggering the panel to open?
If you haven't already, please try `r.showHelp` with e.g. package: `base`, function: `print`, and also from within an R terminal using `?`

tdeenes:
I tried both, with the same result.

tdeenes:
What is the best way to debug the init.R file?

tdeenes:
[Info  - 15:03:16.907] Starting TS Server 
[Info  - 15:03:16.908] Using tsserver from: /usr/share/code/resources/app/extensions/node_modules/typescript/lib/tsserver.js
[Info  - 15:03:16.910] <syntax> Forking...
[Info  - 15:03:16.917] <syntax> Starting...
[Info  - 15:03:16.917] <semantic> Forking...
[Info  - 15:03:16.922] <semantic> Starting...

html/theme.css
renkun-ken: I tested the dark/light/high-contrast and it already works quite nicely with this added. I guess we could consider this as resolved.

ManuelHentschel:
> 
> 
> [Info - 15:03:16.907] Starting TS Server
> [Info - 15:03:16.908] Using tsserver from: /usr/share/code/resources/app/extensions/node_modules/typescript/lib/tsserver.js
> [Info - 15:03:16.910] Forking...
> [Info - 15:03:16.917] Starting...
> [Info - 15:03:16.917] Forking...
> [Info - 15:03:16.922] Starting...

Is that output from when you try to use the help panel? There sould be something similar to this:
```
pkg: base - fnc: print
requesting help for: library/base/html/print.html
```
(and maybe some error messages)

tdeenes:
Yes, I checked your recent commits, and was wondering why I can not see those logs. I did the following:

1. Remove the current R extension.
2. Git pull + vsce package
3. Install the R extension from the VSIX file.
4. Ctrl+Shift+P, select R: Show help, type: base, print

The result:
image

ManuelHentschel:
To see the output to the debug console, you got to debug the extension as follows:
1. Remove the current R extension
2. Git pull
3. Open the root directory of the extension in vscode
4. Press F5. This should launch a new vscode instance labeled `[Extension Development Host]`.
5. The "original" vscode instance should now indicate a running debug session (e.g. via the orange status bar). In this instance press Ctrl+Shift+Y to open the debug console.

renkun-ken:
User could use Help -> Toggle Developer Tools to see the `console.log` output.

tdeenes:
Great, thanks. This helped me identify the bug :) I will fix it soon.

ManuelHentschel:
Just to b clear, is this a problem in the code of the R help panel which you are planning to fix, or  an issue with your personal setup?

tdeenes:
The problem is with the R help panel. Actually I am not sure how the fix should happen, so I rather put a comment in the problematic part.

src/rHelpProviderBuiltin.ts
```diff
@@ -0,0 +1,103 @@
...
+    public constructor(options: RHelpClientOptions){
+        this.rPath = options.rPath || 'R';
+        this.port = this.launchRHelpServer(); // is a promise for now!
+    }
+
+    public async launchRHelpServer(){
+        // starts the background help server and waits forever to keep the R process running
+        const cmd = (
+            `${this.rPath} --silent --no-echo --vanilla -e ` +
+            `"cat(tools::startDynamicHelp(),'\\n'); while(TRUE) Sys.sleep(1)" `
```
tdeenes, line 31: "--no-echo" is a replacement of "--slave" in R >4.0.0. Note that "--slave" can still be used.

I see three options:

1. Simply use "--slave" instead of "--no-echo" -> this will cause problems when "--slave" gets deprecated.  
2. Extract somehow the version of R: E.g., run `${this.rPath} --version` first, extract the R version from that string, and then start the R help session with the appropriate switch.   
3. Drop the "--no-echo"/"--slave" switch, and extract the port number using a regular expression from the stdout string.

tdeenes:
Btw, the new help looks awesome :)

src/rHelpProviderBuiltin.ts
ManuelHentschel: Thanks for finding this, I went with option 1 for now (we can always implement more sophisticated methods, if/when they deprecate `--slave` at some point)

tdeenes:
Found an issue: package Description files are displayed without line breaks.

ManuelHentschel:
> 
> 
> Found an issue: package Description files are displayed without line breaks.

Thanks, I added this issue to the to-do list in the top comment.

I guess these files don't get rendered properly since they are plain text, but passed to the webview as if they were html. This can probably be fixed by manually inserting line breaks (and maybe other formatting) for non-html files.

ManuelHentschel:
I'll probably not get around to fixing/adding all entries on the to-do list in the next couple of weeks. On my personal computer I am already using the current state of the help panel and am happy with it so far. If you think it's stable and helpful enough (or close to it) we could consider merging the current state.

Anyone interested in solving some of the issues is of course highly welcome :)

renkun-ken:
Do we need an option to opt-in or opt-out overwrite `print.help_files_with_topic`?

ManuelHentschel:
I'd probably do this similar to the other features provided by the session watcher/init.R. Are these enabled individually or all together?

Regardless, we could still consider making the help panel opt-in until it's been tested by more people

renkun-ken:
[Here](https://github.com/Ikuyadeu/vscode-R/wiki/R-Session-watcher#available-functions-and-options) is a list of available options of session watcher so that we could disable features. (Plus `options(vsc.rstudioapi = TRUE)`, forgot to update the wiki).

I guess we could add a `vsc.help_panel = FALSE | "Two" | "Active" | "Besides"` to support opt-in and customizable viewColumn).

tdeenes:
I am wondering what would be the best way to consider the .Rprofile file in the workspace where R is launched. For example if one uses **renv**, which modifies `.libPaths`, the Help view can not find the packages which are only present in the local renv library.

R/init.R
```diff
@@ -531,27 +532,29 @@ if (interactive() &&
       environment()
     })
 
-    .First.sys <- function() {
-      # first load utils in order to overwrite its print method for help files
-      base::.First.sys()
+    if(!isFALSE(getOption("vsc.helpPanel", "Two"))){
```
renkun-ken: `isFALSE()` is added in R 3.5.0. Maybe we cannot rely on this function.

tdeenes: Or we can conditionally import [backports](https://github.com/r-lib/backports). Certainly not a must have for `isFALSE`, but helps avoid such kind of issues.

ManuelHentschel:
> 
> 
> I am wondering what would be the best way to consider the .Rprofile file in the workspace where R is launched. For example if one uses **renv**, which modifies `.libPaths`, the Help view can not find the packages which are only present in the local renv library.

I intentionally added the `--vanilla` flag to keep the .Rprofile from doing "weird" things, but I guess this would be a sensible usecase. We could add a configuration entry (in vscode) to change the flags used when launching R to solve this, but then we probably need to adjust the parsing of info printed from within R, in case .Rprofile generates output to stdout.

R/init.R
```diff
@@ -532,7 +532,7 @@ if (interactive() &&
       environment()
     })
 
-    if (!isFALSE(getOption("vsc.helpPanel", "Two"))) {
+    if (getOption("vsc.helpPanel", "Two") != FALSE) {
```
renkun-ken, line 535: We better use `!identical(getOption("vsc.helpPanel", "Two"), FALSE)` here to avoid being stopped by invalid values. Also consistent with other existing ones.

renkun-ken:
In both vscode-R and langaugeserver, we try to respect user profile as much as possible or there will be significant inconsistencies in some important use cases (e.g. renv).

vscode-R starting a session from terminal will source user profile before loading session watcher.
languageserver will start from workspace folder and also respect user profile to know the correct library to use when providing language services. I suggest that the help panel should respect the user profile too since it is highly related to the user library.

renkun-ken:
On macOS:

```r
?mean
```

The help pane look normal. Then click the link to `weighted.mean`, the page navigates to the following:

```
Link weighted.mean in package base could not be located
However, you might be looking for one of
/library/stats/html/weighted.mean.html
```

Click the link above, I got

```
Only help files, NEWS, DESCRIPTION and files under doc/ and demo/ in a package can be viewed
```

ManuelHentschel:
> It looks like the links that will go to other package's page won't work at the moment.

More precisely links that start with a `/` and links that point to specific parts of a documents (`#...`) do not work due to the way links are replaced [here](https://github.com/ManuelHentschel/vscode-R/blob/e9b46967a43c4fe53363617497e221c92e224215/html/script.ts#L39:L64). This part could be replaced with a proper link resolving library (which might be a bit more difficult to integrate) or by properly resolving/replacing the links ourselves.

This is already on the to-do list, but I don't know when I'll get around to fixing this, since I'm not very familiar with this topic and it would make sense to fix these two together:
> - [ ] Fix non-standard hyperlinks (there are some links in help files that are not relative to the current file, e.g. `base::colSums`->`complete.cases`)
> - [ ] Implement hyperlinks that point to a different section of the same document

ManuelHentschel:
The code now uses javascript's integrated `new URL()` to parse hyperlinks, so the different kinds of links should work properly now.

renkun-ken:
Thanks for the fix of cross-package hyperlinks. I'm wondering if it is possible that when the link points to a function in another package, if there's only one link, then it should jump directly to that link?

ManuelHentschel:
> Thanks for the fix of cross-package hyperlinks. I'm wondering if it is possible that when the link points to a function in another package, if there's only one link, then it should jump directly to that link?

When viewing the R help in a normal browser, the help server sends a redirect response (HTTP 302), but when querying the site from within the extension, the help server returns that "forwarding page" with an Ok response (HTTP 200). The cleanest way to fix this would be to somehow indicate that we also want to receive the redirect response when querying the help site, not sure how this would be done.

ManuelHentschel:
Btw, when using `"r.helpPanel.helpProvider": "custom"`, cross-package links seem to work fine.
However, the custom helpProvider seems to have some problems on linux (and probably also macOS?).

tdeenes:
@ManuelHentschel  FYI: after your new commits, cross-pkg hyperlinks, DESCRIPTION line breaks, forward/back buttons all work nicely on Linux.

renkun-ken:
I tried `??help` and it prints an `hsearch` object and the original R html help server is started and a WebView is shown. I guess we have not done anything about `hsearch` yet.

Also, I navigate though different links around and sometimes clicking the "Index" at the bottom of the help page will result in a text page saying "Server error: invalid response from R".

Another thing is that if I view `User guides, package vignettes and other documentation.` and click `utils::Sweave` which links to a PDF file, the help pane directly tries to show its binary content in text. I guess we could call `vscode.env.openExternal()` in this case as if it was an external link.

ManuelHentschel:
> I tried `??help` and it prints an `hsearch` object and the original R html help server is started and a WebView is shown. I guess we have not done anything about `hsearch` yet.

Yup, the `hsearch` object is not handled by us yet. The method described [here](https://github.com/wch/r-source/blob/6d36d375b598c62574b606d5ba192f2550e64081/src/library/utils/R/help.search.R#L699:L725) might work for us.

ManuelHentschel:
> Also, I navigate though different links around and sometimes clicking the "Index" at the bottom of the help page will result in a text page saying "Server error: invalid response from R".

I haven't encountered this yet, so if you find a reproducible example, please share :)

ManuelHentschel:
> Another thing is that if I view `User guides, package vignettes and other documentation.` and click `utils::Sweave` which links to a PDF file, the help pane directly tries to show its binary content in text. I guess we could call `vscode.env.openExternal()` in this case as if it was an external link.

This might be a bit more complicated, since we receive the contents of the pdf file directly from the help server. Hence we do not know the location of these pdf files on disk for sure. We can take a guess using `R.home()` and `.libPaths` but these values are usually not queried by the default helpProvider (`"r.helpPanel.helpProvider": "Rserver"`, implemented in `rHelpProviderBuiltin.ts`).

Since most of the documentation is provided as html, I wouldn't consider this a priority right now

renkun-ken:
> > Also, I navigate though different links around and sometimes clicking the "Index" at the bottom of the help page will result in a text page saying "Server error: invalid response from R".
> 
> I haven't encountered this yet, so if you find a reproducible example, please share :)

I could reproduce this on both Ubuntu and macOS.

1. `?mean`
2. Click link `weighted.mean`
3. Click Index at the bottom of page.
4. `Server error: invalid response from R`

ManuelHentschel:
> I could reproduce this on both Ubuntu and macOS.
> 
>     1. `?mean`
> 
>     2. Click link `weighted.mean`
> 
>     3. Click Index at the bottom of page.
> 
>     4. `Server error: invalid response from R`

Thanks that example works on windows as well. Does the last push fix the issue for you?

renkun-ken:
It works nicely now. Thanks for the quick fix!

ManuelHentschel:
If there aren't any more (known) bugs to fix, I would consider the remaining items on the to-do list "optional" improvements and propose merging the PR.

html/theme.css
```diff
@@ -0,0 +1,88 @@
+
+
+
+/* I tried to mimic VS Code's Dark+ theme with these colors: */
+
+.hljs-literal,
+.hljs-keyword {
```
renkun-ken, line 7: These hard-coded colors might not appear clearly in light themes. For example, with Solarized Light:

image

Maybe we need the following to make the color work under all types of themes?

```css
body.vscode-light .hljs-literal,
body.vscode-light .hljs-keyword {
    color: ???
}

body.vscode-dark .hljs-literal,
body.vscode-dark .hljs-keyword {
    color: ???
}

body.vscode-high-contrast .hljs-literal,
body.vscode-high-contrast .hljs-keyword {
    color: ???
}
```

ManuelHentschel: The best way to solve this would be to access the colors of the current theme, either directly as RGB values or through CSS variables (as is possible for [theme colors](https://code.visualstudio.com/api/references/theme-color)). AFAIK, this is currently not possible, but comes up frequently ([e.g. here](https://github.com/microsoft/vscode/issues/32813)).

A somewhat hacky workaround would be to manually adapt the CSS files from the default themes and select the correct one using the approach you described (though we would still be limited to dark/light/hc). Taking this one step further, we could also try to automatically parse/convert the CSS files and maybe even let users pass their own theme files (for people who use themes that are vastly different from the default ones).

Either way, a clean solution would either require changes to the colors exposed by vscode, or take some time to understand and convert the involved CSS, which is why I'd consider this a "long-term todo".

ManuelHentschel: I think the current syntax highlighting works ok with the three default themes. To make sure the help view is still usable by people with very different themes, I'd propse adding an option to disable syntax highlighting altogether

renkun-ken: I think you are right. The current syntax highlighting already works quite well for most cases.

renkun-ken (approved):
LGTM. Thanks for the great work!

I believe this PR is ok for merge. For the remaining items, we could always improve later.

@Ikuyadeu Any more comments?

Ikuyadeu:
LGTM, it is great.
I will publish a new version this weekend.

@ManuelHentschel If you have an interest in this extension, I want to invite you as a collaborator.
Because you already submit the 2 pull requests that have large contributions.
And, normally new functions are unstable and required high knowledge to resolving issues.

ManuelHentschel:
> 
> 
> LGTM, it is great.
> I will publish a new version this weekend.
> 
> @ManuelHentschel If you have an interest in this extension, I want to invite you as a collaborator.
> Because you already submit the 2 pull requests that have large contributions.
> And, normally new functions are unstable and required high knowledge to resolving issues.

Thanks! I'm definitely interested in this extension and would be happy to help where I can :)

Issue #439, opened 2020-11-06
User friendly error message when vsc.rstudioapi is not set


matthiasgomolka:
**Describe the bug**
When I want to use an RStudio Addin from VS Code I receive the following error:

> Command 'R: Launch RStudio Addin' resulted in an error (ENOENT: no such file or directory, open '/tmp/RtmpKarM0O/vscode-R/addins.json')

I set `options(vsc.rstudioapi = TRUE)` in my radian profile. This works (checked using `getOption("vsc.rstudioapi")`).

**To Reproduce**
1. Open VS Code
2. Press 'F1' to open the command palette
3. Choose 'R: Launch RStudio Addin'

**Do you want to fix by self? (I hope your help!)**

No (don't know how)

**Expected behavior**
I should be able to choose from installed RStudio Addins.

**Screenshots**
![image](https://user-images.githubusercontent.com/6517388/98413255-95b91b00-2079-11eb-9b48-2eac55867e40.png)

**Environment (please complete the following information):**
 - OS: Linux
 - VSCode Version: 1.51.0
 - R Version: 4.0.3
 - vscode-R version: 1.5.2

renkun-ken:
Before the R session sources session watcher `init.R` script, `options(vsc.rstudioapi = TRUE)` has to be set to enable addins emulation layer which provides the list of addins available in the R session.

This issue is more like we could show a more user-friendly error message to tell user that `options(vsc.rstudioapi = TRUE)` has to be set in user profile to enable this feature.

matthiasgomolka:
Thanks. But which user profile? I think I already tried all (or most) of the possibilities and the radian profile was the one where the option was set right away when the session started.

renkun-ken:
The user profile (`~/.Rprofile`) and project profile `{profileFolder}/.Rprofile` both should work. The radian profile is used when you start radian which does not work with original R terminal. If you intend to make it only work with radian, then I guess using radian profile is fine.

matthiasgomolka:
It's somehow related to radian. When I switch to the standard R console, it works as expected. Any ideas what might have gone wrong with radian?

matthiasgomolka:
Thanks for your patience! Figured it out:

I needed to set the following in VS code:
![grafik](https://user-images.githubusercontent.com/6517388/98439837-e15cda80-20f4-11eb-9e32-6cd00bf2c836.png)

Or as JSON:
```json
"r.alwaysUseActiveTerminal": true
```

[...]

PR #441, opened 2020-11-08
Friendly error message when trying to launch addin picker and vsc.rstudioapi = FALSE

MilesMcBain:

In the case that the user has not set the option `vsc.rstudioapi = TRUE` the addin picker gives an unhelpful error message. This PR adds an error message that directs the user to set the option to TRUE.

A root cause of #439 was the confusing error.

In your `.Rprofile` set `vsc.rstudioapi = FALSE`. Start a new R terminal in VSCode and run the `R: Launch RStudio Addin` command. 

An error should appear:

```
Command 'R: Launch RStudio Addin' resulted in an error (Could not find list of installed addins. options(vsc.rstudioapi = TRUE) must be set in your .Rprofile to use RStudio Addins)
```

### DIFF
--- a/src/rstudioapi.ts
+++ b/src/rstudioapi.ts
@@ -272,7 +272,16 @@ let addinQuickPicks: AddinItem[] = undefined;
 export async function getAddinPickerItems() {
 
   if (typeof addinQuickPicks === 'undefined') {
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

Issue #385, opened 2020-07-22
Error identifying terminal


bersbersbers:
**Describe the bug**
- After hitting F5, my code runs fine in the terminal
- Then executing another line by Ctrl-Enter issues the "Error identifying terminal"
- Focusing terminal once and executing the same line again from the same editor as before makes it work

**Do you want to fix by self? (I hope your help!)**

No, but I can help debug.

**Environment (please complete the following information):**
 - OS: VS Code on Windows 10, using **VS Code Remote SHH** into Linux
 - VSCode Version: 1.47.2
 - R Version: 4.0.2
 - vscode-R version: 1.4.3

andycraig:
Hi @bersbersbers, thank you for reporting this.

When I hit F5 (command 'Debug: Start Debugging'), I get a popup 'Select Environment', with options 'Node.js' and 'More...'. Do you  have some additional setup associated with F5?

bersbersbers:
@andycraig yes, sorry I forgot about this:

```
    {
        "key": "f5",
        "command": "r.runSource",
        "when": "editorTextFocus && editorLangId == 'r'"
    },
```

andycraig:
@bersbersbers Thank you! I have a suspicion that this might be related to using a remote session. I will look into it.

If you experience this error again, could you post here with the names of all the terminals that were open at the time? (E.g., '1: R Interactive', '2: bash' etc.) That will help with investigating the problem. Thanks in advance!

bersbersbers:
Will do! It happens less regularly now, but it did happen once yesterday.

andycraig:
I had a try at duplicating this error using a Docker container and the ‘Remote - Containers’ extension, but I couldn’t get it to happen. (Not quite the same as SSH, but I had thought that just using a remote might be the problem.)

The error ‘Error identifying terminal! This shouldn\'t happen, so please file an issue at https://github.com/Ikuyadeu/vscode-R/issues’ should only happen if the active terminal is undefined, and there is not exactly one terminal. I think it must take a very particular set of circumstances for VS Code to get into that state.

Other information that would be useful for investigating this:

- Whether anything else happens between F5 and Ctrl-Enter (anything at all, even just clicking somewhere)
- Whether this seems to happen at a particular time (e.g., just after starting VS Code, just after creating a new terminal, just after closing a terminal)

Thank you!

bersbersbers:
[...]

PR #447, opened 2020-11-14
Add terminal information to chooseTerminal error

andycraig:

This is related to #385. I can't reproduce the error, so this PR adds more information to the error message, which users can pass on to us.
  
**Screenshot**

I can't reproduce the error, so I made this screenshot by copy-pasting the message code to the top of `chooseTerminal()`.

EDIT: Updated the image. 

terminals3

I don't have instructions for how to reproduce the error, so I think review will have to be code review only, without testing.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -95,8 +95,12 @@ export async function chooseTerminal() {
                     return window.terminals[0];
                 }
             } else {
+                let msg = `There are ${window.terminals.length} terminals: `;
+                for (let i = 0; i < window.terminals.length; i++){
+                    msg += `Terminal ${i}: ${window.terminals[i].name} `;
+                }
                 // tslint:disable-next-line: max-line-length
-                window.showInformationMessage('Error identifying terminal! This shouldn\'t happen, so please file an issue at https://github.com/Ikuyadeu/vscode-R/issues');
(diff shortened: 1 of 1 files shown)


andycraig:
Ready for review.

src/rTerminal.ts
```diff
@@ -72,6 +72,11 @@ export async function chooseTerminal() {
         return window.activeTerminal;
     }
 
+    let msg = '[chooseTerminal] ';
+    msg += `A. There are ${window.terminals.length} terminals: `;
```
Ikuyadeu, line 76: What is this `A.` and 96 lines `B.`, and `C`?
If it means defects depth, how is using the indent?

Ikuyadeu:
Good! It can be merged.
I put only one comment, but it is not so important and you can ignore it.

src/rTerminal.ts
andycraig: Sorry, I should have added more info to the PR. My theory is that the error may be occurring because the number of terminals changes during that section of code. The `A`, `B` and `C` are so we can check the numbers and names of the terminals are three different points. We can see if they are the same each time or if they are changing.

Ikuyadeu: Thank you for the detail. It is clear.
In the future,  we may need to define our extension error code such as `chooseTerminalA` when we use a similar process to identify the point of the defect,

andycraig:
@Ikuyadeu Thanks for the review!

PR #449, opened 2020-11-18
Add GitHub Action for release

### LINKED ISSUE #382: Wish: GitHub release to follow development
Currently the changes are logged in `Changelog.md`. However, there is no way to stay informed about updates to this file.
It is also cumbersome to look up the vscode extension every once in a while for updates. Also for both, no timestamps exist.

It would be great if the "GitHub release" feature could be used (again). Users can subscribe to releases and they contain timestamps.

Thanks for all your work! (Trying) to follow development as close as possible 😄

ManuelHentschel:
Fix #382 (or at least make it easier to fix)

This action file is being used in https://github.com/ManuelHentschel/VSCode-R-Debugger to automatically prepare a release when a tag matching `"v*"` is pushed and so far it seems to be working nicely.

The text of the release is a list of commits since the last release and the .vsix file is added to the release as a binary.

I am using `draft: true` to be able to check the release and add the binaries from the accompanying R package, but that could also be set to `false` to publish the release immediately.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: release
+
+# runs when a tag v* is pushed
+# creates a release draft with the binaries
+
+on:
+  push:
+    tags: ["v*"]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

PR #456, opened 2020-11-22
Fix checking workspaceFolders in rHelpProviderOptions

renkun-ken:

Fix checking `workspaceFolders` in the initialization of the help panel.

1. No open workspace folders.
2. Open or create an R file.
3. vscode-R should be activated properly.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -56,7 +56,7 @@ export async function activate(context: ExtensionContext) {
     }
     const rHelpProviderOptions = {
         rPath: rPath,
-        cwd: (workspace.workspaceFolders.length > 0 ? workspace.workspaceFolders[0].uri.fsPath : undefined)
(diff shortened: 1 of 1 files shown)


renkun-ken:
@Ikuyadeu We might need to quick release for this as the extension activation for no open workspace folder is broken.

ManuelHentschel:
Would it make sense to also check the length of the array?
``` ts
         cwd: ((workspace.workspaceFolders !== undefined && workspace.workspaceFolders.length > 0) ? workspace.workspaceFolders[0].uri.fsPath : undefined)
```

The documentation seems to imply that an array of length 0 won't occurr (`List of workspace folders or `undefined` when no folder is open.`), but better be safe than sorry

renkun-ken:
@ManuelHentschel Agreed.

Note that both vscode-R and languageserver support multi-root workspace. We might revisit the code where `workspaceFolders[0]` is used and see how we could properly handle multi-root workspace in these cases.

ManuelHentschel (approved):

Issue #455, opened 2020-11-22
As of vscode-R 1.6.0 I cannot see help files


nathaneastwood:
I have the latest versions of vscode-R (v1.6.0) and {languageserver} (v0.3.7) on the latest VSCode (v1.51.1) on MacOS 19.6.0. If I type into the console `?rnorm`, as an example, I just see a black tab with nothing displayed there.

These are my current settings:

```json
  "r.rterm.mac": "/usr/local/bin/radian",
  "r.rterm.option": [""],
  "r.bracketedPaste": true,
  "r.sessionWatcher": true,
  "r.lsp.debug": true,
  "r.lsp.diagnostics": true,
  "r.alwaysUseActiveTerminal": true,
```

<img width="585" alt="Screenshot 2020-11-22 at 09 38 45" src="https://user-images.githubusercontent.com/9799530/99899056-91f6dc80-2ca6-11eb-8340-5efb192e7958.png">

renkun-ken:
You might use `options(vsc.helpPanel = FALSE)` to disable the help panel at the moment.

@ManuelHentschel any idea on this?

renkun-ken:
Looks like setting `r.helpPanel.rpath` does not have any effect.

If I set

```json
"r.rterm.mac": "/usr/local/bin/radian",
"r.helpPanel.rpath": "/usr/local/bin/R"
```

it won't work either. and

```json
"r.rterm.mac": "",
"r.helpPanel.rpath": "/something/not/exists"
```

works.

Looks like `r.helpPanel.rpath` is ignored/not used at all. It will always default to `r.rterm.mac`.

nathaneastwood:
Interestingly, if I `devtools::load_all()` and then look at the help file, it will load the development help file in my internet browser.

ManuelHentschel:
`load_all` attaches an environment `devtools_shims` which contains a modified version of `help` that is not affected by the overwrites from `init.R`:
```
> devtools::load_all()
Loading RTestPackage
> search()
 [1] ".GlobalEnv"           "devtools_shims"       "package:RTestPackage"
...
```

renkun-ken:
@nathaneastwood Would you like to install the latest build at https://github.com/Ikuyadeu/vscode-R/actions/runs/377371651 and see if it works for you?

nathaneastwood:
I installed the extension from the vsix and can confirm it works! Thanks very much to all for your quick response to this issue.

Issue #454, opened 2020-11-22
Help View: Blank Window


pat-s:
Hi guys,

very enthusiastic about the new help view.
However, currently I only get a blank window when opening any help page.
I have set `r.helpPanel.rpath` to vanilla R but this does not change anything.
I did not find anything in the dev tools pane.

Also I think `r.helpPanel.rpath` needs to be documented top-level or set to a reasonable default? 
Many peope have set the R interpreter to `radian` and will not know that they need to set the vanilla R path again explicitly here (I only saw this browsing the PR).

Would it make sense to set it to the default `$R_HOME/bin/R` for each OS?

renkun-ken:
No need to set `r.helpPanel.rpath` if you type `?something` to show help viewer. It works on top of session watcher that works with the current session you are running by default. Only the `R: Show help` command will spawn a standalone R process to provide help. Which way do you use and you get a blank page?

If you are using `R: Show help`, then there might be something wrong there. Please try using `?something` in an attached R session.

If you are using `?something`, then something is wrong resolving the help file paths. You might revert to the old behavior via `options(vsc.helpPanel = FALSE)`.

@ManuelHentschel any idea on this?

renkun-ken:
Are you setting `r.rterm.*` to radian?

@ManuelHentschel I think `r.helpPanel.rpath` should not default to `r.rterm.*` as many users are using radian as we recommend, the help panel won't work then. It should default to `getRpath()` instead if not specified.

renkun-ken:
#455 might be the same problem.

ManuelHentschel:
Looks like there was a typo when accessing the config entry `r.helpPanel.rpath`

ManuelHentschel:
> @ManuelHentschel I think `r.helpPanel.rpath` should not default to `r.rterm.*` as many users are using radian as we recommend, the help panel won't work then. It should default to `getRpath()` instead if not specified.

It does default to `getRpath()` [already](https://github.com/Ikuyadeu/vscode-R/blob/master/src/extension.ts#L53), but maybe we should change that function so that the help panel never uses the `r.rterm.*` setting

renkun-ken:
@pat-s Would you like to install the latest build at https://github.com/Ikuyadeu/vscode-R/actions/runs/377371651 and see if it works for you?

pat-s:
Works!

PR #457, opened 2020-11-22
Fix typo in help panel path config

ManuelHentschel:
#454 #455
Fix typo in accessing config entry `r.helpPanel.rpath` which caused this setting to be ignored.

### DIFF
--- a/package.json
+++ b/package.json
@@ -508,7 +508,7 @@
         "r.helpPanel.rpath": {
           "type": "string",
           "default": "",
-          "markdownDescription": "Path to an R executable. Defaults to `r.rterm.XXX`. Must be \"vanilla\" R, not radian etc.! "
+          "markdownDescription": "Path to an R executable. Must be \"vanilla\" R, not radian etc.! Will be read from registry or path if not set."
         },
         "r.helpPanel.helpProvider": {
           "type": "string",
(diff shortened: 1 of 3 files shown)


renkun-ken:
Thanks for the fix! It looks like `r.helpPanel.rpath` does not default to `r.rterm.*` but `getRpath()`, which is inconsistent with the description

>Defaults to `r.rterm.XXX`.

Could you confirm that and update the description if it is the case?

ManuelHentschel:
> 
> 
> Thanks for the fix! It looks like `r.helpPanel.rpath` does not default to `r.rterm.*` but `getRpath()`, which is inconsistent with the description
> 
> > Defaults to `r.rterm.XXX`.
> 
> Could you confirm that and update the description if it is the case?

I used that description, since `getRpath()` will read the corresponding `r.rterm.xxx` setting first and only be different from accessing the setting directly if it needs to fall back to a different  method

ManuelHentschel:
But I'd suggest skipping the `r.rterm.xxx` fallback and instead use the fallback methods used in `getRpath()` directly

renkun-ken:
Yes, that's exactly what I meant. Forgot that getRpath reads config first.

renkun-ken (approved):
LGTM

Tested on macOS and it works nicely now.

PR #460, opened 2020-11-23
Rmd fenced block syntax highlighting for julia, python, etc

kar9222:
Add Rmd fenced block syntax highlighting for common languages in Rmd
- Julia
- Stan
- Python
- SQL
- CSS
- SCSS
- JS

Although there are a total of 42 supported languages, I think we can keep VSCode-R lean by just supporting additional languages above, which should cover almost all use cases. New language can be added upon user request.

```r
❯ names(knitr::knit_engines$get())

 [1] "awk"       "bash"      "coffee"    "gawk"      "groovy"
 [6] "haskell"   "lein"      "mysql"     "node"      "octave"
[11] "perl"      "psql"      "Rscript"   "ruby"      "sas"
[16] "scala"     "sed"       "sh"        "stata"     "zsh"
[21] "highlight" "Rcpp"      "tikz"      "dot"       "c"
[26] "cc"        "fortran"   "fortran95" "asy"       "cat"
[31] "asis"      "stan"      "block"     "block2"    "js"
[36] "css"       "sql"       "go"        "python"    "julia"
[41] "sass"      "scss"
```

# Before PR

# After PR

_(Kindly ignore the minor different colors of my Julia's starting code block because I customize it in my theme extension. They all should have the same colors on your machine e.g. white color as shown here)_

### DIFF
--- a/syntax/Markdown Redcarpet.json
+++ b/syntax/Markdown Redcarpet.json
@@ -13,6 +13,13 @@
                 {"include" : "#fenced_block_cpp"},
                 {"include" : "#fenced_block_yaml"},
                 {"include" : "#fenced_block"},
+                {"include" : "#fenced_block_julia"},
+                {"include" : "#fenced_block_stan"},
+                {"include" : "#fenced_block_python"},
+                {"include" : "#fenced_block_sql"},
+                {"include" : "#fenced_block_css"},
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM.

Tested on macOS and it works nicely. Thanks!

Issue #464, opened 2020-11-23
Add r.rpath.xxx setting


ManuelHentschel:
Currently the extensions [vscode-r-lsp](https://github.com/REditorSupport/vscode-r-lsp/blob/master/package.json#L44:L48), [vscode-r-debugger](https://github.com/ManuelHentschel/VSCode-R-Debugger/blob/master/package.json#L333:L347) and the help panel from [this extension](https://github.com/Ikuyadeu/vscode-R/blob/master/package.json#L519:L523) each have a configuration entry where the user needs to specify the path of an R binary.

These are all "backend" usecases where alternatives like radian offer no advantages (as far as I'm aware) over the standard R executable. To make vscode+R easier to configure, I'd suggest introducing a setting `r.rpath.xxx`, similar to `r.rterm.xxx` which is supposed to be a "vanilla" R binary that can be used by the mentioned functions/packages.

Advantages include:
- Users don't need to make the same configuration in three different places
- For developers it's easier to implement new features that need an R process as backend (e.g. an `R CMD INSTALL .` task)

Disadvantages include:
- Users cannot configure different R executables for the different extensions/functions.
- Changing the settings might invalidate existing settings

As a solution to these two issues, I'd suggest introducing a new unified `r.rpath.xxx` setting, while still honoring the old settings (i.e. to get the R path in the lsp check `r.lsp.path` -> `r.rterm.xxx` -> read from path etc.). Even if the package/function specific settings are not specified in the package.json anymore, it is still possible to read their value from existing `setting.json` files.

ManuelHentschel:
The three PRs linked above contain a first implementation of the unified `r.rpath.xxx` setting.

As mentioned before, there is a tradeoff here, but I think for the majority of users the comfort of setting the rpath only once (if at all) outweighs the possible disadvantage of not being able to configure individual paths for each extension.

I am not sure to what extent this configuration sharing is intended by the authors of vscode, but it seems to work nicely. (In the latest version the hyperlinks in deprecation messages sometimes don't work correctly. In the current insiders version, shared configurations are sometimes displayed twice, but I think that's not intended).

renkun-ken:
>As mentioned before, there is a tradeoff here, but I think for the majority of users the comfort of setting the rpath only once (if at all) outweighs the possible disadvantage of not being able to configure individual paths for each extension.

Agree. There's already too much confusion in these settings and configuration for new users.

renkun-ken:
[...]

PR #467, opened 2020-11-24
Modify config

ManuelHentschel:
Implement changes suggested in https://github.com/Ikuyadeu/vscode-R/issues/464


package.json
```diff
@@ -519,7 +534,9 @@
         "r.helpPanel.rpath": {
           "type": "string",
           "default": "",
-          "markdownDescription": "Path to an R executable. Must be \"vanilla\" R, not radian etc.! Will be read from registry or path if not set."
+          "markdownDescription": "Path to an R executable. Must be \"vanilla\" R, not radian etc.! Will be read from registry or path if not set.",
+          "markdownDeprecationMessage": "Will be deprecated. Use `#r.rpath.windows#`, `#r.rpath.mac#`, or `#r.rpath.linux#` instead.",
```
renkun-ken, line 538: Looks like the deprecation message is not shown in either settings visual editor or json editor:

This entry no longer appears in the completion of settings:

image

But if I use this settings, then I don't see any deprecation warnings in either the description or the diagnostics:

image

ManuelHentschel: That's weird... Have you tried packaging+installing the modified code and checking in a new vscode instance? Sometimes that changes to package.json don't take effect when debugging the extension normally

renkun-ken: I build all three extensions and installed the three vsix files.

ManuelHentschel: Do the deprecation warnings work in any of the other two extensions?

Just for info, this is how the warnings are supposed to look:

deprecation1

deprecation2

ManuelHentschel: Any updates on this? I checked the settings UI on a linux VM and both types of deprecation messages work there

renkun-ken: In the settings viewer, the deprecation message looks nice:

image

But in the `settings.json` editor, I don't see the deprecation message:

image

ManuelHentschel: Thanks for the update. Since I cannot reproduce the issue, I probably won't be able to fix this. I'd suggest we just add e.g. `DEPRECATED! ...` to the front of the normal description to make sure people with this error notice the change (hoping that the proper deprecation message works for most users)

(And as we still honour the old setting in this PR, it's not a breaking change even if users don't notice)

renkun-ken: Sounds good to me.

PR #468, opened 2020-11-25
Add pre-release

ManuelHentschel:
This PR adds a GitHub action that releases a development build marked as pre-release on every push to master.

At [vscode-r-debugger](https://github.com/ManuelHentschel/VSCode-R-Debugger/releases) and [vscDebugger](https://github.com/ManuelHentschel/vscDebugger/releases) the action seems to be working nicely.

This would make it possible to download and test the latest master version, without having to navigate to the correct build action and downloading the artifacts.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,48 @@
+name: pre-release
+
+# creates a pre-release with the .vsix 
+
+on:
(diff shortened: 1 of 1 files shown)


Ikuyadeu:
LGTM, This PR may reduce the risk when I can not release it.

PR #470, opened 2020-11-26
Improve Help Panel

ManuelHentschel:
Improve the help panel and fix https://github.com/Ikuyadeu/vscode-R/issues/471:
- [x] Remove use of `isFALSE`
- [x] Fix some bugs that caused the entire extension to crash, if the help panel path was invalid
- [x] Parse package index to suggest package names when calling `R: Show Help`
- [x] Parse index files of packages to suggest documented functions from a selected package
- [x] Implement `??` functionality
    - [x] Accessible with command `R: Search Help` or by selecting search icon after command `R: Show Help`
    - [x] Make accessible from R session
- [ ] Improve performance:
    - [x] Cache installed packages/functions in package per session
    - [ ] Keep cache between sessions?
    - [x] Add option to clear cached index files
- [ ] Implement `?` functionality (from vscode command palette)
- [x] Identify and highlight package pages like e.g. `base/html/base-package.html`
- [ ] ...

### DIFF
--- a/src/rHelpProviderCustom.ts
+++ b/src/rHelpProviderCustom.ts
@@ -64,6 +64,8 @@ export class RHelp implements rHelpPanel.HelpProvider {
 	readonly homePath: string;
 	// temp directory used to extract .Rd file to
 	readonly tempDir: string;
+	// list of installed packages
+	readonly installedPackges?: string[];
 
 	constructor(options: RHelpOptions = {}) {
 		this.rPath = options.rPath || 'R';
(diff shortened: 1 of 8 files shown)


renkun-ken (approved):
LGTM

Thanks for the nice improvements!

The package and topics picking in "Show help" commands works quite nicely!

ManuelHentschel:
Thanks for the feedback. Regarding the to remaining items on the checklist:

> Implement `?` functionality (from vscode command palette)

I started working on this, but it seems to be a bit more complicated, so I'll do this in a separate PR


> Keep cache between sessions?

This would make sense for people with many installed packages on slower computers, but might complicate things for people working with different package versions/environments. For now I'd skip this feature and implement it later if necessary.

src/rHelpProviderCustom.ts
```diff
@@ -98,11 +100,17 @@ export class RHelp implements rHelpPanel.HelpProvider {
 			// use .libPaths() in R
 			const cmd = `${this.rPath} --silent --no-save --no-restore  --no-echo -e "cat('${lim}', paste(.libPaths(), collapse='\\n'), '${lim}', sep='')"`;
```
renkun-ken, line 101: `--no-echo` only works with >R 4.0. Is there a way to make it more robust?

PR #361, opened 2020-06-16
WIP: Updating whole of source code

Ikuyadeu:
**What problem did you solve?**

* [ ] Fix for the new eslint rules (Now in progress)
Current style `plugin:@typescript-eslint/eslint-recommended` was suitable for JavaScript.
I add the other extends
* `eslint:recommended`: Same with 'plugin:@typescript-eslint/eslint-recommended' 
* `plugin:@typescript-eslint/recommended`: Recommended rules for the TypeScript

* [ ] Refactor `session.ts`
Current `session.ts` has too many global values,
```ts
export let globalenv: any;
let resDir: string;
let watcherDir: string;
let requestFile: string;
let requestLockFile: string;
let requestTimeStamp: number;
let sessionDir: string;
let pid: string;
let globalenvFile: string;
let globalenvLockFile: string;
let globalenvTimeStamp: number;
let plotView: string;
let plotFile: string;
let plotLockFile: string;
let plotTimeStamp: number;
let plotDir: string;
let requestWatcher: FSWatcher;
let globalEnvWatcher: FSWatcher;
let plotWatcher: FSWatcher;
```

to help code understanding, it should be included in the function declaration.

* [ ] Clean up `extension.ts`

Current `extension.ts` has many functions like `getBracketCompletionItems`.
They should be each other files, and `extension.ts` will just call them.
Personally, `break loop1;` is not good for understanding code flows. it can be more easily.

Unfortunately, I do not have enough knowledge for the R, so I will not touch `init.R` in this PR.

* [x] Apply [super-linter](https://github.com/github/super-linter)

PR #474, opened 2020-11-29
Fix whole of style (Extends #361)

Ikuyadeu:

#361 was too large to solve all issues.
This PR only apply the easy fix, and I gave up that fixing all rule violation once, especially for the `any` value warnings.
I want to fix remained violations in another issue, after merge this.

I confirmed that fixing current new functions source code readability is very hard work.
This PR can be the first step to keeping this extension maintainability

Next TODO:
* Refactor any values it can be object
  * extension.ts
  * rHelpPanel.ts
  * RHelpProviderBuiltin.ts
  * rstudioapi.ts
  * session.ts
  * test/suite/index.ts

* Simplify extension.ts
  * Remove goto label it can be simple for statement
  * Move Rmd functions to another file


Ikuyadeu:
Could you review this?
I think this change has no impact on the code behavior, but I want to someones check.

src/test/suite/extension.test.ts
```diff
@@ -20,7 +20,7 @@ suite('Extension Tests', () => {
             y
         }
         `.split('\n');
-        function f(i) {return (doc[i]); }
+        function f(i: number) { return (doc[i]); }
```
renkun-ken, line 23: Here on my side, eslint complains about the use of the deprecated `assert.equal`:

image

Should we do anything about it?

renkun-ken:
I go through the changes and all look good. Tested on macOS and everything works well.

src/rHelpPanel.ts
```diff
@@ -1,5 +1,10 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
 
-import * as vscode from 'vscode';
+import { commands, window, QuickPickItem, Uri, Webview, WebviewPanel, WebviewOptions, WebviewPanelOnDidChangeViewStateEvent, ViewColumn } from 'vscode';
```
ManuelHentschel, line 7: Is there a performance benefit from importing each item explicitly or is this just a stylistic choice?

src/test/suite/extension.test.ts
Ikuyadeu: Thank you for your point.
Maybe it is not eslint, it is a typescript function alert.
But It's good timing, I'll fix it to `assert.strictEqual`

src/rHelpPanel.ts
Ikuyadeu: It is not alerted by ESLint, but it is useful to refactoring based on the vscode API.
In this case, we can confirm  `rHelpPanel.ts` manage the Webview API related function.
If someone implements a related function in another file, we should combine that in this file or implement super-class

Ikuyadeu:
I merge this, thank you @renkun-ken and @ManuelHentschel for your comments.

PR #475, opened 2020-11-29
Fix bug that would leave background R processes running

ManuelHentschel:
Fix a command that would not properly kill background R processes started by the help panel.


renkun-ken (approved):
LGTM

PR #476, opened 2020-11-29
Implement R workspace viewer

### LINKED ISSUE #416: Workspace viewer similar to vscode-julia
It is time-consuming having to individually examine objects through the `View()` command. `View(globalenv())` does not update automatically so cannot be used for this purpose.

I understand that this has been around about brought up in a few issues, such as #15, #23, and #47. However, I think that with the session viewer having access to the current R environment, it would be neat to do something similar to what vscode-julia has semi-recently implemented [here](https://github.com/julia-vscode/julia-vscode/pull/1003).

Considering that the session viewer currently examines the R environment, I think exposing what it shows in the sidebar would be helpful. This exists in vscode-julia, and in R Studio (i.e. the environment viewer). We currently sort of have this through the `View(globalenv())` call (which is great!), but the returned dataframe is not interactive, doesn't update without reloading the object, and takes up a pane in the workspace.

At the moment, we are able to access objects, graphs, etc. with the `View()` command when we have the session viewer enabled. In vscode-julia, when you click on, for example, a dataframe, it opens up the dataframe in the active panel. I think we could have a similar thing, where objects can be opened by clicking them in the "workspace viewer", so that we can easily and quickly move between examining objects, functions, etc. It could be it calls a `View()` command in the active R terminal when you click on the object.

Overall, I think what is missing right now is:
- Workspace viewer that updates without reload
    - It could be that when a variable is defined (or re-defined), the workspace viewer is updated
- Ability to preview environment objects by clicking them in the workspace viewer
- Workspace viewer contained within sidebar area to not take up space

If the dataframe from the `View(globalenv()` was updated following an object being defined, it could act as a simple environment viewer.

The `View()` command can be bound to a keybind in order to offset typing, and can be used with objects selected in order to speed this up even more. However, it does not offset the fact that it must be run at all, as a live workspace would remove this step.

**Additional context**

Here is the julia workspace preview, indicated by the red arrow:
Capture

Here is a small mockup I made for example:
mockup

ElianHugh:

Closes #416, implementing a workspace/environment viewer.

**Screenshot**

image

What should happen:
1. The workspace viewer is available on the activity bar
2. When the global environment is updated, the viewer is automatically refreshed
3. Clicking the "View" button on a variable should call View() in the R terminal.
4. Clicking the Clear all button should call rm(list = ls()) in the R terminal.

**Changelog**

c3f3c17
- Fixed redundant view title
- Added remove command
- Simplified workspace.ts

2a97b29 
- Implemented save/load workspace buttons
- Fixed typings

7b4edd4 
- Implemented confirmation box for workspace clearing
- Changed default folder for save/load dialog

8c86577
- Implemented element sorting for the workspace viewer
  - Sort priority is: data variables (data.frame, list, environment, etc.)  -> alphabetical label sort -> alphabetical class sort
- Added settings option for removing hidden items
- Changed default folder for saving/opening .RData files (subject to change)
- Renamed workspace.ts -> workspaceViewer.ts
- Fixed some typings

44f0b79
- Changed getWorkspacePaths() to workingDir

9e2eef3 
- Changed save string from 'workspace' to 'workspace.RData' due to save filters not applying on non-Windows OS

fd06ed0
- Changed save string from concatenation to template literal
   - `defaultUri: Uri.file(workingDir + path.sep + 'workspace.RData')` to `defaultUri: Uri.file(`${workingDir}${path.sep}workspace.RData`)`

dd912c9
- Changed the sort function from ternary form to if-else form

cf1ef1f
- Changed the priority list for sorting from variable classes (e.g. data.frame) to variable types (e.g. list)

d9eb94c
- Added globalenv conditional for workspace commands, so that they can't be run if session isn't attached

2bae9c0
- Changed if statements from equality (==) to identity (===)

6cc95e8
- Use dim() as description for eligible variables
- Fix double quote usage

4036a19 
- Fix lint errors
- Fix matrix and array descriptions

f822da8 
- Dim() only applies to list-type data structures
- Fixed typo pertaining to dim() descriptions with only 1 variable


package.json
```diff
@@ -63,6 +63,31 @@
...
+          "title": "R Workspace",
+          "icon": "./images/Rlogo.svg"
+        }
+      ]
+    },
+    "views": {
+      "workspaceViewer": [
+        {
+          "id": "workspaceViewer",
+          "name": "R Workspace",
```
renkun-ken, line 79: This name seems a bit redundant as I move the R icon  from the side bar to one of the views, it becomes

image

renkun-ken: Maybe the `activitybar` should take a title `R`, and `workspaceViewer` has the name `Workspace`?

renkun-ken:
One thing to notice is that when an R session exits, the `globalenv.json` is removed, but we don't have such case at the moment, and then the workspace view does not reset. We'll probably fix this in a separate PR later.

package.json
```diff
@@ -531,6 +571,22 @@
           "when": "resourceScheme =~ /webview/ && r.helpPanel.active",
           "group": "navigation"
         }
+      ],
+      "view/item/context": [
+        {
+          "command": "r.workspaceViewer.view",
+          "group": "inline"
+        }
+      ],
```
renkun-ken, line 580: Is it useful to add a `r.workspaceViewer.remove` which sends `rm(<symbol>)` so that user could remove a symbol from the global environment?

src/extension.ts
```diff
@@ -279,6 +287,9 @@ export async function activate(context: ExtensionContext) {
         commands.registerCommand('r.runCommandWithEditorPath', runCommandWithEditorPath),
         commands.registerCommand('r.runCommand', runCommand),
         commands.registerCommand('r.launchAddinPicker', launchAddinPicker),
+        commands.registerCommand('r.workspaceViewer.refreshEntry', () => rWorkspace.refresh()),
+        commands.registerCommand('r.workspaceViewer.view', (node: WorkspaceItem) => runTextInTerm(`View(${node.label})`)),
+        commands.registerCommand('r.workspaceViewer.clear', () => runTextInTerm(`rm(list = ls())`)),
```
renkun-ken, line 292: Do we ignore those symbols starting with `.`? Or we should send `rm(list = ls(all.names = TRUE))`?

src/workspace.ts
```diff
@@ -0,0 +1,81 @@
...
+	constructor() {}
+
+	getTreeItem(element: WorkspaceItem): TreeItem | Thenable<TreeItem> {
+		return element;
+	}
+
+	getChildren(element?: WorkspaceItem): ProviderResult<WorkspaceItem[]> {
+		return Promise.resolve(
+			this.getWorkspaceItems(this.data));
+	}
```
ManuelHentschel, line 35:
``` ts
	getTreeItem(element: WorkspaceItem): TreeItem {
		return element;
	}

	getChildren(): WorkspaceItem[] {
		return this.getWorkspaceItems(this.data);
	}
```

src/workspace.ts
```diff
@@ -0,0 +1,81 @@
...
+			toItem(
+				key,
+				data[key].class,
+				data[key].type,
+				data[key].str,
+				data[key].length
+			)) : [];
+
+		return items;
+	}
```
ManuelHentschel, line 65: We should probably consider sorting the elements by name and/or by type here

ManuelHentschel (commented):
Thanks for the nice work! :)

src/extension.ts
ElianHugh: I think ideally we would give the users the choice, either through a settings option or a dialog box?

package.json
ElianHugh: Solved in c3f3c17 

image

package.json
ElianHugh: Added remove in c3f3c17

image

src/workspace.ts
```diff
@@ -0,0 +1,81 @@
...
+        length: number;
+        str: string;
+    }
+}
+
+export class WorkspaceDataProvider implements TreeDataProvider<WorkspaceItem> {
+	private _onDidChangeTreeData: EventEmitter<WorkspaceItem | undefined | null | void>
+		= new EventEmitter<WorkspaceItem | undefined | null | void>();
+	readonly onDidChangeTreeData: Event<WorkspaceItem | undefined | null | void>
+		= this._onDidChangeTreeData.event;
```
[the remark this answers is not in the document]
ElianHugh: Thank you! Simplified in c3f3c17

src/workspace.ts
ElianHugh: Simplified in c3f3c17

src/extension.ts
renkun-ken: Removing all variables is quite destructive. I guess we should show a dialog box for the user to confirm.

As for whether to remove all variables (including hidden variables), the RStudio behavior is following

image

Could we do something similar? Not sure if we could put a checkbox on the dialog.

renkun-ken:
In RStudio, the environment pane has `Load` and `Save` buttons so that the user could click to load a workspace from or save one to `.RData` file.

image

It seems we could do similar by sending  `load(file)` and `save.image(file)` to do the same.

ElianHugh:
> 
> 
> In RStudio, the environment pane has `Load` and `Save` buttons so that the user could click to load a workspace from or save one to `.RData` file.
> 
> image
> 
> It seems we could do similar by sending `load(file)` and `save.image(file)` to do the same.

As per your suggestion, 2a97b29 implements save/load workspace buttons.

src/workspace.ts
ElianHugh: I've sorted the elements by name and type like so (2a97b29):

- key: string,
- rClass: string,
- str: string,
- type: string,
- length: number

Is this what you had in mind?

src/workspace.ts
```diff
@@ -65,14 +65,44 @@ export class WorkspaceDataProvider implements TreeDataProvider<WorkspaceItem> {
...
 		super(label, collapsibleState);
-		this.description = typeDetailed;
+		this.description = str;
 		this.tooltip = `${label} (${rClass}, length of ${length})`;
 	}
 }
+
+export function saveWorkspace(): void {
+	window.showSaveDialog({
+		defaultUri: Uri.file('workspace.RData'),
```
renkun-ken, line 82: Looks like it will go to the root folder every time the dialog is open. Should we make a Uri from the workspace folder? For example, get the active document uri, and get workspace folder from that uri. If no workspace folder contains the active document, then we go to the folder of the active document. If there's no active folder, then we go to user home?

ElianHugh: Agreed, I'll work on a fix

src/workspace.ts
```diff
@@ -65,14 +65,44 @@ export class WorkspaceDataProvider implements TreeDataProvider<WorkspaceItem> {
...
+		if (uri) {
+			runTextInTerm(
+				`save.image(\"${(uri.fsPath.split(path.sep).join(path.posix.sep))}\")`
+			);
+		}
+	});
+}
+
+export function loadWorkspace(): void {
+	window.showOpenDialog({
```
renkun-ken, line 97: https://github.com/Ikuyadeu/vscode-R/pull/476/files/c3f3c17ba1c79c67dbf4bcc555b7232b33c308a6..2a97b296da7902887977bd8177eeb2ac10f14c03#r533010048

src/extension.ts
ElianHugh: I've implemented a message box: 

image

There doesn't seem to be an option to use a checkbox as per https://github.com/Microsoft/vscode/issues/56722. I think we could have two confirm buttons, or rely on the settings menu?

src/workspace.ts
```diff
@@ -77,32 +77,57 @@ export class WorkspaceItem extends TreeItem {
...
+			void runTextInTerm(
+				`load("${(uri[0].fsPath.split(path.sep).join(path.posix.sep))}")`
 			);
 		}
 	});
 }
+
+function getWorkspacePath(): string {
+		if (workspace.workspaceFolders) {
+			return workspace.workspaceFolders?.map(folder => folder.uri.path)[0]
```
renkun-ken, line 127: Since we support multi-root workspace, we should find the workspace folder of the active document.

You could use `workspace.getWorkspaceFolder(window.activeTextEditor.document.uri)` to get which workspace folder the active document belongs to.

The logic should be:

1. There's an active document
    1. Find its workspace folder
    2. If no workspace folder is found, then find its file folder via  `path.dirname(window.activeTextEditor.document.uri.fsPath)`.
    3. If it is not a file (e.g. untitled  document), use `os.homedir()`.
2. There's no active document.
    1. If there is a workspace folder, use the first one.
    2. If there is no open workspace folder, use `os.homedir()`.

src/workspace.ts
ManuelHentschel: Sorry that comment was worded stupidly. I meant the items shown in the workspace viewer. Currently, assigning e.g. 
``` r
> a <- 1
> c <- 3
> b <- 2
> l <- list(1,2,3)
> ll <- list(3,4,5)
> h <- 99
> x <- 4
```
will be shown in the workspace view as
```
x num 4
ll List of 3
a num 1
b num 2
c num 3
h num 99
l List of 3
```

I was trying to suggest either sorting by variable name:
```
a num 1
b num 2
c num 3
h num 99
l List of 3
ll List of 3
x num 4
```
and/or variable type:
```
a num 1
b num 2
c num 3
h num 99
x num 4

l List of 3
ll List of 3
```

Sorry again for being so unclear

renkun-ken: In RStudio, the environment pane looks like

image

Not sure if we could do something similar?

ManuelHentschel:
Maybe we should rename the file `workspace.ts` to e.g. `workspaceViewer.ts` to be more precise about what is implemented in the file

src/workspace.ts
ElianHugh: @ManuelHentschel Ah, no, not at all, that makes a lot more sense in retrospect (although sorting the inputs is probably helpful for reading the code anyway). 

Sorting should be easy enough, with something like `return items.sort((a, b) => a.label.localeCompare(b.label) || a.contextValue.localeCompare(b.contextValue));` where contextValue is defined as the type in the WorkspaceItem constructor.

@renkun-ken The tree view is a bit limited in terms of customisation/UI, but I think it should be possible if I create parent elements "Data" and "Values" and then make the make the tree items children of them. I might have to fiddle around with this a bit!

renkun-ken: >I create parent elements "Data" and "Values" and then make the make the tree items children of them. I might have to fiddle around with this a bit!

I think there's no need to do make it complicated. Just find `data.frame`, `list`, and `environment` as data variables and put them on top, and data variables and non-data variables are sorted alphabetically, respectively.

src/workspace.ts
renkun-ken: Wait, the workspace viewer gets data from `globalenv.json` from session watcher. I think it should use the session working directory.

src/workspace.ts
ElianHugh: image

Sorting should now

- Prioritise data at the top
- Then sort by name alphabetically
- Then sort by class alphabetically

renkun-ken: If sort by name, then there won't be duplicates, why "Then sort by class alphabetically"?

src/workspace.ts
ElianHugh: Do you mean the temporary folder housing the globalenv.json? E.g. "C:\Users\elian\AppData\Local\Temp\Rtmp2fxYLc\vscode-R"

src/extension.ts
ElianHugh: Implemented a settings option in 8c86577

image

src/workspace.ts
ElianHugh: Err... good point. Oops. I'll fix that up

src/workspace.ts
renkun-ken: No, I mean the working directory of the session, which is known at `request.wd` (https://github.com/Ikuyadeu/vscode-R/blob/master/src/session.ts#L502) but we don't store it in a variable in `session.ts` yet.

renkun-ken: You could add an exported variable (e.g. `workingDir`) in `session.ts` and set the variable on session attach (e.g. after https://github.com/Ikuyadeu/vscode-R/blob/master/src/session.ts#L512), and use this variable instead of any workspace folder detection.

src/workspaceViewer.ts
```diff
@@ -113,7 +111,7 @@ export function clearWorkspace(): void {
 
 export function saveWorkspace(): void {
 	void window.showSaveDialog({
-		defaultUri: Uri.file(getWorkspacePath() + path.sep + 'workspace'),
+		defaultUri: Uri.file(workingDir + path.sep + 'workspace'),
```
renkun-ken, line 114: Looks like it should be `'workspace.RData'`.

ElianHugh: I think it should save as workspace.RData because of `filters: {'Data': ['RData']}`

renkun-ken: I tried it on Ubuntu but no luck, it saves to workspace rather than workspace.RData.

ElianHugh: Ah, rats, thanks for the heads up! I'll fix that up

renkun-ken: eslint complains that "Operands of '+' operation must either be both strings or both numbers. Consider using a template literal.".

src/extension.ts
renkun-ken: LGTM

src/workspaceViewer.ts
```diff
@@ -0,0 +1,143 @@
...
+				'data.frame',
+				'data.table',
+				'environment',
+				'list',
+				'tibble',
+				'tbl_df',
+				'tbl'
+			]
+
+			return priorityAttr.includes(a.contextValue) > priorityAttr.includes(b.contextValue) ? -1 : priorityAttr.includes(b.contextValue) > priorityAttr.includes(a.contextValue) ? 1 : 0 || a.label.localeCompare(b.label);
```
renkun-ken, line 72: This line is a bit too long. Would you like to rewrite it in `if else` form instead?

src/workspaceViewer.ts
```diff
@@ -0,0 +1,149 @@
...
+				'data.frame',
+				'data.table',
+				'environment',
+				'list',
+				'tibble',
+				'tbl_df',
+				'tbl'
+			]
+
+			if (priorityAttr.includes(a.contextValue) > priorityAttr.includes(b.contextValue)) {
```
renkun-ken, line 72: I think we should use `type` to tell if an item is data.

For example,

```r
a <- 1
m <- lm(mpg ~ cyl, data = mtcars)
```

`m` has type `list` and class `lm`. In RStudio, it is treated as data. But here, its class is not in the `priorityAttr` list, then it is treated as non-data.

Treat all items of the following `type`s will be much simplified:

* `list`
* `environment`

It will cover all these cases and will be consistent with RStudio behavior.

ElianHugh: Oh, yes, you're right - that's a lot better. Fix incoming

ElianHugh:
I've noticed that open/save workspace require the R session to be attached for them to work. Is there a way to force attach or require attach?

src/workspaceViewer.ts
renkun-ken: The fix LGTM

renkun-ken:
>I've noticed that open/save workspace require the R session to be attached for them to work. Is there a way to force attach or require attach?

If `r.alwaysUseActiveTerminal = false` (default), then an R terminal is started if there's no active terminal. Otherwise, it will directly send text to the current terminal (if any) no matter whether there's an open terminal.

Currently, I suggest that you check if `globalenv` is `undefined` before sending commands to the terminal. In the future, we'll fix this by resetting `globalenv` to `undefined` if the active R session is terminated.

src/workspaceViewer.ts
```diff
@@ -95,50 +95,56 @@ export class WorkspaceItem extends TreeItem {
...
-		"Are you sure you want to clear the workspace? This cannot be reversed.",
-		"Confirm",
-		"Cancel"
-	).then(selection => {
-		if (selection == "Confirm") {
-			if (removeHiddenItems) {
-				return runTextInTerm(`rm(list = ls(all.names = TRUE))`)
-			} else {
-				return runTextInTerm(`rm(list = ls())`)
+	if (globalenv != undefined) {
```
renkun-ken, line 98: We better use `!==` here and the other places.

src/workspaceViewer.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export function clearWorkspace(): void {
+	const removeHiddenItems: boolean = config().get('workspaceViewer.removeHiddenItems');
+	if (globalenv !== undefined) {
+		void window.showInformationMessage(
+			"Are you sure you want to clear the workspace? This cannot be reversed.",
+			"Confirm",
+			"Cancel"
+		).then(selection => {
+			if (selection === "Confirm") {
```
renkun-ken, line 104: Please use single quotes for strings instead of double quotes in the typescripts. Not sure why eslint does not complain. Will take a look later.

ElianHugh: Oops, sorry - still getting used to this! Will fix.

Ikuyadeu: @renkun-ken  I removed custom rules in #474.
It's my miss, and I just backed them now.

renkun-ken: @Ikuyadeu Thanks for the notice!

ElianHugh:
Last commit allows for more RStudio-esque data descriptions. Instead of relying just on str(), we use dim() to get the number of observations for eligible variables

image

src/workspaceViewer.ts
```diff
@@ -84,24 +88,33 @@ export class WorkspaceItem extends TreeItem {
...
 		super(label, collapsibleState);
-		this.description = str;
+		this.description = this.getDescription(dim, str);
 		this.tooltip = `${label} (${rClass}, length of ${length})`;
 		this.contextValue = type;
 	}
+
+	private getDescription(dim: number[], str: string): string {
+		if (dim !== undefined) {
+			return `${dim[0]} obs. of ${dim[1]} variables`
```
renkun-ken, line 102: This should apply to `matrix` and `array` (which have `dim`) but somehow it does not as I tested.

renkun-ken: Just checked `init.R` and we only write `dim` when the object is `list` or `environment` which has `names()`. In this case, only `data.frame` will have `dim`.

renkun-ken, line 102: Is it better to show the first class of the object such that 

```
data.table: ? objs. of ? variables
```

Also, we need to make it right for one variable case by using "variable".

renkun-ken:
@ElianHugh Thanks for the updated. Please notice and fix the eslint complains in the diffs.

src/workspaceViewer.ts
ElianHugh: Oops, yes that's my bust, I completely forgot about non-data.frame objects. Thanks for the heads up!

ElianHugh:
> 
> 
> @ElianHugh Thanks for the updated. Please notice and fix the eslint complains in the diffs.

Will do! For some reason my local eslint isn't picking up some of these errors...

src/workspaceViewer.ts
```diff
@@ -92,16 +93,22 @@ export class WorkspaceItem extends TreeItem {
...
 		super(label, collapsibleState);
-		this.description = this.getDescription(dim, str);
+		this.description = this.getDescription(dim, str, rClass);
 		this.tooltip = `${label} (${rClass}, length of ${length})`;
 		this.contextValue = type;
 	}
 
-	private getDescription(dim: number[], str: string): string {
+	private getDescription(dim: number[], str: string, rClass: string): string {
 		if (dim !== undefined) {
```
renkun-ken, line 102: In RStudio, it looks like only data.frame will get a description of `n obs. m variables` and array will use str as its description. It makes sense because matrix and array may not fit the description in terms of nobs and variables but number of rows and columns, or other dimensions.

I suggest we adopt this way. Only `list` type with `dim` uses `class: n obs. m variables` description, otherwise use its `str` instead.

ElianHugh:
I think we should make a decision re: which data types use the `? obs.` format. Currently we have:

- List-type data structures
   - e.g. data.frame, tbl, data.table
- Array-type
   - e.g. simple `array()`, matrices 

This leaves some data types potentially unaccounted for:
- vectors created through `c()`
- vectors created through `x:y`
   - eg. `vec <- 1:10`
- factors


From what I can tell, RStudio seems to only treat list-type data structures as the `? obs.` format. This means that matrices and arrays are shown as, for example `num [1:2, 1] 1 1`.

R/init.R
```diff
@@ -97,6 +97,10 @@ if (interactive() &&
               if (isS4(obj)) {
                 info$slots <- slotNames(obj)
               }
+              if (is.list(obj) ||
+                  is.array(obj)) {
+                info$dim  <- dim(obj)
```
renkun-ken, line 102: I think we could always write `dim` as long as there's `!is.null(dim(obj))`?

Or minimal output, only write `dim` when `is.list(obj) && !is.null(dim(obj))`.

ElianHugh: I think to be in line with RStudio, we would opt for the minimal output you suggested

renkun-ken: Sure

renkun-ken (approved):
LGTM

Thanks for your contribution to this feature!

@Ikuyadeu @andycraig Any more comments?

andycraig:
@renkun-ken I won't have time to look at this for a while, so please proceed without me.

renkun-ken:
@ElianHugh the workspace viewer relies on session watcher to work. Therefore, I guess we don't need to create workspace viewer if session watcher is disabled? Not sure if it is possible, but an alternative way is to tell user one must enable session watcher to use the feature in the default text if it finds that session watcher is disabled?

Ikuyadeu:
@renkun-ken LGTM
After squashing this PR, I think we need to refactor the command category like the python extension.

ElianHugh:
> 
> 
> LGTM
> 
> Thanks for your contribution to this feature!
> 
> @Ikuyadeu @andycraig @ManuelHentschel Any more comments?

Great! Thank you very much for help and suggestions!

> 
> 
> @ElianHugh the workspace viewer relies on session watcher to work. Therefore, I guess we don't need to create workspace viewer if session watcher is disabled? Not sure if it is possible, but an alternative way is to tell user one must enable session watcher to use the feature in the default text if it finds that session watcher is disabled?

Hmm, that's a good point. I think it should be possible to hide the viewer when session watcher is disabled via the package.json file (when property). I'll try it out in a little bit.

PR #477, opened 2020-11-29
Reorganize helppanel, add `?` function

ManuelHentschel:
This PR adds the functionality of R's `?` function to the help panel and does some restructuring to better organize the different ways of accessing the help panel.

To see the new behaviour:
1. Call `R: show Help` from the command palette
2. Observe the restructured menu
3. Select `?`
4. Start typing as you would after `? ` in an R terminal and press enter to confirm help entry


renkun-ken:
Thanks for working on this!

Could you be more specific about how we could test this PR?

ManuelHentschel:
Sure, see top comment

R/getAliases.R
```diff
@@ -0,0 +1,28 @@
+
+# get installed packages, making sure that high priority packages are on top
+ip <- installed.packages()
+
+ord <- order(ip[, "Priority"])
+ip <- ip[ord, ]
+
+ret <- list()
+
+for (row in rownames(ip)) {
```
renkun-ken, line 10: Better use `lapply()` instead of `for` in this case to avoid many re-allocation of `ret`:

```r
ret <- lapply(rownames(ip), function(row) {
    libPath <- ip[row, "LibPath"]
    pkg <- ip[row, "Package"]
    filename <- file.path(libPath, pkg, "help", "aliases.rds")
    info <- list(
        package = pkg,
        libPath = libPath,
        aliasFile = filename,
        aliases = NULL
    )
    if (file.exists(filename)) {
        info[["aliases"]] <- as.list(readRDS(filename))
    }
    info
})
names(ret) <- rownames(ip)
```

R/getAliases.R
```diff
@@ -0,0 +1,30 @@
+
+# get installed packages, making sure that high priority packages are on top
+ip <- installed.packages()
+
+ord <- order(ip[, "Priority"])
+ip <- ip[ord, ]
+
+ret <- list()
```
renkun-ken, line 8: This line could be removed then.

R/getAliases.R
```diff
@@ -0,0 +1,30 @@
...
+        aliases = NULL
+    )
+    if (file.exists(filename)) {
+        info[["aliases"]] <- as.list(readRDS(filename))
+    }
+    info
+})
+names(ret) <- rownames(ip)
+
+json <- jsonlite::toJSON(ret, auto_unbox = TRUE)
```
renkun-ken, line 25: This requires `jsonlite`. If user has not installed this package, then the menu will not show and no error message appears either. Do you think we should handle such case and tell user that `jsonlite` is required?

src/rHelpAliases.ts
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        this.aliases = ret;
+    }
+
+    // call R script `getAliases.R` and parse the output
+    private readAliases(): void {
+		const lim = '---vsc---'; // must match the lim used in R!
+		const re = new RegExp(`^.*?${lim}(.*)${lim}.*$`, 'ms');
+        const cmd = `${this.rPath} --silent --no-save --no-restore --slave -f ${this.rScriptFile}`;
```
renkun-ken, line 109: I'm curious about how robust it is as many Windows users may have spaces in the absolute paths of both `rPath` and `rScriptFile`?

src/rHelpAliases.ts
```diff
@@ -0,0 +1,114 @@
...
+        const txt = cp.execSync(cmd, {encoding: 'utf-8'});
```
renkun-ken, line 110: I don't see a way here to handle the error (e.g. https://github.com/Ikuyadeu/vscode-R/pull/477/files#r532560750)?

src/rHelpAliases.ts
ManuelHentschel: I added quotes for the rScriptFile, the rPath is already quoted [here](https://github.com/ManuelHentschel/vscode-R/blob/bc7714ae2cd002a900b5f0ddb14f096c4a3665ce/src/extension.ts#L59:L61)

R/getAliases.R
ManuelHentschel: I added a `require('jsonlite')` in the beginning to avoid throwing errors.

src/rHelpAliases.ts
ManuelHentschel: Here I added a `try` block and a check for an empty json to gracefully handle failed alias reading. The remaining alias provider functions will simply return `null` when no aliases were found.

The rHelpPanel will then show an error message if the aliases are `null`

renkun-ken (approved):
LGTM

Thanks!

PR #487, opened 2020-12-08
Conditionally show view

ElianHugh:
Only show workspace view when session watcher is enabled

Workspace viewer only functions when session watcher is enabled. Currently, the workspace viewer is displayed even when non-functional because the session watcher is not enabled.

If session watcher is enabled, the workspace viewer should be visible. When session watcher is disabled, workspace viewer should not be shown.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -56,15 +56,6 @@ export async function activate(context: ExtensionContext): Promise<RExtension> {
         rWorkspace
     );
 
-    // creates a custom context value for the workspace view
-    // only shows view when session watcher is enabled
-    const sessionBool = workspace.getConfiguration('r')['sessionWatcher'];
-    if (sessionBool === true) {
-        void commands.executeCommand('setContext', 'r.WorkspaceViewer:show', true);
-    } else {
-        void commands.executeCommand('setContext', 'r.WorkspaceViewer:show', false);
-    }
-
     // used to export an interface to the help panel
     // used e.g. by vscode-r-debugger to show the help panel from within debug sessions
     const rExtension = new RExtension();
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -56,6 +56,15 @@ export async function activate(context: ExtensionContext): Promise<RExtension> {
         rWorkspace
     );
 
+    // creates a custom context value for the workspace view
+    // only shows view when session watcher is enabled
+    const sessionBool = workspace.getConfiguration('r')['sessionWatcher'];
```
renkun-ken, line 61: I think this could be moved down to https://github.com/Ikuyadeu/vscode-R/pull/487/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R314-R315

where the session watcher setting is checked.

src/extension.ts
```diff
@@ -56,6 +56,15 @@ export async function activate(context: ExtensionContext): Promise<RExtension> {
...
+    if (sessionBool === true) {
+        void commands.executeCommand('setContext', 'r.WorkspaceViewer:show', true);
```
Ikuyadeu, line 63: Here can reduce if statement by 
```ts
void commands.executeCommand('setContext', 'r.WorkspaceViewer:show', sessionBool);
```
And I suggest the name `enabledSessionWatcher`instead of `sessionBool`

src/extension.ts
ElianHugh: Fix inc! Is there much of a difference between config() and workspace.getConfig? I hadn't seen that command in my googling

src/extension.ts
ElianHugh: Much more elegant, thank you!

renkun-ken (approved):
LGTM

src/extension.ts
Ikuyadeu: Good, thank you for the early fix

Ikuyadeu (approved):

Issue #463, opened 2020-11-23
Help view: search functionality


tdeenes:
**Is your feature request related to a problem? Please describe.**
Search for words, code snippets, etc. in a manual opened by HelpView

**Describe the solution you'd like**
Add an icon to the top of the panel which opens a search field if clicked. If the user starts typing, highlight the matching elements in the content of the view, and let the user step through the hits.

ManuelHentschel:
This would definitely be nice to have, but quite difficult to implement cleanly, I think.

Since webviews behave similar to separate websites, we cannot use vscode's integrated search features in the help panel, but would need to implement our own search functionality.

If anyone know how to do this cleanly or has alternative ideas, please let me know :)

tdeenes:
My understanding is that a WebView is basically an HTML content within an iframe-like context, so 

1. As you described, we can not use vscode's integrated search features, but
2. We can run any javascript within the HTML.

So we could inject a CDN reference to e.g. [mark.js](https://markjs.io), a JavaScript keyword highlighter library. Or is it more involved?

ManuelHentschel:
That sounds like it should work. Integrating mark.js in the help panel might take some effort, since communication between the webview and the rest of the vscode-extension is limited to `postMessage()` commands and access to local files/other resources is somewhat limited.

Would you be interested in working on this? I probably won't get around to doing this myself, but I'll be happy to help, if you want to give it a go :)

tdeenes:
OK, I can give it a try. However, it will not happen soon; I will let you know if I make some progress with it.

tdeenes:
Just after implementing my own solution based on mark.js, I found a trivial way to add this functionality: there is an option "enableFindWidget" in WebviewPanelOptions. 

@ManuelHentschel , can you please take a look at the PR (#488)?

PR #488, opened 2020-12-10
Find in topic (help panel) #463

tdeenes:
See #463 

Search for words, code snippets, etc. in a manual opened by HelpView

Add an icon to the top of the panel which opens a search field if clicked. If the user starts typing, highlight the matching elements in the content of the view, and let the user step through the hits.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "onCommand:r.runCommandWithEditorPath",
     "onCommand:r.runCommand",
     "onCommand:r.showHelp",
-    "onCommand:editor.action.webvieweditor.showFind",
     "onCommand:r.helpPanel.back",
     "onCommand:r.helpPanel.forward"
   ],
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "onCommand:r.runCommandWithEditorPath",
     "onCommand:r.runCommand",
     "onCommand:r.showHelp",
+    "onCommand:editor.action.webvieweditor.showFind",
     "onCommand:r.helpPanel.back",
     "onCommand:r.helpPanel.forward"
   ],
@@ -279,6 +280,11 @@
       {
         "command": "r.showHelp",
         "title": "R: Show help"
+      },
+      {
+        "command": "editor.action.webvieweditor.showFind",
+        "title": "R Help Panel: Find in Topic",
+        "icon": "$(search)"
       },
       {
         "command": "r.helpPanel.back",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "onCommand:r.runCommandWithEditorPath",
     "onCommand:r.runCommand",
     "onCommand:r.showHelp",
+    "onCommand:editor.action.webvieweditor.showFind",
```
ManuelHentschel, line 61: I think this should be removed, since we don't want the extension to activate on any `showFind` command and if the find command is meant for a help-panel-webview, the extension will already be active anyways

src/rHelpPanel.ts
```diff
@@ -395,11 +396,12 @@ export class HelpPanel implements api.HelpPanel {
...
-				enableScripts: true,
+			const webViewOptions: WebviewOptions & WebviewPanelOptions = {
+        enableScripts: true,
+        enableFindWidget: true
 			};
-			this.panel = window.createWebviewPanel('rhelp', 'R Help', this.viewColumn, webViewOptions);
-
+      this.panel = window.createWebviewPanel('rhelp', 'R Help', this.viewColumn, webViewOptions);
+      
 			// virtual uris used to access local files
```
ManuelHentschel, line 405: Please fix the indentation here :)

ManuelHentschel (commented):
Thanks for the improvement, works really nicely!

renkun-ken:
Tested on macOS and it works quite nicely.

src/rHelpPanel.ts
```diff
@@ -13,6 +13,7 @@ import * as hljs from 'highlight.js';
 import { config } from './util';
 
 import * as api from './api';
+import { downloadAndUnzipVSCode } from 'vscode-test';
```
Ikuyadeu, line 16: Here is warned by eslint and downloadAndUnzipVSCode can be removed.

package.json
tdeenes: oops, fixed (9a5705d85a533df4bb872433de8a2253a5bca70d)

src/rHelpPanel.ts
tdeenes: done (47b869a9ce519aaa893a6365a54ec27a13093ff2)

src/rHelpPanel.ts
tdeenes: thanks, fixed (21182c1f9ab2bda6cfc31c631c34ad466a4647b1)

src/rHelpPanel.ts
tdeenes: Is there a reason why space- and tab-indentation is mixed in the project? Why not stick with one style (preferring spaces)?

ManuelHentschel: I'd assume that is just the result of the different defaults/preferences from different developers. This could probably be fixed by some linter rule

ManuelHentschel (approved):
LGTM

renkun-ken (approved):
LGTM

PR #492, opened 2020-12-10
Modify pre-release action

### LINKED ISSUE #484: Current GitHub release/pre-release policy
Hi guys,

I like that there is an automated GH release approach now. However, I currently get a notification for every commit in master because it updates the "pre-release" release of the repo (I am watching "releases" for this repo).

Would it be possible to only publish automated release for actual releases? Otherwise I would need to unwatch the repo again and loose the newly gained release cycles completely again 😢

ManuelHentschel:
Fix #484 

This action modifies the pre-release-action by updating the last pre-release instead of creating a new one, which avoids notifying subscribers about a new release after each push to master.
Notifications about proper releases are not affected.

I tested this in a private repository and it seems to work nicely. The first pre-release using this action might trigger the notification or need some cleanup, but subsequent pre-releases should be fine.

To test: clone some vscode-extension to a private repo and push to master a couple of times.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -6,6 +6,9 @@ on:
   push:
     branches: ["master"]
 
+env:
+  FILE_OUT: r-latest.vsix
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

PR #494, opened 2020-12-12
Add browser WebView command buttons

renkun-ken:

This PR adds the following buttons to WebViews created by `browser`.

* Refresh: useful If the browser WebView is loaded too early (e.g. https://github.com/microsoft/vscode/issues/112297) or not properly loaded, or  the user wants to reload the page for some reason.
* Open in external browser: useful if the WebView cannot show the page properly (most likely due to some WebView limitations).
* Find: find text in WebView. This is identical to the "R Help Panel: Find in Topic" command. Therefore I rename this command so that it sounds more general.

Also, this PR uses `vscode.env.asExternalUri` introduced at <https://code.visualstudio.com/api/advanced-topics/remote-extensions#forwarding-localhost> instead of [port mapping](https://code.visualstudio.com/api/advanced-topics/remote-extensions#option-2-use-a-port-mapping). Now the port forwarding should work smoothly under remote development when a browser WebView is created without relying on explicit port mapping (currently not working, see https://github.com/microsoft/vscode/issues/102449) and auto port forwarding (currently not working smoothly, see https://github.com/microsoft/vscode/issues/112297).

**(If you have)Screenshot**

image

1. Create a R terminal with session watcher attached.
2. Run `shiny::runExample("01_hello")` and a WebView is created
3. Focus on the WebView, and the two commands should appear on the right side of the tab.
4. Click "Refresh", and the WebView is reloaded.
5. Click "Open in external browser", and the user web browser is launched to show the web page.
6. Click "Find", and the find widget should work.

"Open in external browser" should also work under remote development as it will automatically setup the port forwarding from remote to local. To test, just repeat the above steps under remote development and the behavior should be exactly the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -600,6 +600,11 @@
         },
         {
           "command": "editor.action.webvieweditor.showFind",
+          "when": "resourceScheme =~ /webview/ && r.browser.active",
+          "group": "navigation"
+        },
+        {
+          "command": "editor.action.webvieweditor.showFind",
           "when": "resourceScheme =~ /webview/ && r.helpPanel.active",
           "group": "navigation"
         },
--- a/package.json
+++ b/package.json
@@ -292,7 +292,7 @@
       },
       {
         "command": "editor.action.webvieweditor.showFind",
-        "title": "R Help Panel: Find in Topic",
+        "title": "R WebView: Find in text",
         "icon": "$(search)"
       },
       {
--- a/package.json
+++ b/package.json
@@ -279,12 +279,14 @@
       {
         "command": "r.browser.refresh",
         "title": "Refresh",
-        "icon": "$(refresh)"
+        "icon": "$(refresh)",
+        "category": "R"
       },
       {
         "command": "r.browser.openExternal",
         "title": "Open in external browser",
-        "icon": "$(link-external)"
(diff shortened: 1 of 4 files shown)


ManuelHentschel:
Works nicely, thanks!

I added a small change that hides the help-panel controls after closing the help panel (this was previously done only on changeViewState events).

Would it be possible/feasible to implement the back and forward buttons for general webviews? E.g. after calling `utils:::print.help_files_with_topic(help('mean'))` and clicking some hyperlink it would be nice to be able to go back to the initial help page. There seems to be some mechanism to implement this for [iframes](https://stackoverflow.com/questions/3254985/back-and-forward-buttons-in-an-iframe), though I haven't tried using this in vscode yet.

ManuelHentschel (approved):

renkun-ken:
Thanks for the review!

>Would it be possible/feasible to implement the back and forward buttons for general webviews?

The browser WebView is mainly used to show reactive content (e.g. shiny apps, httpgd graphics) rather than general purpose web browsing. I suspect if it is really useful to be able to go backward/forward on the page since most reactive content is not designed to work with such navigation operations.

PR #497, opened 2020-12-14
WIP: Add Help-Tree-View, improve help panel

ManuelHentschel:
Includes a number of changes to the help view:
* Adds a treeview titled "R Help Pages", shown in the sidepanel below the workspace viewer, that gives more convenient access to the help pages
* Improves caching of help pages. E.g. noticable when opening the same help page multiple times using hyperlinks (not back/forward).
* Separates general help related code from webview specific code -> users can open and use multiple help panels simultaneously

This PR is still work in progress, but most things should already work. If you find bugs or have ideas for improvements, please let me know.


renkun-ken:
Thanks for working on this!

image

I tried running the extension and have the following thoughts at the moment:

1. I think the package nodes in the help viewer as shown above are a bit overloaded as they provide a list of child nodes of package topics but not searchable. And these child nodes are displayed with topics description as label, making it look a bit messy since many function titles in R packages could be a bit long. I suggest that we don't make the package nodes collapsible to show the package topics in the help viewer but clicking the package node shows up the picker for the user to navigate and search package topics directly, which reduces the overlap of the functionality and makes it more useful.
2. At the moment, the help viewer is mostly a duplication of the help topics picker. To make it more useful, I'm wondering if we could turn it into a package manager viewer by providing features like viewing installed packages, showing package help topics (as we have now), installing new packages, removing existing package, etc.

ElianHugh:
Small nitpick but when the help viewer has its own view, the title is a bit redundant.

image

ManuelHentschel:
> Small nitpick but when the help viewer has its own view, the title is a bit redundant.

The help viewer is supposed to show up in the same view as the worksapce viewer (as in the screenshot posted by @renkun-ken ). From my experience, you sometimes have to reinstall the extension to properly reflect changes to package.json (`ctrl+shift+b` -> wait for installation to finish -> `F1` -> `Developer: Reload Window`).

ManuelHentschel:
> I think the package nodes in the help viewer as shown above are a bit overloaded as they provide a list of child nodes of package topics but not searchable. And these child nodes are displayed with topics description as label, making it look a bit messy since many function titles in R packages could be a bit long. I suggest that we don't make the package nodes collapsible to show the package topics in the help viewer but clicking the package node shows up the picker for the user to navigate and search package topics directly, which reduces the overlap of the functionality and makes it more useful.

I agree that the viewer might be a bit of an information overload, as it is right now.
However, I really like the option to view all help pages of a topic and quickly navigate between them. E.g. for the `anytime` package (visible in the screenshot) it's very easy to get an overview of its functionality by checking out the "package home" and its four help topics, rather than havnig to check the help page for each function from the package.

Regarding the names of the nodes in the tree view I agree that they might be too long, but I couldn't think of a better way to label them, since often a single help topic is shared by a large number of functions.

Implementing a package-scoped search function is a nice idea. I'd suggest adding a command button to the package node (the ones that show op on the right hand side when hovering over an entry; currently none of these are implemented in the help viewer) that triggers this.

ManuelHentschel:
> At the moment, the help viewer is mostly a duplication of the help topics picker. To make it more useful, I'm wondering if we could turn it into a package manager viewer by providing features like viewing installed packages, showing package help topics (as we have now), installing new packages, removing existing package, etc.

To some extent this duplication is intentional since I am using/planning to use the same "backend" functions for both. The current topic piker can be more convenient to quickly find a help page using the keyboard, but can also get a little annoying when you're just trying to "look around" in the help topics of a package.

The package manager functionality is a nice idea, I guess we could easily implement e.g. a `Remove Package` command that shows up for each package node. Grouping packages by lib path should also be possible, as well as installing new packages (maybe we can even parse a list of available packages from CRAN and suggest them in a quickpick?).

ManuelHentschel:
In the last commit I added a search button to each package in the help viewer, as suggested by @renkun-ken. To test, open the help pages viewer, select `Help Topics by Package`, hover over any entry and click the magnifying-glass-icon that appears on the right hand side.

To the topic pages I added a similar button (`+`) that opens the page in a new webview instead of the currently open one.

ManuelHentschel:
The last commit adds some more commands to each package entry in the tree view:
* "Remove package": Uninstalls the package using `remove.packages()`. I am not 100% happy with the confirmation prompt. Using `window.showWarningMessage` would also work, but these messages might be hard to notice/easy to overlook
* "Add to favorites"/"Remove from favorites": Pins the package to the top of the package list/Unpins it
* "Clear cached files": Clears the help files cached for that package. Might be useful to refresh the help pages of a single package without losing all other cached files.

ManuelHentschel:
This has gotten a bit messy so I'll close for now and open a new PR after some refactoring etc.

PR #502, opened 2020-12-22
Improve help view

ManuelHentschel:
Follow up to #497.

Adds a treeview titled "Help Pages" shown in the sidepanel below the workspace viewer:
* Open help pages by using one of the commands previously (only) available through the command "R: Show Help".
* View installed packages and help-topics per package in a tree view
* Mark packages as favorites for quicker access
* List topics either by alias (the name of the R function, contains duplicate links) or page title (unique links)
* Remove installed packages
* Install packages from CRAN, picking from the packages listed [here](https://cran.r-project.org/web/packages/available_packages_by_date.html)
* Each level of the treeview can also be opened as quickpicker, allowing faster navigation with the keyboard

Includes several other improvements to the help viewer:
* Improves caching of help pages. E.g. noticable when opening the same help page multiple times using hyperlinks (not back/forward).
* Separates general help related code from webview specific code -> users can open and use multiple help panels simultaneously
* Back/Forward buttons in the editor now remember the current scroll position (previously only worked when using mouse buttons 3 & 4)
* Adds an "Open in external browser" command to help pages, which is particularly useful for viewing PDF files
* Automatically opens code samples in a new editor instead of the webview


.eslintrc.yml
```diff
@@ -13,8 +13,11 @@ plugins:
...
-    no-extra-semi: "warn"
-    curly: "warn"
-    quotes: ["error", "single", { "allowTemplateLiterals": true } ]
-    eqeqeq: "error"
+  semi: "error"
+  no-extra-semi: "warn"
+  curly: "warn"
+  quotes: ["error", "single", { "allowTemplateLiterals": true } ]
+  eqeqeq: "error"
+  no-empty: "off"
```
Ikuyadeu, line 21: Ignoring these rules can be potential bugs.
The empty block should be implemented before merging pulls.
`no-inferrable-types` can be defined by the `type` or `interface`.
no-unused-vars will lots the code readability

src/rHelpPanel.ts
```diff
@@ -392,47 +138,27 @@ export class HelpPanel implements api.HelpPanel {
...
-		return this.panel.webview;
+		if(!helpFile){
+			helpFile = this.currentEntry.helpFile;
+		}
+		const url = helpFile.url;
+		if(!url){
+			return false;
+		}
+		const uri = vscode.Uri.parse(url);
+		const externalUri = await vscode.env.asExternalUri(uri);
```
renkun-ken, line 153: According to https://code.visualstudio.com/api/advanced-topics/remote-extensions#opening-something-in-a-local-browser-or-application and https://code.visualstudio.com/api/advanced-topics/remote-extensions#forwarding-localhost, `vscode.env.asExternalUri()` is not needed in this case. Directly using `vscode.env.openExternal(uri)` should work.

renkun-ken:
>* Install packages from CRAN, picking from the packages listed [here](https://cran.r-project.org/web/packages/available_packages_by_date.html)

In my network, it takes forever to fetch something from <https://cran.r-project.org> instead of the mirror I use. Is it possible to respect user mirror in some way and add a timeout for this?

src/rHelpPanel.ts
renkun-ken: The docs of `asExternalUri` also says:

>Note that uris passed through openExternal are automatically resolved and you should not call asExternalUri on them.

.eslintrc.yml
ManuelHentschel: I agree about the `no-unused-vars`, the other two rules I disabled can be a bit annoying though:
* `no-inferrable-types`: The point of this rule is to reduce unnecessary verbosity. Verbosity could make code harder to read, but so can missing (explicit) type definitions, so I think this should be left to the programmer to decide.
* `no-empty`: The only instance where I encounter this rule are empty `catch` blocks and empty callbacks passed to functions (when these cannot be left out). The rule can be circumvented by adding a comment in the otherwise empty block, but that adds e.g. two extra lines to an empty catch block and still allows people to leave empty functions in the code.

src/rHelpPanel.ts
ManuelHentschel: Thanks 👍

ManuelHentschel:
> In my network, it takes forever to fetch something from https://cran.r-project.org instead of the mirror I use. Is it possible to respect user-specified mirror in some way and add a timeout for this?

`https://SOME_CRAN_MIRROR/web/packages/available_packages_by_date.html` seems to work, e.g. `https://www.freestatistics.org/cran/web/packages/available_packages_by_date.html`.

Does this work for you? If the `SOME_CRAN_MIRROR` part matches the CRAN mirror you specify in the R options, this shouldn't be too difficult to implement.

.eslintrc.yml
Ikuyadeu: Yes, your examples point to the limitation of the current rule. So how is using rule options?
In this case, we can be staying in the TypeScript standard.

* For the `no-inferrable-types`, we can use  `ignoreProperties` option. I think `ignoreParameters` option is also good in the future.
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-inferrable-types.md#ignoreparameters
* For the `no-empty`, then we can use the `allowEmptyCatch` option. https://eslint.org/docs/rules/no-empty#allowemptycatch

renkun-ken:
Looks the the R command does not work:

```
rror: Command failed: /usr/local/bin/R --silent --slave --no-save --no-restore -e cat('---vsc---') -e "cat(getOption('repos')['CRAN'])" -e cat('---vsc---')
/bin/sh: -c: line 0: syntax error near unexpected token `('
/bin/sh: -c: line 0: `/usr/local/bin/R --silent --slave --no-save --no-restore -e cat('---vsc---') -e "cat(getOption('repos')['CRAN'])" -e cat('---vsc---')'

	at checkExecSyncError (child_process.js:630:11)
	at Object.execSync (child_process.js:666:15)
	at Object.execSync (electron/js2c/asar.js:727:23)
	at executeRCommand (/Users/ken/Workspaces/github/vscode-R/dist/extension.js:4795:27)
	at async Object.getCranUrl (/Users/ken/Workspaces/github/vscode-R/dist/extension.js:4760:21)
	at async PackageManager.getPackages (/Users/ken/Workspaces/github/vscode-R/dist/extension.js:74882:49)
	at async /Users/ken/Workspaces/github/vscode-R/dist/extension.js:4748:15
```

.eslintrc.yml
ManuelHentschel: These rules pretty much cover all my annoyances, thanks for pointing them out :)

ManuelHentschel:
> 
> Looks the the R command does not work:
> 

Adding quotes resolved the issue on WSL for me.

renkun-ken:
The R command works now but install package command has the same problem:

```
> Executing task: /usr/local/bin/R --silent -e install.packages('data.table') <

zsh:1: number expected
The terminal process "/bin/zsh '-c', '/usr/local/bin/R --silent -e install.packages('data.table')'" terminated with exit code: 1.

Terminal will be reused by tasks, press any key to close it.

```

renkun-ken:
Note that RStudio install package dialog supports installing multiple packages.

image

I'm wondering if it makes sense to allow multi-select in the picker.

ManuelHentschel:
> The R command works now but install package command has the same problem:

I tried fixing these issues using vscode's builtin `vscode.ShellQuotedString` and it seems to be working on Windows and WSL. Could you please check on macOS/ubuntu as well?

ManuelHentschel:
> I'm wondering if it makes sense to allow multi-select in the picker.

This required just some small changes to the package installation function, so I added an option to install multiple packages via an extra button shown next to the main command.

Ikuyadeu (approved):

renkun-ken:
>so I added an option to install multiple packages via an extra button shown next to the main command

Is it added already? Can't find it yet.

ManuelHentschel:
> 
> 
> > so I added an option to install multiple packages via an extra button shown next to the main command
> 
> Is it added already? Can't find it yet.

I forgot to push, should be there now

renkun-ken:
In the picker of installing multiple packages, does it make sense and is it possible to make the selected packages always show? Or otherwise, I can't see what packages I have selected to install.

Just like this

image

ManuelHentschel:
> 
> 
> In the picker of installing multiple packages, does it make sense and is it possible to make the selected packages always show? Or otherwise, I can't see what packages I have selected to install.

That would be nice to have, but rather difficult to implement. We are currently calling `vscode.window.showQuickPick` with a list of packages, which handles all of the user interaction. The function simply returns one of the items (if `canPickMany: false`), an array of items (if `canPickMany: true`) or `undefined`. The behavior of the quickpick menu can't really be configured by us and is not optimized for this many items (~16k).

If I interpret your screenshot correctly, the settings sync quickpick doesn't have this problem since it only contains 5 items in total.

A workaround for this issue could be to use `vscode.window.createQuickPick`, which is more flexible than `showQuickPick`, but also a lot more complicated to implement (and I still wouldn't know how to solve this issue).

ManuelHentschel:
I've got the impression that modifying the quickpick to always show the selected items would be quite difficult and/or not work properly. Instead, I replaced the confirmation dialogue with another quickpick that includes only the previously selected packages and selects all of them by default. Do you think that's a sensible compromise?

renkun-ken:
Thanks for the quick update. It works quite well already.

renkun-ken:
BTW, it seems that there's a notable latency or delay on the extension activation. Is it caused by populating the help topics? If it is the case, is it possible to delay the action so that the long list is populated on user collapsing the item?

ManuelHentschel:
> 
> BTW, it seems that there's a notable latency or delay on the extension activation. Is it caused by populating the help topics? If it is the case, is it possible to delay the action so that the long list is populated on user collapsing the item?

The tree view is already populated on demand and deactivating the help related stuff (by commenting out the line [```await initializeHelp(context, rExtension)```](https://github.com/ManuelHentschel/vscode-R/blob/c95ddd9361d716b88ac123ee70d2b55158b7320d/src/extension.ts#L58) in the activation function) did not improve the extension activation on my machine.

renkun-ken (approved):
LGTM Thanks!

PR #505, opened 2020-12-24
releaseAction

### LINKED ISSUE #503: GitHub Releases
We currently have a GitHub [action](https://github.com/Ikuyadeu/vscode-R/blob/master/.github/workflows/release.yml) that prepares a release draft, when a tag matching `v*` is pushed.

Modifying the action to actually publish the release instead of drafting can easily be done, but this still requries the corresponding tag to be pushed.

@Ikuyadeu, would you be ok with tagging the releases you publish to the marketplace to make sure they are also released to GitHub? In principle, this can be done by any collaborator, but the safest way would be to do this at the same time as releasing to the marketplace.

Two alternatives would be:
1. We also automate the release to the marketplace, as done e.g. [here](https://github.com/ManuelHentschel/vscode-github-linker/blob/master/.github/workflows/release.yml#L49:L61). This would make it possible to simultaneously release to the marketplace and GitHub by pushing a single tag
2. After each push, we check if the version changed and automatically tag and release the commit containing the version change. This might be a bit cumbersome to implement, since we'd need to compare a number of different commits to tag the correct one, and lead to another workflow running on each push.

I'd prefer alternative 1 or sticking with the current action (with the modification that releases are published, not just drafted), what are your thoughts on this?

ManuelHentschel:
Fix #503. This PR updates the GitHub release action to automatically release the extension to GitHub and the marketplace when a tag matching `v*` is pushed.

This PR is kind of hard to check safely since the action is triggered automatically and produces a release to the marketplace, which can potentially affect all users of the extension. To be safe, I'd suggest keeping a `.vsix` file ready as backup when using the action for the first time.

For the action to publish to the marketlace, an access token is required, more details e.g. [here](https://github.com/marketplace/actions/github-action-for-vsce).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,5 +43,5 @@ jobs:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           files: "artifacts/*/*"
           prerelease: false
-          draft: true # Could also be false to publish the release immediately
+          draft: false
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 name: release
 
 # runs when a tag v* is pushed
-# creates a release draft with the binaries
+# releases the extension to GitHub and the vscode marketplace
 
 on:
   push:
@@ -45,3 +45,17 @@ jobs:
           prerelease: false
           draft: false
 
+
+  publish:
+    name: Publish
+    timeout-minutes: 30
+    needs: release # doesn't really need these, just to make sure that packaging etc worked
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: npm install
+      - uses: lannonbr/vsce-action@master
+        with:
+          args: "publish -p $VSCE_TOKEN"
+        env:
+          VSCE_TOKEN: ${{ secrets.VSCE_TOKEN }}


Ikuyadeu:
@ManuelHentschel Thank you for improving our releasing process. The changes look good to me.
To make security safety, I'll check the detail of secret tokens' access this weekend.

ManuelHentschel:
While we're on it, we could also consider publishing a development build of the extension, similar to what Microsoft does with their GiHub Pull Requests:
* [GitHub Pull Requests and Issues](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github)
* [GitHub Pull Request Nightly Build](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github-insiders)

This would allow people that are interested in the extension to follow development closely without having to manually install from .vsix files all the time.

Setting this up should be rather easy, we'd just need to add a step that sets `"preview": true,` in the package.json and adds a note to the README.

Do you think people would be interested in this?

Ikuyadeu:
OK, I set the secret, and I'll try it now.

Ikuyadeu:
@ManuelHentschel Before using the new function of vscode, we need to fix the source code for the new vscode API.
Usually, I did that work before publishing, but I am working on #508 because it will be large fixing in this time.
I hope you will help me.

PR #508, opened 2020-12-27
WIP: Pre fixing for version 1.6.3

Ikuyadeu:

Extends from #474

To publish a new version, we need to clean up the whole new source code.

I want to solve the following problems

* [x] Fix eslint problems on rHelps.ts and rHelpTree.ts by defining the interface
* [x] Apply for new cheerio API
* [x] Apply for new vscode API


ManuelHentschel:
I fixed the eslint warnings in the help related code and adapted it to the new cheerio/vscode APIs.
I left some of the `eslint-disable` statements in the code, in places where I don't see a better/safer way to achieve the same thing (e.g. when the entire line is wrapped in `try-catch` anyways).

The tests performed by GitHub actions seem to fail "randomly" sometimes and re-running solves the problem, no idea what's going on there...

Ikuyadeu:
@ManuelHentschel Thank you for your help!
OK, I merge this and publish a new version.

Issue #516, opened 2021-01-04
R helpPanel not able to list packages for "Open help topic using ?"


alex-gable:
**Describe the bug**
When clicking "Open help topic using `?`" an error is produced. 

Additionally, the error message advises that `r.helpPanel.rpath` ought to be set and pointing to a valid R executable. However, `r.helpPanel.rpath` is planned to be deprecated, so I'd imagine it should instead refer to `r.rpath.*` (in my case `* = mac`)

**To Reproduce**
1. Ensure all rpath settings are set to valid R executable
2. Open side panel
3. Click on "Open help topic using `?`"

setting.json

```json
  "r.bracketedPaste": true,
  "r.rpath.mac": "/usr/local/bin/R",
  "r.rterm.mac": "/Users/<me>/.local/bin/radian",
  "r.helpPanel.rpath": "/usr/local/bin/R",
  "r.sessionWatcher": true,
```

**Expected behavior**
Opens QuickPick with all available help aliases

**Screenshots**
You can show the keybord contents by pressing `F1` and `Developer: toggle screencast mode`
![image](https://user-images.githubusercontent.com/5963623/103589684-cd3d1b80-4ea0-11eb-9440-67def72fdca4.png)

**Environment (please complete the following information):**
 - OS: macOS
 - VSCode Version: 1.52.1
 - R Version: 4.0.3
 - vscode-R version: 1.6.3

renkun-ken:
Currently, the error message seems not helpful as the stderr of the R process is not captured. @ManuelHentschel Any idea how we could see what's happening in the background process?

ManuelHentschel:
@alex-gable Thanks for the detailed bug report! To get an idea what is happening you could try running the script [getAliases.R](https://github.com/Ikuyadeu/vscode-R/blob/master/R/getAliases.R) directly in R and check if/why it fails. Do the other help view related commands fail as well or is only this command causing problems?

[...]

PR #518, opened 2021-01-05
Better error message when reading aliases

ManuelHentschel:
Related to #516 

Shows a more detailed error message produced if the script `R/getAliases.R` fails.


renkun-ken (approved):
LGTM

It works nicely to capture the error message from the R session.

PR #521, opened 2021-01-06
Keep promises and active bindings in globalenv

### LINKED ISSUE #520: Should not evaluate delayed-assign object before it gets called
**Describe the bug**

VSCode-R evaluates the delayed-assign object too early. 

**To Reproduce**

1. Open VSCode (with VSCode-R installed)
2. Open the radian terminal in VSCode
3. Paste the following code in to the radian terminal

```r
delayedAssign("delayed_obj", {
  stop("should not execute")
})
```

### in VSCode's terminal

The code inside has been executed but it shouldn't as `delayed_obj` is not evaluated

image

### in a normal terminal

The code inside is executed only when `delayed_obj` gets evaluated

image

**Do you want to fix by self? (I hope your help!)**

Yes

**(If yes,) what kind of help do you want? (e.g. Which file should I fix, Survey (related documents)**

I want to know where to get started...

(If related)setting.json

```json
{
    "remote.SSH.connectTimeout": 30,
    "remote.SSH.maxReconnectionAttempts": 5,
    "remote.SSH.remotePlatform": {
        "10.4.119.58": "linux",
        "10.4.120.45": "linux"
    },
    "r.sessionWatcher": true,
    "r.rterm.windows": "D:\\app\\Python\\Python39\\Scripts\\radian.exe",
    "r.rpath.windows": "D:\\app\\R-4.0.0\\bin\\R.exe",
    "diffEditor.renderSideBySide": true,
    "[r]": {
        "editor.formatOnSave": false
    },
    "git.autofetchPeriod": 300,
    "terminal.integrated.shell.windows": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
    "window.zoomLevel": 0,
    "r.lsp.diagnostics": false,
    "git.confirmSync": false,
    "r.alwaysUseActiveTerminal": true,
    "r.bracketedPaste": true,
    "diffEditor.ignoreTrimWhitespace": false,
    "git.autofetch": true,
    "editor.tabSize": 2,
    "terminal.external.windowsExec": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
    "typescript.suggest.paths": false
}
```

**Expected behavior**

Should not evaluate the delayed-assign object so early - we should only evaluate the object when we really need it

**Screenshots**

See the comment in Reproduce Step

 - OS:  Win 10
 - VSCode Version: 1.52.1
 - R Version: 4.0.2
 - vscode-R version: 1.6.3

renkun-ken:

Close #520

This PR uses `rlang::env_binding_are_lazy()` and `rlang::env_binding_are_active()` to know which symbols are promises and active bindings so that they are not forced on task callback to write globalenv info.

**(If you have)Screenshot**

image

Run the following code in the global environment:

```r
x <- 1
delayedAssign("d", 1)
delayedAssign("err", stop("err"))
makeActiveBinding("a1", function() 1, environment())
```

No promise or active binding are forced.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -73,6 +73,7 @@ if (interactive() &&
       inspect_env <- function(env) {
         all_names <- ls(env)
         is_promise <- rlang::env_binding_are_lazy(env, all_names)
+        is_active <- rlang::env_binding_are_active(env, all_names)
         objs <- lapply(all_names, function(name) {
           if (is_promise[[name]]) {
             info <- list(
@@ -80,6 +81,13 @@ if (interactive() &&
               type = unbox("promise"),
               length = unbox(0L),
               str = unbox("<promise>")
+            )
+          } else if (is_active[[name]]) {
+            info <- list(
+              class = "active_binding",
+              type = unbox("active_binding"),
+              length = unbox(0L),
+              str = unbox("<active-binding>")
             )
           } else {
             obj <- env[[name]]
--- a/R/init.R
+++ b/R/init.R
@@ -1,652 +1,661 @@
-if (interactive() &&
-  Sys.getenv("RSTUDIO") == "" &&
-  Sys.getenv("TERM_PROGRAM") == "vscode") {
-  if (requireNamespace("jsonlite", quietly = TRUE) &&
-      requireNamespace("rlang", quietly = TRUE)) local({
-    # cleanup previous version
-    removeTaskCallback("vscode-R")
-    options(vscodeR = NULL)
+local({
+  if (interactive() &&
+    Sys.getenv("RSTUDIO") == "" &&
+    Sys.getenv("TERM_PROGRAM") == "vscode") {
 
-    .vsc.name <- "tools:vscode"
-    if (.vsc.name %in% search()) {
-      detach(.vsc.name, character.only = TRUE)
-    }
+    required_packages <- c("jsonlite", "rlang")
+    missing_packages <- required_packages[
+      !vapply(required_packages, requireNamespace,
+        logical(1L), quietly = TRUE)
+    ]
 
-    .vsc <- local({
-      pid <- Sys.getpid()
-      wd <- getwd()
-      tempdir <- tempdir()
-      homedir <- Sys.getenv(
-        if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"
+    if (length(missing_packages)) {
+      message(
+        "VSCode R Session Watcher requires ",
+        toString(missing_packages), ". ",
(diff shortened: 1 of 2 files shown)


R/init.R
```diff
@@ -1,7 +1,8 @@
 if (interactive() &&
   Sys.getenv("RSTUDIO") == "" &&
   Sys.getenv("TERM_PROGRAM") == "vscode") {
-  if (requireNamespace("jsonlite", quietly = TRUE)) local({
+  if (requireNamespace("jsonlite", quietly = TRUE) &&
+      requireNamespace("rlang", quietly = TRUE)) local({
```
shrektan, line 5: Do we need to throw a warning to alarm user that we need the two packages (if they are not installed). Otherwise, users may get confused.

shrektan, line 5:
```suggestion
  is_pkg_installed <- function(x) {
    not_installed <- x[!vapply(x,  requireNamespace, quietly = TRUE, FUN.VALUE = logical(1L))]
    if (length(not_installed)) message("VSCode-R depends on R package: ", toString(not_installed), ". Please install manually in order to use VSCode-R.")
    length(not_installed) == 0L
  }
  if (is_pkg_installed(c("jsonlite", "rlang"))) local({
```

renkun-ken: We already have the warning at the bottom of `init.R` but I haven't updated it yet.

renkun-ken: These functions will leak to user global environment. Let me put them in a local environment to avoid that.

R/init.R
```diff
@@ -1,87 +1,109 @@
...
-    .vsc <- local({
-      pid <- Sys.getpid()
-      wd <- getwd()
-      tempdir <- tempdir()
-      homedir <- Sys.getenv(
-        if (.Platform$OS.type == "windows") "USERPROFILE" else "HOME"
+local({
+  if (interactive() &&
+    Sys.getenv("RSTUDIO") == "" &&
+    Sys.getenv("TERM_PROGRAM") == "vscode") {
```
shrektan, line 4: Just a trival issue, hope you don't mind:  

Will rewrite this like below better? as it reduces the layer of the code and it clearly states that we only execute the code under such condition and eval in a `local()` environment.

```r
 if (interactive() &&
    Sys.getenv("RSTUDIO") == "" &&
    Sys.getenv("TERM_PROGRAM") == "vscode") local({
 # code 
 })

renkun-ken: Yes you are right. It looks better now.

ManuelHentschel (approved):
LGTM, thanks!

Issue #479, opened 2020-11-30
Refactor extension.ts


Ikuyadeu:
Extends from #474 

`extension.ts` is literally the enter of our extension.
However, it is too complicated for newcomers.

As the first step, I want to solve the following problems.

* [ ] Remove goto label it can be simple for-statement
* [ ] Move Rmd functions to another file
* [ ] Move [completion related functions](https://github.com/ManuelHentschel/vscode-R/blob/522a21df283a2bf1516763b4a3f4415a0c6d7dc8/src/extension.ts#L296:L419) to their own file
* [ ] Move `saveDocument()` to `utils.ts`.

ManuelHentschel:
This is definitely a good idea!

Adding to your points, I'd strongly suggest moving the [these commands](https://github.com/ManuelHentschel/vscode-R/blob/522a21df283a2bf1516763b4a3f4415a0c6d7dc8/src/extension.ts#L140:L224) to rTerminal.ts. I don't think they need to be implemented in the scope of the `activate()` function and thematically belong to the functions in rTerminal.ts anyways.

Similarly, I'd suggest moving the [completion related functions](https://github.com/ManuelHentschel/vscode-R/blob/522a21df283a2bf1516763b4a3f4415a0c6d7dc8/src/extension.ts#L296:L419) to their own file and [`saveDocument()`](https://github.com/ManuelHentschel/vscode-R/blob/522a21df283a2bf1516763b4a3f4415a0c6d7dc8/src/extension.ts#L123:L138) to utils.ts.

This would make it easier for new contributors to get an overview over what the extension does, since only the individual contributions to vscode remain in extension.ts/`activate()` and the implementation details of each feature are in their own file.

Ikuyadeu:
@ManuelHentschel Thank you, I added your ideas.
We already have four change ideas, and it should be fixed by separated PR.

Ikuyadeu:
Close by #525

PR #525, opened 2021-01-15
Refactor extension.ts

ManuelHentschel:
Implements the refactoring suggested in #479.

This PR is not intended to change the behavior of the extension. 

I also changed the imports at the top of `extension.ts` to namespace imports:
* Improves the readibility of the code
* Makes it easier to keep track of where each function comes from
* Makes life easier for developers. E.g.
    * you can use intellisense to access imported functions (e.g. `vscode.` + `ctrl+space` to view all exported variables from vscode) 
    * it's easier to move code around between files (you only need to add/remove the namespace imports sometimes, previously you had to add/remove each function individually)
    * Fewer (redundant) changes in Git diffs

If you disagree, this can of course be reverted, but I'd strongly suggest using namespace imports here.

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -1,0 +1,151 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+
+
+import * as vscode from 'vscode';
+
+import * as session from './session';
+
+
+// Get with names(roxygen2:::default_tags())
+const roxygenTagCompletionItems = [
+    'export', 'exportClass', 'exportMethod', 'exportPattern', 'import', 'importClassesFrom',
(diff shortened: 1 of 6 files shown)


Ikuyadeu:
> Improves the readibility of the code
> Makes it easier to keep track of where each function comes from

@ManuelHentschel Great works! Our source code looks more cleaned.

> you can use intellisense to access imported functions (e.g. vscode. + ctrl+space to view all exported variables from vscode)
> it's easier to move code around between files (you only need to add/remove the namespace imports sometimes, previously you had to add/remove each function individually)

Good, I denied it before, but it may be better to change the way of thinking.

> Fewer (redundant) changes in Git diffs

In other words, we can see the files that got the impact of the main changes.
How about using an asterisk for external packages such as vscode and individual imports for vscode -r packages such as `session` and `util`?

ManuelHentschel:
>  How about using an asterisk for external packages such as vscode and individual imports for vscode -r packages such as `session` and `util`?

That sounds like a good rule of thumb, although I would still prefer to use namespace imports for vscode-r files when the number of imported functions is rather large (as e.g. in extension.ts).

Ikuyadeu:
@ManuelHentschel I agree with you.
And I'll update flexible eslint rules after.
Link: https://eslint.org/docs/rules/no-restricted-imports

Anyway, I'll merge the PR that is enough good.

PR #526, opened 2021-01-18
(Maybe) Fix Aliases

### LINKED ISSUE #516: R helpPanel not able to list packages for "Open help topic using ?"

ManuelHentschel:
This PR might fix #516 .

This PR changes the code used to retrieve a list of aliases from R. Now the aliases are written to a temporary file and read from there. Previously the aliases were transmitted using stdout, which could lead to a buffer overflow for large numbers of packages.

### DIFF
--- a/src/rHelpProvider.ts
+++ b/src/rHelpProvider.ts
@@ -5,6 +5,9 @@ import * as cp from 'child_process';
 import * as kill from 'tree-kill';
 
 import * as rHelp from './rHelp';
+import * as path from 'path';
+import * as fs from 'fs';
+import * as os from 'os';
 
 
 export interface RHelpProviderOptions {
@@ -153,6 +156,8 @@ export class HelpProvider {
 export interface AliasProviderArgs {
 	// R path, must be vanilla R
 	rPath: string;
+    // cwd
+    cwd?: string;
 	// getAliases.R
     rScriptFile: string;
     
@@ -172,6 +177,7 @@ interface PackageAliases {
 export class AliasProvider {
 
     private readonly rPath: string;
+    private readonly cwd?: string;
     private readonly rScriptFile: string;
     private allPackageAliases?: null | {
         [key: string]: PackageAliases;
@@ -181,6 +187,7 @@ export class AliasProvider {
 
     constructor(args: AliasProviderArgs){
         this.rPath = args.rPath;
+        this.cwd = args.cwd;
         this.rScriptFile = args.rScriptFile;
         this.persistentState = args.persistentState;
     }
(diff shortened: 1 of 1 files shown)


src/rHelpProvider.ts
```diff
@@ -284,16 +291,26 @@ export class AliasProvider {
         this.allPackageAliases = null; 
 		const lim = '---vsc---'; // must match the lim used in R!
 		const re = new RegExp(`^.*?${lim}(.*)${lim}.*$`, 'ms');
-        const cmd = `${this.rPath} --silent --no-save --no-restore --slave -f "${this.rScriptFile}"`;
+        const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'vscode-R-aliases-'));
+        const tempFile = path.join(tempDir, 'aliases.json');
+        const cmd = `${this.rPath} --silent --no-save --no-restore --slave -f "${this.rScriptFile}" > "${tempFile}"`;
```
renkun-ken, line 296: Does this work on Windows?

src/rHelpProvider.ts
```diff
@@ -284,16 +291,26 @@ export class AliasProvider {
         this.allPackageAliases = null; 
 		const lim = '---vsc---'; // must match the lim used in R!
 		const re = new RegExp(`^.*?${lim}(.*)${lim}.*$`, 'ms');
-        const cmd = `${this.rPath} --silent --no-save --no-restore --slave -f "${this.rScriptFile}"`;
+        const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'vscode-R-aliases-'));
```
renkun-ken, line 294: Is there a reason to create a temp dir rather than a simple temp file?

src/rHelpProvider.ts
ManuelHentschel: It does work on my windows machine, and the `>` operator seems to be part of cmd/powershell, so I think it should work.

src/rHelpProvider.ts
ManuelHentschel: Just the convenience of using `fs.mkdtempSync` to ensure we have a valid, unique path. Using a static filename might cause problems if multiple vscode instances are running simultaneously and generating a random filename (and checking its availability) would make things more complicated.

src/rHelpProvider.ts
renkun-ken: Thanks for your confirmation!

src/rHelpProvider.ts
renkun-ken: Thanks for the clarification!

renkun-ken (approved):
LGTM

Issue #529, opened 2021-01-21
Add Update Packages to help pages viewer


renkun-ken:
Now we have the following commands in the help viewer

<img width="243" alt="image" src="https://user-images.githubusercontent.com/4662568/105271728-fe248c80-5bd2-11eb-9d4a-a303b654c987.png">

Does it make sense to add "Update packages" command?

Ikuyadeu:
At least adding the function will not be a problem, but it's better to create it under the package tree for the future.

Ikuyadeu:
Fixed by #532

PR #532, opened 2021-01-22
Add updatePackage command

ManuelHentschel:
#529 

This PR adds a command `Update/reinstall package` to the package tree view that reinstalls/updates the package. This is done by a simple call to `install.packages('PKG_NAME')` and does not check if the package is installed from CRAN and does not compare versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -580,6 +580,11 @@
         "icon": "$(trash)"
       },
       {
+        "title": "Update/reinstall Package",
+        "command": "r.helpPanel.updatePackage",
+        "icon": "$(cloud-download)"
+      },
+      {
         "title": "Install multiple packages",
         "command": "r.helpPanel.installPackages",
         "icon": "$(expand-all)"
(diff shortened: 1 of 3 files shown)


renkun-ken:
Looks like the update button already works with single package.

I'm wondering if it makes sense to add similar to 

image

so that it is easier to update all installed packages by executing something like `update.packages(ask=FALSE)`?

ManuelHentschel:
I added a command button to the entry `Install CRAN Package` (I felt it fits better there, since the other commands next to `Help topics by Package` only affect the tree view but don't install/modify any actual packages).

As you suggested, the command executes `update.packages(ask=FALSE, repos=<CRAN_URL>)`. On my machine this worked for most packages but some packages seem to cause problems. I think this is caused by them being loaded somewhere, so that their .DLL files cannot be overwritten. I don't think there is a simple solution for this, so I wouldn't try to fix this, but if you have any ideas, feel free to push/suggest a fix.

renkun-ken:
>I added a command button to the entry Install CRAN Package (I felt it fits better there, since the other commands next to Help topics by Package only affect the tree view but don't install/modify any actual packages).

Yes, makes good sense.

>As you suggested, the command executes update.packages(ask=FALSE, repos=<CRAN_URL>). On my machine this worked for most packages but some packages seem to cause problems. I think this is caused by them being loaded somewhere, so that their .DLL files cannot be overwritten. I don't think there is a simple solution for this, so I wouldn't try to fix this, but if you have any ideas, feel free to push/suggest a fix.

I think it is a known issue in Windows. I guess we could leave it as is at the moment.

renkun-ken (approved):
LGTM

PR #535, opened 2021-01-23
add sendToConsole to rstudioapi emulation

MilesMcBain:

Adds support for [`rstudioapi::sendToConsole`](https://rstudio.github.io/rstudioapi/reference/sendToConsole.html) to VSCode.

Previously I had thought that this was probably not needed, but now it's attractive to support: https://github.com/MilesMcBain/breakerofchains.

The function is a bit more limited than the RStudio version since we can't easily implement the `execute = FALSE` and `echo = FALSE` options (sending to terminal without running or terminal session without echoing). The function errors with informative message if `TRUE` is supplied for these args.

Try the following code:

```r
rstudioapi::sendToConsole("version")
rstudioapi::sendToConsole("version", focus = TRUE)
rstudioapi::sendToConsole("version", execute = FALSE) #error
rstudioapi::sendToConsole("version", echo = FALSE) #error
rstudioapi::sendToConsole(c(      
    "library(dplyr)",
      "mtcars %>%",
      "summary()",
      "",
      "starwars %>%",
      "group_by(species, sex) %>%",
      "select(height, mass) %>%",
      "summarise(",
      "height = mean(height, na.rm = TRUE),",
      "mass = mean(mass, na.rm = TRUE))"))
```

### DIFF
--- a/R/rstudioapi.R
+++ b/R/rstudioapi.R
@@ -240,6 +240,19 @@ versionInfo <- function() {
         release_name = "vscode"
     )
 }
+
+sendToConsole <- function(code, execute = TRUE, echo = TRUE, focus = FALSE) {
+    if (!execute || !echo) {
+          stop("rstudioapi::sendToConsole only supports execute = TRUE and echo = TRUE in VSCode.")
+      }
+
+    code_to_run <- paste0(code, collapse = "\n")
+    invisible(
+        rstudioapi_call("send_to_console", code = code_to_run, focus = focus)
+    )
+}
+
+
 # Unimplemented API calls that will error if called.
 
 .vsc_not_yet_implemented <- function(...) {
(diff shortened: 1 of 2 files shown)


renkun-ken (approved):
The tests work as expected. LGTM.

krlmlr:
Thanks for providing this, needs to be added to the wiki: https://github.com/Ikuyadeu/vscode-R/wiki/RStudio-addin-support. I can't edit it.

krlmlr:
@randy3k: What would it take to support `sendToConsole(execute = FALSE)` ?

randy3k:
I don't think it has anything to do with radian, `vscode-R` just needs to send the strings to the radian console without an "enter".

krlmlr:
Got you, we need `sendText(code, false)`:

https://code.visualstudio.com/api/references/vscode-api#Terminal

PR #544, opened 2021-02-01
Change workspace tooltip

ElianHugh:
Change workspace tooltip from showing length (e.g. length of 5) to showing size in bytes. does not apply to active-bindings or promises

At the moment, the workspace viewer tooltip shows the length of a given object. This is a little redundant given that this can be inferred from the object description (e.g. `data.frame: 32 obs. of 11 variables`). 

As a result, this PR changes the tooltip to show the memory size of the object when possible, bringing the workspace viewer closer in behaviour to RStudio's environment pane. Object sizes are returned for all variables, barring promises and active-bindings. 

One difference between the PR and RStudio's environment pane is that the environment pane does not show sizes for numeric lists (e.g. `x <- 1:5`), whereas this PR does. 

**Screenshot**
*VScode-R*
image

*RStudio*
image

Object sizes should be shown in bytes for all eligible object types

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -43,7 +43,10 @@ if (interactive() &&
         format.default(Sys.time(), nsmall = 6)
       }
 
-      unbox <- jsonlite::unbox
+      scalar <- function(x) {
+        class(x) <- c("scalar", class(x))
+        x
+      }
 
       request <- function(command, ...) {
         obj <- list(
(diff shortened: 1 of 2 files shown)


renkun-ken (approved):
Works as expected.

LGTM. Thanks!

Issue #540, opened 2021-01-28
Increase or make configurable level of utils::str used for session global env watcher?


MilesMcBain:
I like the fact that an object summary is shown on the hover but I find it a bit brief, particularly for data.frames. E.g.

```
tibble [100 x 4] (S3: tbl_df/tbl/data.frame)
```

Would it be possible to increase the level of description or make it configurable via an option?

From my reading of the code it would be a matter of changing the `max.level` in the call to utils::str here: https://github.com/Ikuyadeu/vscode-R/blob/master/R/init.R#L63

For example, if this is set to 2, it gives a useful summary of data frame columns:

```
tibble [100 x 4] (S3: tbl_df/tbl/data.frame)
 $ h3_index          : chr [1:100] "889d4a0643fffff" "889c236007fffff" "88be8d10a9fffff" "889c236001fffff" ...
 $ n_incidents       : int [1:100] 128 155 465 112 610 1675 537 43 31 77 ...
 $ mean_residual_mins: num [1:100] 18.96 10.88 2.81 9 1.59 ...
 $ score             : num [1:100] 2427 1686 1305 1008 967 ...
```

I realise with big data frames this may make for a large amount of hover information. It's not so much of an issue for me because I use another extension which shows the hover in a view at the side. Probably an argument for making an option.

Any interest in a PR?

renkun-ken:
Sure, you could file an PR making it an option?

PR #545, opened 2021-02-01
add option vsc.hover.str.max.level

MilesMcBain:

Added an option, `vsc.hover.str.max.level`, that configures how R objects hover information is displayed. This is the `max.level` arg in the `str` call in `capture_str` in `init.R`. It was previously set to `0` which just captured type/dimension info for data frames.

Now the user can configure it to show more info for data frames, e.g. `vsc.hover.str.max.level = 2`:

df_preview

Try this example code with  `vsc.hover.str.max.level = 0` and  `vsc.hover.str.max.level = 2` Hover over the variables and notice the improved info, without becoming overly large:

```r

library(nycflights13)

foo <- flights

bar <- airports

biz <- letters

baz <- as.list(letters)

var <- list(first = "a", 
            second = "b",
            third = list("c", list("d")))

```

Note I also added the `vec.len = 1` arg into the `str` call. It has no effect at the default level.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -59,8 +59,15 @@ if (interactive() &&
       }
 
       capture_str <- function(object) {
-        utils::capture.output(
-          utils::str(object, max.level = 0, give.attr = FALSE)
+        paste0(
+          utils::capture.output(
+            utils::str(object,
+              max.level = getOption("vsc.hover.str.max.level", 0),
+              give.attr = FALSE,
+              vec.len = 1
+            )
+          ),
+          collapse = "\n"
         )
       }
 
--- a/R/init.R
+++ b/R/init.R
@@ -62,7 +62,7 @@ if (interactive() &&
         paste0(
           utils::capture.output(
             utils::str(object,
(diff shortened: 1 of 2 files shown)


MilesMcBain:
Adressing #540

MilesMcBain:
I still want to just download the VSIX from this branch to check for any interaction with the LSP extension. Probably don't merge until I've done that.

MilesMcBain:
No issues when using with the LSP extension.

renkun-ken:
This string will appear not only in the hover, but also in the description of the completion items of session symbols. 

image

I guess we could name this option `vsc.str.max.level`?

MilesMcBain:
Whoah... I had no idea this was a thing. It seems it never shows for me unless I click the little 'Read more' arrow next to the df in the completion list, and then it always shows. Nice!

New name sounds good.

MilesMcBain:
Okay done, the option is now `vsc.str.max.level`.

renkun-ken (approved):
LGTM

PR #556, opened 2021-02-16
Refactoring and implementation of webviewPanelSerializer

ManuelHentschel:
This PR mostly contains refactoring/cleanup of help related files.
The only change to the extension's behaviour is the implementation of `rHelp.deserializeWebviewPanel`

https://github.com/Ikuyadeu/vscode-R/blob/c4c6334798b694185619f85d321fd92e1880fa39/src/rHelp.ts#L180-L184

implementing `vscode.WebviewPanelSerializer`.

Open some help pages and reload/restart vscode, check if the help pages are still open

### DIFF
--- a/html/script.js
+++ b/html/script.js
@@ -33,6 +33,8 @@ window.document.body.onload = function () {
     var _a;
     // make relative path for hyperlinks
     var relPath = (document.body.getAttribute('relPath') || '');
+    // notify vscode, used to restore help panels between sessions
+    vscode.setState(relPath);
     var loc = document.location;
     var url0 = new URL(loc.protocol + '//' + loc.host);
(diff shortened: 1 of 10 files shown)


renkun-ken:
If a single Help Page is open, it will recover after reloading.

If multiple Help Pages are open, only the last one will recover and others are blank. Do you think we need to handle this so that all help pages could be recovered?

ManuelHentschel:
> If multiple Help Pages are open, only the last one will recover and others are blank. Do you think we need to handle this so that all help pages could be recovered?

Thanks for catching that! Should be fixed now

renkun-ken:
Thanks! It works properly with multiple help tabs now.

renkun-ken (approved):
LGTM

Issue #555, opened 2021-02-16
Implement shared R workspace in VS Live Share


gvelasq:
**Is your feature request related to a problem? Please describe.**
Following our recent [post](https://ivelasq.rbind.io/blog/vscode-live-share/) and @ivelasq's [tweet](https://twitter.com/ivelasq3/status/1357136366568689664) trialing VS Code and VS Live Share for remote pair programming in R, one of the features we hoped could be implemented in `vscode-R` is a shared R workspace across host and guest(s) sessions in [VS Live Share](https://marketplace.visualstudio.com/items?itemName=MS-vsliveshare.vsliveshare).

**Describe the solution you'd like**
Implementation of a shared R workspace while using VS Live Share. Per the [issue](https://github.com/MicrosoftDocs/live-share/issues/4098) I filed in the VS Live Share repository, this could be implemented using the [VS Live Share Extension API](https://www.npmjs.com/package/vsls).

**Describe alternatives you've considered**
Currently the host and guest(s) have separate R workspaces. This is useful behavior when host and guest(s) are inspecting the codebase and objects separately. However, sometimes to debug a problem it is useful for users to inspect the object(s) in one another's R workspace directly.

**Additional context**
I initially filed an issue [here](https://github.com/MicrosoftDocs/live-share/issues/4098) and was directed to this repository as a shared R workspace is considered custom extension behavior.

ElianHugh:
I've only skimmed the live share API, but this might be possible using the URI conversion exposed by it. At the moment, the workspace viewer relies on a globalenv.json file to populate its contents,  so this should theoretically be doable if we expose the file to the guest session. 

As I haven't used live share, I'm not sure if there would be read/write issues, but presumably there is a method of resolving this if that were the case.

If no one is eager to fiddle with this, I can have a further look!

ElianHugh:
Some progress!

![image](https://user-images.githubusercontent.com/60372411/109295989-8e37a080-7827-11eb-97e8-6151edc1b03f.png)

ElianHugh:
Unfortunately, I think this is going to need a little more work. It seems that whilst uri conversion works, the files themselves are not actually exposed to the guest session. I'll update when I have more progress :)

gvelasq:
@ElianHugh, thank you very much for your efforts. TypeScript is outside of my expertise but I did have a few questions:

[...]

PR #564, opened 2021-03-02
LiveShare functionality

ElianHugh:
Solves #555. Shares global environment between host and guest sessions, allowing for workspace viewer to function for guests. May need more testing to ensure 100% functionality (I would appreciate any help with testing, as I've been purely testing on a local setup). The commits should probably be squashed as well.

As an aside, this adds vsls as a dependency of vscode-r. I feel like this probably isn't ideal, are optional dependencies a thing?

TODO/Issues: 
* run .vsc.attach() on guest join
  * guest terminals are not attached unless either A) the terminal is created after they have joined or B) .vsc.attach() is run after they join
 * using the LiveShare test join feature seems to incorrectly attach the terminal for guest sessions
    * likely due to the inability to distinguish between the test session + main session
 * View() seems to sometimes not open files on the guest session
    * possibly a pathing issue? Can't replicate consistently

**Screenshot**

*Guest Session*
guest

*Host Session*
host

Try LiveShare between two vscode sessions. Workspace viewer *should* be fully functional, and sending text to terminal should work as well.


ElianHugh:
I've had a chance to look at cross-computer live share and it looks like the current functionality doesn't apply there. I'm going to take this PR down until I finish working on an RCC that exposes the files appropriately!

PR #566, opened 2021-03-05
Add option to keep terminal hidden after running code

### LINKED ISSUE #510: Send the code to the terminal without automatic toggling of terminal pane.
Hi, this might be a little bit off-topic. So I just found out that I can launch into the same TMUX window from two separate terminals instance. Basically, this makes it possible to send code to the integrated terminal, but read it from iTerm on a different monitor. It kind of a work-around for the long-standing multi-screen support issue of VScode.

So the only problem here is that, when I hit cmd+enter to send code to integrate terminal, the terminal pane will still be automatically toggled even if it is hidden. I'm wondering if there is any modification I can do to make it "silently" send the code to the terminal without actually toggling the terminal window in vscode?

andycraig:
Closes #510 

This PR adds an option 'none' to `r.source.focus` to keep the terminal hidden after sending code to the terminal.

With session watcher on.

1. Add this line to `settings.json`:

```json
    "r.source.focus": "none"
```

2. Run command `R: Create R terminal`
3. Hide the terminal by clicking the X at the top right
4. In a file 'temp.R', write this code and send it to the terminal: `plot(1, 1)`
5. Observe that the plot window appears (so the code was executed), but the terminal stays hidden

**Note**

I believe `r.source.focus` is the right setting for this option, but I am concerned that the difference between the options 'editor' and 'none' is not obvious. I have updated the description with this in mind so please check whether you think it is clear enough.

### DIFF
--- a/package.json
+++ b/package.json
@@ -958,7 +958,7 @@
             "terminal",
             "none"
           ],
-          "description": "Keeping focus when running."
+          "description": "What to show/focus after running code. Set to 'editor' or 'terminal' to show the terminal. Set to 'none' to not show the terminal."
         },
         "r.alwaysUseActiveTerminal": {
           "type": "boolean",
--- a/package.json
+++ b/package.json
@@ -955,7 +955,8 @@
           "default": "editor",
           "enum": [
             "editor",
-            "terminal"
+            "terminal",
+            "none"
           ],
           "description": "Keeping focus when running."
         },
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -265,5 +265,7 @@ export async function runTextInTerm(text: string): Promise<void> {
 
 function setFocus(term: vscode.Terminal) {
     const focus: string = config().get('source.focus');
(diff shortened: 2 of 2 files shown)


renkun-ken:
Works as expected.

It looks like we have two ways to address this:

1. Two options to control what to focus and whether to reveal terminal. But I guess nobody really needs "focus on the terminal but not reveal it".
2. One option (like implemented in this PR) to control both behavior.

From the perspective of improving an existing option rather than introducing a new one, I guess the one option makes more sense to me.

renkun-ken (approved):
LGTM

andycraig:
@renkun-ken Thank you for the review! Exactly, I agree that there's two possible approaches here. In the issue discussion I initially proposed a new setting. Then I changed my mind and suggested it be added as a new option to the existing setting. As you note, it's hard to imagine anyone wanting the combination 'focus on the terminal but don't reveal it', and this way  we avoid people choosing that combination accidentally.

I will go ahead and merge.

PR #567, opened 2021-03-05
Fix so code can be run after creating terminal 

### LINKED ISSUE #483: Have to press ctrl+enter two times to "Run Command in Terminal" when set to radian
- VSCode Version: 
  Version: 1.52.0-insider (system setup)
  Commit: ac165d7f974d70212be80cb15eaecfc1fd20b21a
  Date: 2020-11-30T17:36:25.607Z
  Electron: 11.0.2
  Chrome: 87.0.4280.63
  Node.js: 12.18.3
  V8: 8.7.220.24-electron.0
  OS: Windows_NT x64 10.0.19042

- VSCode-R Version: 1.6.1
- OS Version: Windows 10

Steps to Reproduce:

1. Set `radian` as your default Rterm.
1. Press `CTRL+Enter` on any selected R command to `Run Command in Terminal`.

Unlike `R.exe`, for radian we need to press `CTRL+Enter` two times (first one just opens radian in terminal; second one sends command to the console).

Do you want to fix by self? (I hope your help!)

No

My `settings.json` is like below:

``` json
{
    "r.sessionWatcher": true,
    "r.rterm.windows": "C:/ProgramData/Miniconda3/Scripts/radian.exe",
    "r.bracketedPaste": true,
    "r.rterm.option": [
        "--no-save",
        "--no-restore",
        "-q"
    ],
    "r.source.focus": "terminal",
    "r.lsp.path": "C:/program files/R/R-4.0.3/bin/x64/R.exe",
    "r.lsp.debug": true
}
```

### LINKED ISSUE #563: There maybe a bug with r.runSource / Run Source
Hi! There maybe a bug with r.runSource / Run Source.
**Description**
I'm a beginner in Coding. After I configured this extension for R language on VSCode, I found a bug in the process of using it.  
When I finish a script in R language, naturally, I want to run the script on the terminal. So I used the `r. runsource` command. But the code didn't run as expected. Instead, VSCode just opens an R terminal. When I use the `r. runsource` command for the second time, the terminal runs the script.
PS: I opened a folder which has R source file.

**Steps to reproduce the behavior**
1. Open VSCode without starting any terminal
2. Write an R script (eg.: print("demo"))
3. `Ctrl + Shift + S` (An R Terminal named `R interactive` appears)
4. `Ctrl + Shift + S` again (The Terminal shows [1] "demo")

**Do you want to fix by self? (I hope your help!)**
I'd love to, but I'm not that good at coding and I've never been developing a VSCode extension.

**Expected behavior**
When I use `r.runSource`, it should start an `R interactive` terminal and run the script. I just need to press `Ctrl + Shift + S` for once.  

**Screenshots**
screenshot

 - OS: Windows 10 20H2
 - VSCode Version: 1.53.2
 - R Version: 4.0.4
 - vscode-R version: 1.6.4

**Additional context**
If this is a bug, I hope you can fix it as soon as possible.
If this is not a bug, it would be great if you could tell me how to set up this extension to implement this function or develop this new function in the next version!
Thank you very much!

andycraig:
Fixes #563 

Fixes #483 

Trying to create a terminal and run code at the same time results in a new terminal but does not run the code. The user has to run the command again to send the code to the terminal. This PR fixes that.

The problem was that even when terminal was successfully created, `success` was `undefined` and so `chooseTerminal` would always return `undefined`. Since failure to create a terminal should result in `rTerm` being `undefined` anyway, it should be safe to remove `success` and always return `rTerm`.

1. Create a file `temp.R` containing the code `x <- 1`
1. Disable setting `R: Always Use Active Terminal`
1. Run command `R: Run Selection/Line`. Observe that a new R terminal is created and that `x <- 1` is sent to that terminal
1. Close the R terminal by clicking the trash icon
1. Run command `R: Run Source`. Observe that a new a new R terminal is created and that `source(...)` is sent to that terminal
1. Close the R terminal by clicking the trash icon
1. Check that it still fails gracefully if there is a problem creating the terminal:
    1. Change setting `R: Rterm Linux/Mac/Windows` to a nonsense path like 'abc'
    2. Run command `R: Run Selection/Line`. Observe appropriate error message 'Cannot find R client ...'

**Note**

This is the output I get when using `R: Run Selection/Line`. Note the doubled `x <- 1`. I could avoid that by increasing the delay amount but the amount required is presumably going to vary according to how fast the user's PC is. I've left it as-is for now. We could add a setting if it bothers people.

```
Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

x <- 1
> x <- 1
> 
```

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -206,11 +206,8 @@ export async function chooseTerminal(): Promise<vscode.Terminal> {
     }
 
     if (rTerm === undefined) {
-        const success = await createRTerm(true);
+        await createRTerm(true);
         await delay(200); // Let RTerm warm up
-        if (!success) {
-            return undefined;
-        }
     }
 
     return rTerm;
--- a/src/rstudioapi.ts
+++ b/src/rstudioapi.ts
@@ -344,7 +344,9 @@ export async function sendCodeToRTerminal(code: string, focus: boolean) {
   await runTextInTerm(code);
   if (focus) {
     const rTerm = await chooseTerminal();
-    rTerm.show();
+    if (rTerm !== undefined) {
+      rTerm.show();
+    }
   }
 }
 


renkun-ken (approved):
LGTM

Works as expected.

andycraig:
@renkun-ken Thanks for the review! I was about to merge this but looking at the recent commit history there seem to be fewer merge commits than there used to be. Are we favouring 'Squash and merge' over 'Create a merge commit'?

renkun-ken:
All recent PRs were merged with "Squash and merge". @Ikuyadeu Do we have a preference on this? I'm totally ok with both if the PR is clean and short.

Ikuyadeu:
@andycraig  Thank you for your great work.

@renkun-ken Thank you for your every review and discussion.
>  I'm totally ok with both if the PR is clean and short.

I agree with you. It depends on the reviewers' feelings. 
In my opinion, two methods can be used for the following pourpose

* `Squash and Merge`: Reduce the tangled commits from history to avoid hiding other commits
* `Merge`: Show the separated commits for the maintenance and revert one commit

Additionally, like this PR, @andycraig 's two commits have separated works.
I think `Merge` is better in this case for checking code history.

andycraig:
@Ikuyadeu @renkun-ken Thanks both!

PR #578, opened 2021-03-09
Add links to help pages in hover

ManuelHentschel:
This PR adds a hover provider that that shows links to matching help pages when hovering over names of R functions.

Hover over the name of a (documented) R function, wait for hover to show up and check if a link to a help page shows up.

**Limitations:**
On my machine the link shows up below other hovers, which can be a bit cumbersome if a large help page is shown e.g. by the languageserver.
The hover can be a bit eager, e.g. showing `dplyr::n` when hovering over a single `n`. This would be rather hard to solve properly, since the hover is unaware of any tokenization or other languageserver functions. If this is too annoying, the hover can be disabled completely using `r.helpPanel.enableHoverLinks = false`

### DIFF
--- a/package.json
+++ b/package.json
@@ -615,6 +615,11 @@
         "title": "Open help for selection",
         "category": "R",
         "command": "r.helpPanel.openForSelection"
+      },
+      {
+        "title": "Open help for path",
+        "category": "R",
+        "command": "r.helpPanel.openForPath"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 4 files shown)


renkun-ken (approved):
LGTM

renkun-ken:
@ManuelHentschel I guess we could move this feature to CodeActions so that hover from languageserver will not already appear above these help links?

renkun-ken:
Looks like the CodeActions may not be a proper place to do this as it does nothing to the source code.

ManuelHentschel:
> @ManuelHentschel I guess we could move this feature to [CodeActions](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) so that hover from languageserver will not already appear above these help links?

I didn't manage to add the help links anywhere in the language server, since commands in markdown links only work if [`MarkdownString.isTrusted`](https://github.com/microsoft/vscode/blob/1436b523d4aaa773ab246daa68daaa70cef48b1d/src/vs/vscode.d.ts#L2493-L2497) is set to `true` somehow, which I don't think is possible (and probably wouldn't be a good idea anyways)

renkun-ken:
I tried using CodeActions to do this, and it looks better I think.

image

ManuelHentschel:
That does look nice. How does the communication between languageserver and helppanel work here?
I only tried command-links in markdown, which are affected by the `isTrusted` problem I mentioned above.

Providing these links from within the languageserver might also allow us to only show relevant links (e.g. if the ls knows, which package the function is from, or if the currently selected work is inside a string).

renkun-ken:
It is not provided by languageserver. I just switch from `HoverProvider` to `CodeActionProvider` in vscode-R.

renkun-ken:
In this case, I'm not sure if the setting `helpPanel.enableHoverLinks` is useful anymore? Maybe we could rename it to `helpPanel.enableCodeActions` or something more obvious? Or we simply remove it until somebody really needs such a setting?

PR #579, opened 2021-03-13
Provide help links in CodeActions

renkun-ken:

This PR moves the help links provided in hover (#578) to CodeActions so that they appear more clearly, especially along with languageserver which already shows help documentation in hover.

An added value of this is that user could use keyboard only to move cursor around and press a shortcut key (e.g. cmd+.) to show the CodeActions menu and choose the action, in this case, show help for a particular function.

**(If you have)Screenshot**

image

* Type some code and move cursor to a token, a lightbulb will show near the token.
* Click the lightbulb and all aliases are listed.
* Click one aliases and the help viewer will show up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -951,10 +951,15 @@
           "markdownDeprecationMessage": "Will be deprecated. Use `#r.rpath.windows#`, `#r.rpath.mac#`, or `#r.rpath.linux#` instead.",
           "deprecationMessage": "Will be deprecated. Use r.rpath.windows, r.rpath.mac, or r.rpath.linux instead."
         },
-        "r.helpPanel.enableHoverLinks": {
-          "type": "boolean",
-          "default": true,
-          "markdownDescription": "Show links to matching help pages in hover"
+        "r.helpPanel.showHelpLinks": {
+          "type": "string",
+          "default": "hover",
+          "enum": [
+            "hover",
+            "codeActions",
+            "none"
+          ],
+          "description": "Where to show links to matching help pages in editor."
         },
         "r.source.encoding": {
           "type": "string",
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -12,6 +12,7 @@ import { extendSelection } from './selection';
 import { cleanLine } from './lineCache';
 import { globalRHelp } from './extension';
 import { config } from './util';
+import { activeEditorContext } from './rstudioapi';
 
 
 // Get with names(roxygen2:::default_tags())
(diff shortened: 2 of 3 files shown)


ManuelHentschel:
Showing the links as code actions does look nice and is more prominent than the hover-links. However, this change might be a bit counterintuitive since the code actions show up in a different place than other documentation (the hover provided by the languageserver) and opening documentation doesn't strictly match the description of code actions in the vscode API:
> A code action represents a change that can be performed in code, e.g. to fix a problem or to refactor code.

So maybe we could let the user choose by using a setting like `r.helpPanel.showHelpLinks = "codeActions"|"hover"|"none"`?

renkun-ken:
When editing vscode-R code, I notice that

image

Looks like code action is already used to show some documentation or open external links here. I think code action might be a better place to show these show-help commands especially when we add more code actions like `Change group to ggplot2::group` or `Use library(ggplot2)`.

> So maybe we could let the user choose by using a setting like r.helpPanel.showHelpLinks = "codeActions"|"hover"|"none"?

That's indeed doable, but I think it over-complicates things. Hover is a bit overloaded, by both languageserver and sesson watcher here.

ManuelHentschel:
> That's indeed doable, but I think it over-complicates things. Hover is a bit overloaded, by both languageserver and sesson watcher here.

I wouldn't mind implementing this, since we already have the code for both options

Currently, when I click any R function that matches a help page, this light bulb pops up:
image
suggesting that something's wrong with my code that can be fixed by opening the documentation for `base::print`, which might be a bit irritating for some users (including myself).

I think having the help links as a code action is a nice feature, especially if we add more more code actions that trigger the bulb anyways, but I'd still include the option to move them to the hover (or deactivate them completely).

renkun-ken:
OK. The tooltip "Show fixes" does not sound consistent with what it is supposed to do. For example, a typical code action provider usually supports code refactoring such as move function to a new file, extract variable, etc, these are obviously not "fixes", but improvements or alternative actions one could do with the code.

Anyway, let me revert the hover part and let `r.helpPanel.showHelpLinks` take value from `"codeActions"|"hover"|"none"`.

renkun-ken:
Just tried working with code actions for a while and I find it quite annoying to see the lightbulb with cursor being put almost anywhere in the code. To my understanding, code actions should not appear too often like this, or it would be too distracting.

Personally, I think I won't merge this PR for this reason. Is that okay I close this PR?

ManuelHentschel:
> Just tried working with code actions for a while and I find it quite annoying to see the lightbulb with cursor being put almost anywhere in the code.

That bothered me as well, so I'm fine with closing this

Issue #580, opened 2021-03-15
Using object.size() in init.R could be expensive


renkun-ken:
**Describe the bug**

87f9bb6f8b450d897fb191780b086de07805fd4e in #544  uses `object.size(obj)` in `inspect_env` in `init.R` to support showing object size in the workspace viewer. However, the performance of this function could be extremely poor if the object contains character vectors in it.

```r
library(data.table)
dt <- data.table(id = 1:5000000)
for (i in 1:20) {
  dt[, paste0("x", i) := rep("hello", .N)]
}
system.time(object.size(dt))
```

```
   user  system elapsed
  1.068   0.687   1.741
```

This means, with session watcher enabled, whenever this data table is present in the global environment, then user has to wait for 1.7s after evaluating each top-level expression.

Therefore, I don't think we should call `object.size()` so eagerly in this way.

renkun-ken:
@ ElianHugh Do you think we should remove this call or at least be more conservative using this?

renkun-ken:
@Ikuyadeu This issue is quite blocking for users who work with large data with a few character vectors as the R session will become too laggy to use. I suggest that we make a quick release when #581 is merged.

ElianHugh:
Apologies, that's definitely not intended! I agree, we should remove the call.

ElianHugh:
I'm not sure of a way around this in base-R given this [stackoverflow post](https://stackoverflow.com/questions/19548938/faster-alternative-to-object-size).

Comparing alternatives in RStudio:

`system.time(object.size(dt))`
   user  system elapsed 
   2.06    0.75    3.14

`system.time(lobstr::obj_size(dt))`
   user  system elapsed 
   6.89    0.02    7.69

Or we could estimate it by getting the size of a subset of a given object?

renkun-ken:
>Or we could estimate it by getting the size of a subset of a given object?

It might add too much complexity if we take into account nested list, environment, object attributes, etc. where character vector could appear anywhere.

ElianHugh:
Agreed. I was looking through RStudio and it looks like they use object.size for their environment pane. I suppose using object.size would require a more substantial rewrite than is worth at the moment

tdeenes:
Maybe one can record the timings on first invocation, and turn off the auto-object.size-calculation if the timing result is above a given threshold. Most people do not work with huge character vectors in interactive sessions.

renkun-ken:
[...]

PR #581, opened 2021-03-15
Being more conservative to call object.size() in task callback

renkun-ken:

This PR is a temporary fix for #580 by being more conservative to call `object.size()` in the task callback so that it will only be called once when a symbol in the global environment has changed its memory address.

I do not use existing functions in packages to get memory address such as `lobstr::obj_addr()` or `data.table::address()` because I don't want session watcher to depend on either at the moment. The current CRAN release of `{lobstr}` still depends on Rcpp while the latest development only depends on `crayon` and `rlang` (which we already depend on) so that we could use lobstr in the future.

1. Start a new session with session watcher enabled.
2. Run the following code:

```r
library(data.table)
dt <- data.table(id = 1:5000000)
for (i in 1:20) {
  dt[, paste0("x", i) := rep("hello", .N)]
}
system.time(object.size(dt))
1+1
dt[, x21 := 0]
dt[, x21 := NULL]
rm(dt)
```

The time is significant but subsequent evaluation of expressions (e.g. `1+1`) does not trigger a long wait. ~~However, the object size in the workspace viewer is no longer available.~~

The object size that appears in the tooltip is the initial object size before any in-place modification is done, i.e. the object size is only updated when the address of the data pointed to by the symbol is changed.


src/workspaceViewer.ts
```diff
@@ -110,9 +110,16 @@ export class WorkspaceItem extends TreeItem {
 			return str;
 		}
 	}
+
+	private getSizeString(bytes: number): string {
```
ManuelHentschel, line 114: There's also `format(object.size(x), units = 'auto')` in R, which might look a bit nicer

R/init.R
```diff
@@ -113,14 +121,24 @@ if (interactive() &&
             )
           } else {
             obj <- env[[name]]
-            str <- capture_str(obj)[[1L]]
+
+            addr <- address(obj)
+            cobj <- cache[[name]]
+            if (is.null(cobj) || cobj$address != addr) {
+              cache[[name]] <- cobj <- list(
```
ManuelHentschel, line 128: Would using addresses as keys in the cache work here?
Variables might share the same address, but I'd expect them to be identical then (i.e. have the same size as well).

ManuelHentschel (commented):
Should we make this configurable with an option? To me personally, the size of a variable in memory is usually not important, but I would mind the delay occurring regularly

ManuelHentschel:
> The time is significant but subsequent evaluation of expressions (e.g. `1+1`) does not trigger a long wait. ~However, the object size in the workspace viewer is no longer available.~

Using `dt` from your example, subsequent calls to `dt[1,1]` have a significant delay in a vscode-R-session but are pretty much instantaneous in a normal R session.

src/workspaceViewer.ts
renkun-ken: Thanks for reminding! I thought it might be useful to keep passing `number` instead of `string` to keep the interface clean.

R/init.R
renkun-ken: Yes, it should work. I'm curious about if it is possible that the same address will be reused?

renkun-ken:
> Should we make this configurable with an option? To me personally, the size of a variable in memory is usually not important, but I would mind the delay occurring regularly

Makes good sense. This is my case too.

renkun-ken:
> > The time is significant but subsequent evaluation of expressions (e.g. `1+1`) does not trigger a long wait. ~However, the object size in the workspace viewer is no longer available.~
> 
> Using `dt` from your example, subsequent calls to `dt[1,1]` have a significant delay in a vscode-R-session but are pretty much instantaneous in a normal R session.

I cannot reproduce this. Currently, the first time a slow object appears in the globalenv, there will be a significant delay, but subsequent expressions should not if the address of the slow object is not changed.

R/init.R
renkun-ken:
```
> rm(x); gc(); x <- rnorm(1000); .Internal(inspect(x));                         
         used (Mb) gc trigger (Mb) max used (Mb)
Ncells 278099 14.9     655202   35   462665 24.8
Vcells 556656  4.3    8388608   64  1779488 13.6
@1f83740 14 REALSXP g0c7 [REF(1)] (len=1000, tl=0) -0.491836,-0.481699,-0.205983,-0.750808,1.92488,...

> rm(x); gc(); x <- rnorm(1001); .Internal(inspect(x));                         
         used (Mb) gc trigger (Mb) max used (Mb)
Ncells 278099 14.9     655202   35   462665 24.8
Vcells 556656  4.3    8388608   64  1779488 13.6
@1f83740 14 REALSXP g0c7 [REF(1)] (len=1001, tl=0) -1.16595,0.958288,1.88609,0.510728,0.0115951,...

```

Looks like the memory address could be reused.

src/workspaceViewer.ts
ManuelHentschel: Agree that having the number availabl in vscode can be useful (e.g. to sort by size), but I think having both `size` and `sizeString` would be worth the UI improvement. We could also try to replicate the `format(...)` behaviour in typescript, but I think using `format()` is somewhat cleaner

renkun-ken: Is there a notable difference between the result string from `format()` and `getSizeString()`? Looks like they are almost the same?

ManuelHentschel: `x <- 1` gives `56 bytes` vs `56.00  B` (with a double space).
`v <- as.list(1:10000)` gives `625 Kb` vs `625.05 KB`
In both cases I personally prefer the output of `format()`

R/init.R
ManuelHentschel: (How) does the function recognize that `x` has a different size in the second callback? (with variable name and address being equal)

ManuelHentschel:
> I cannot reproduce this. Currently, the first time a slow object appears in the globalenv, there will be a significant delay, but subsequent expressions should not if the address of the slow object is not changed.

Cannot reproduce either at the moment, no idea what happended there... Probably something in my .Rprofile...

R/init.R
renkun-ken: I update the mechanism so that when the symbol is not in cache or address changes or length changes, the object size will be re-calculated.

src/workspaceViewer.ts
renkun-ken: I see. I update the format of `getSizeString()`, and it returns exactly the same format now.

ManuelHentschel (approved):
LGTM, thanks for the nice work!

Issue #577, opened 2021-03-09
Workspace viewer str() functionality


danielbasso:
Following up on the discussion that took place in #416, I would like to suggest two improvements upon the current workspace viewer (which is already nice) that would make my experience using this extension almost perfect:

1) The most important one: the environment pane from RStudio has this button near each object that when pressed, print the result of `str(object)` under it:

![image](https://user-images.githubusercontent.com/20959177/110412801-89ff5480-806b-11eb-86d9-c4d13861a98a.png)

 I use this feature *all the time* as I always dealing with large dataframes with lots of columns, and sometimes you have to double check the actual type of one column in particular (like if a column is a `Date` a `character`). Sometimes I can't remember the name of some variable (not even how it starts) and this helps me. 

A similar button could be implemented in vscode-R:

![image](https://user-images.githubusercontent.com/20959177/110413214-3e00df80-806c-11eb-9de9-793fb6a858a7.png)

2) Apart from that, I think that the workspace panel could be moved to the explorer tab, under the folder contents panel. This, alongside the first suggestion, would make really easy to explore objects. The whole R tab (workspace + help) could be moved there, so you could manage your files and R objects without the need to switch between them:

![image](https://user-images.githubusercontent.com/20959177/110413959-abf9d680-806d-11eb-97a6-b2f89bb05108.png)

This happens with extensions from other languages (eg. Spring Boot Dashboard, Sonar Lint); having to switch tabs to access the workspace panel kinda hinders its potential to me.

ElianHugh:
> Following up on the discussion that took place in #416, I would like to suggest two improvements upon the current workspace viewer (which is already nice) that would make my experience using this extension almost perfect:
> 
> 1) The most important one: the environment pane from RStudio has this button near each object that when pressed, print the result of `str(object)` under it:
> 
> ![image](https://user-images.githubusercontent.com/20959177/110412801-89ff5480-806b-11eb-86d9-c4d13861a98a.png)
> 
>  I use this feature *all the time* as I always dealing with large dataframes with lots of columns, and sometimes you have to double check the actual type of one column in particular (like if a column is a `Date` a `character`). Sometimes I can't remember the name of some variable (not even how it starts) and this helps me. 
> 
> A similar button could be implemented in vscode-R:
> 
> ![image](https://user-images.githubusercontent.com/20959177/110413214-3e00df80-806c-11eb-9de9-793fb6a858a7.png)
> 

[...]

PR #583, opened 2021-03-18
Initial Workspace Viewer str() functionality

danielbasso:

As stated in #577, the current workspace viewer displays a list of globalEnv objects without further detail. For lists and dataframes, would be nice to see a sub-list of its elements, like the function `str()` output. I've implemented this functionality taking advantage of the already implemented `.Rprofile` option for `src()`-like output on hover, `vsc.str.max.level = 0 | 1 | 2`

**Screenshot**

Take for example this code:

```
library(tidyverse)

my_list  <-  list(a = function(x) x, y = " small string", z = c(1, 2, 3))

my_tibble  <- tibble(x = c(1,2,4,5,5,5,5,5,5,5), 
                  y = c(2,2,2,2,2,2,2,7,8,9), 
                  z = rep("aaa\n", 10),
                  shrnm = rnorm(10),
                  really_long_variable_name = rnorm(10))                  

my_number  <-  123

my_string  <-  "lala \n lala"
```

**`vsc.str.max.level = 2`**

image

**`vsc.str.max.level = 1`**

image

**`vsc.str.max.level = 0` (as before)**

image

As far as I tested, it is robust to line breaks (`\n`); it won't show any tooltip or View/Remove buttons for the subitens (harder to implement that I expected); It will only show expandable icons (>) for objects of list type:
image

Known issue: as it uses the same option for `str()` display on object hover, you can't set one option apart from the other. I decided to let it this way mainly because two reasons:

1) The option name `vsc.str.max.level` is generic enough to allow this and;

2) It would have required extra work to implement another option just for this, risking breaking something that is already working. I can't think of any concrete example of someone that would lika an complete `str()` on hover and none in the workspace, or vice-versa. Seemed to me an extra work for little return.

### DIFF
--- a/package.json
+++ b/package.json
@@ -747,12 +747,12 @@
         {
           "command": "r.workspaceViewer.view",
           "group": "inline",
-          "when": "view == workspaceViewer"
+          "when": "view == workspaceViewer && viewItem == rootNode"
         },
         {
           "command": "r.workspaceViewer.remove",
           "group": "inline",
-          "when": "view == workspaceViewer"
+          "when": "view == workspaceViewer && viewItem == rootNode"
         },
         {
           "command": "r.helpPanel.searchPackage",
--- a/src/workspaceViewer.ts
+++ b/src/workspaceViewer.ts
@@ -29,8 +29,25 @@ export class WorkspaceDataProvider implements TreeDataProvider<WorkspaceItem> {
 		return element;
 	}
 
-	getChildren(): WorkspaceItem[] {
-		return this.getWorkspaceItems(this.data);
+	getChildren(element?: WorkspaceItem): WorkspaceItem[] {
+		if (element) {
+			return element.str
+				.split('\n')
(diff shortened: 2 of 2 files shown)


renkun-ken (approved):
LGTM

Tested on macOS and Ubuntu and it works nicely.

danielbasso:
Not related at all with the PR, but i don't know where to ask: could you guys enable the 'Discussion Tab' for this project? I have some questions about the project inner workings that I would like to clarify, but the only way to communicate right now is through issues, which IMO is not ideal since issues should be specific for bugs or feature requests.

renkun-ken:
I agree that we could enable the discussion. We need @Ikuyadeu to proceed in the project settings.

Ikuyadeu:
@danielbasso  @renkun-ken OK, now I opened Discussion.
https://github.com/Ikuyadeu/vscode-R/discussions

PR #585, opened 2021-03-21
Thread execute argument through to term.sendText()

### LINKED ISSUE #575: Support sending to console without execution
fledge calls `rstudioapi::sendToConsole(execute = FALSE)`, this is currently unsupported. Also, we could provide a command that sends to the console without execution, this can be useful when we want to edit the code before executing.

I'd like the `execute = FALSE` variant supported, I think we need to call `term.sendText(code, false)` in this situation.

I have worked around in fledge for now, would be great to support this out of the box.

**Additional context**
Comment in original PR: https://github.com/Ikuyadeu/vscode-R/pull/535#issuecomment-792508390

VS Code API link: https://code.visualstudio.com/api/references/vscode-api#Terminal

krlmlr:
Closes #575.

I have no Idea if this code works, can't build: seeing `Error: Cannot find module 'copy-webpack-plugin'` when pressing F5 or Ctrl + Shift + B in VS Code.

There must be a better way to detect the last item in an iteration.

This code now also avoids the delay before the first line in non-bracketed paste.


andycraig:
Thanks for the PR! Can you try running this at the command line before pressing <kbd>F5</kbd>? That will hopefully fix the error.

```
# In directory vscode-R
npm install
```

krlmlr:
Thanks. I'm seeing:

```
npm WARN EBADENGINE Unsupported engine { package: 'devreplay@1.9.26',
npm WARN EBADENGINE   required: { node: '>=12.0.0' },
npm WARN EBADENGINE   current: { node: 'v10.19.0', npm: '7.6.0' } }
```

and then a ton of error messages. Running Ubuntu Focal.

krlmlr:
After `npm install -g node-gyp` as root, this worked -- the warning remains.

krlmlr:
Works for me now.

Unrelated: I think I have made a mess on my system with various runs of `apt install node-...` and `sudo npm install -g` calls. Ideally I'd like to have a clean per-project setup, without systemwide installations. What's a good reference to get started?

krlmlr:
I forgot to mention: the console message changed, "inserting" is shown with `execute = FALSE`, "sending" is now the default.

krlmlr:
Done, tests succeed now. What's a good location for a test of this functionality?

andycraig:
We don't have any unit tests for these functions that call `term.sendText()`, so I don't think it's necessary to add any tests as part of this PR.

Sounds like it's ready for review?

krlmlr:
Yes, the implementation works for me.

src/rTerminal.ts
```diff
@@ -237,7 +237,7 @@ export async function runChunksInTerm(chunks: vscode.Range[]): Promise<void> {
     }
 }
 
-export async function runTextInTerm(text: string): Promise<void> {
+export async function runTextInTerm(text: string, execute = true): Promise<void> {
```
andycraig, line 240: Suggest adding parameter type to function signature.

```suggestion
export async function runTextInTerm(text: string, execute: boolean = true): Promise<void> {
```

src/rTerminal.ts
```diff
@@ -247,12 +247,24 @@ export async function runTextInTerm(text: string): Promise<void> {
...
+        term.sendText(text, execute);
     } else {
         const rtermSendDelay: number = config().get('rtermSendDelay');
-        for (const line of text.split('\n')) {
-            await delay(rtermSendDelay); // Increase delay if RTerm can't handle speed.
-            term.sendText(line);
+        const split = text.split('\n');
+        const n_splits = split.length;
+        let count = 0;
+        for (const line of split) {
```
andycraig, line 256: I think a couple of lines could be saved here using `.entries()`:

```typescript
for (const [count, line] of split.entries()) {
```

andycraig (commented):
I've added a couple of minor notes, and everything else looks good.

I think you're using this with {fledge}, right? Could you write down the steps I can follow to cause it to send some text to the console without executing it? I'll follow those steps to test this PR. Thanks!

src/rTerminal.ts
krlmlr: Much nicer indeed.

krlmlr:
I test with:

```r
rstudioapi::sendToConsole("test()", execute = FALSE)
rstudioapi::sendToConsole("test(\n)", execute = FALSE)
```

I have tested the following combinations:

- :heavy_check_mark: radian with bracketed paste
- :heavy_check_mark: R with non-bracketed paste
- :crossed_swords: radian with non-bracketed paste: In the above example, the closing parenthesis is inserted automatically

andycraig (approved):
LGTM and works as described. Thank you for the PR!

PR #596, opened 2021-03-27
Clarify R path error messages

### LINKED ISSUE #568: Clarify 'can't use R' error messages
I was setting up VS Code on a new PC the other day and ran into the error '[Windows/Mac/Linux] can't use R'.

It's not very clear to the user what they need to do to fix it so I think the error message should include what setting they need to change (`rpath`).

The error message appears in `getRterm()` and `getRpath()`.

andycraig:
Closes #568 

Adds info to error messages when there are problems with the R path so that user knows what they need to do.

**Verify it doesn't break anything**
1. Run command `R: Create R terminal`
2. Observe that R terminal is created as expected

**Test error when specified R path doesn't exist**
1. Set `rterm` setting to a nonsense path (e.g., 'abc')
2. Run command `R: Create R terminal`
3. Observe error message that mentions `rterm`

These next two are slightly tricky to test because you have to intentionally create an error.

**Test `rpath` error**
1. Change line 79 of `util.ts` to `rpath = undefined`, so that an error will always occur
2. In R sidebar, use "Open Help Topic using '?'" and choose the top option
3. Observe error message that mentions `rpath`

**Test `rterm` error**
1. Change line 103 of `util.ts` to `rpath = ''`, so that error with always occur
2. Run command `R: Create R terminal`
3. Observe error message that mentions `rterm`

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -127,7 +127,7 @@ export async function createRTerm(preserveshow?: boolean): Promise<boolean> {
 
             return true;
         }
-        void vscode.window.showErrorMessage('Cannot find R client.  Please check R path in preferences and reload.');
+        void vscode.window.showErrorMessage(`Cannot find R client at ${termPath}. Please check r.rterm setting.`);
 
         return false;
     });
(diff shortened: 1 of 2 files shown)


Ikuyadeu (approved):
LGTM

andycraig:
@Ikuyadeu Thanks for the review!

PR #602, opened 2021-04-01
Refactor R code

ManuelHentschel:
This PR contains some restructuring of the R code used in the extension. Changes are:
- Move function definitions from `init.R` to their own file `vsc.R`. This hopefully makes it easier to contribute new functions etc. (in `vsc.R`) without having to worry about the technical issues solved in `init.R`.
- Attach `tools:vscode` after loading default packages. This makes it possible to mask e.g. `View` without having to mess with the environment of the `utils` package itself.
- Source R code directly from the installation directory of the extension. Since vscode is guaranteed to have access to its own files and the R process is launched by vscode, I don't think this should cause any read-access issues (please correct me if I'm wrong here)
- ~Move temporary files (`request.log`, `request.lock`) to a temporary directory instead of the home directory.~

~The last two points imply that the directory `~/.vscode-R` is not needed anymore.~

*~Todo:~*
- ~Check Rstudio api~
- ~Find solution for self-managed sessions (Maybe we should leave an optional "link file" in `~/.vscode-R`? Or expose the path to `init.R` in the extension files somehow and let the user copy it to their `.Rprofile`?)~

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -36,13 +36,14 @@ initFirst <- function() {
   }
 }
 
+old.First.sys <- .First.sys
 
 # Overwrite for `.First.sys`
 # Is used to make sure that all default packages are loaded first
 # Will be assigned to and called from the global environment,
 # Will be run with wd being the user's working directory (!)
 initLast <- function() {
-  base::.First.sys()
+  old.First.sys()
 
   # cleanup previous version
   removeTaskCallback("vscode-R")
@@ -103,7 +104,7 @@ initLast <- function() {
   # Attach to vscode
   exports$.vsc.attach()
 
-  invisible(NULL)
+  invisible()
 }
 
 initFirst()
--- a/R/init.R
+++ b/R/init.R
@@ -13,10 +13,10 @@ initFirst <- function() {
     !interactive()
     || Sys.getenv("RSTUDIO") != ""
     || Sys.getenv("TERM_PROGRAM") != "vscode"
-  ){
+  ) {
     return()
   }
-  
+
   # check requried packages
   required_packages <- c("jsonlite", "rlang")
   missing_packages <- required_packages[
@@ -66,7 +66,7 @@ initLast <- function() {
     .vsc.page_viewer <- .vsc$show_page_viewer
 
     # assign functions that are optional:
-    for(funcName in c("View", ".External.graphics")) {
+    for (funcName in c("View", ".External.graphics")) {
       if (funcName %in% ls(.vsc)) {
         assign(funcName, .vsc[[funcName]])
       }
@@ -77,7 +77,7 @@ initLast <- function() {
 
   # overwrite S3 bindings from other packages
   suppressWarnings({
-    if(!identical(getOption("vsc.helpPanel", "Two"), FALSE)) {
+    if (!identical(getOption("vsc.helpPanel", "Two"), FALSE)) {
       # Overwrite print function for results of `?`
       .vsc$.S3method(
         "print",
--- a/R/init.R
+++ b/R/init.R
@@ -1,13 +1,12 @@
-
 # This file is executed with its containing directory as wd
 
-# Remember absolute path of vsc.R
+# Remember the working directory (should be extension subfolder that contains this script)
 dir_init <- getwd()
 
 
 # This function is run at the beginning of R's startup sequence
-# Code that is meant to be run at the end of the startup should go in `initLast`
-initFirst <- function() {
+# Code that is meant to be run at the end of the startup should go in `init_last`
+init_first <- function() {
   # return early if not a vscode term session
   if (
     !interactive()
@@ -32,7 +31,7 @@ initFirst <- function() {
     )
   } else {
     # Initialize vsc utils after loading other default packages
-    assign(".First.sys", initLast, envir = globalenv())
+    assign(".First.sys", init_last, envir = globalenv())
   }
 }
 
@@ -42,7 +41,7 @@ old.First.sys <- .First.sys
 # Is used to make sure that all default packages are loaded first
 # Will be assigned to and called from the global environment,
 # Will be run with wd being the user's working directory (!)
-initLast <- function() {
+init_last <- function() {
   old.First.sys()
 
   # cleanup previous version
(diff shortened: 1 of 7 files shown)


src/session.ts
```diff
@@ -43,21 +43,13 @@ let activeBrowserExternalUri: Uri;
 export function deploySessionWatcher(extensionPath: string): void {
     console.info(`[deploySessionWatcher] extensionPath: ${extensionPath}`);
     resDir = path.join(extensionPath, 'dist', 'resources');
-    watcherDir = path.join(os.homedir(), '.vscode-R');
+    const tempPath = os.tmpdir();
+    watcherDir = path.join(tempPath, '.vscode-R');
```
renkun-ken, line 47: We put the directory in user home because I want to ensure that the directory and files in it are private to the user. Does putting in the temp directory also guarantee that?

Also, on a multi-user server, the path will conflict with other users if they also use the extension.

AFAIK, there's also practice putting some files in user home in other extensions for similar purposes.

I'm not seeing a clear advantage of putting files in system temp folder here.

src/session.ts
```diff
@@ -43,21 +43,13 @@ let activeBrowserExternalUri: Uri;
...
     console.info(`[deploySessionWatcher] watcherDir: ${watcherDir}`);
     if (!fs.existsSync(watcherDir)) {
         console.info('[deploySessionWatcher] watcherDir not exists, create directory');
         fs.mkdirSync(watcherDir);
     }
-    console.info('[deploySessionWatcher] Deploy init.R');
```
renkun-ken, line 52: We put these files in a easily known directory because there's an important use case to start self-managed R sessions (.e.g in tmux window) where user sources the `init.R` scripts on their own.

If we don't put them in such an directory, user will have know the extension path to source the script, which does not seem to be friendly enough.

src/session.ts
ManuelHentschel: On my windows machine, the temp dir is private for each user, so I didn't consider this limitation.
Considering this, I guess it makes sense to leave these files where they are.

renkun-ken: On a linux machine, the if we create a folder in temp, it does not seem to make it `700` private to user by default. I guess we have to put it in home folder to be secure for all platforms.

ManuelHentschel: Agreed

src/session.ts
ManuelHentschel: That's definitely an important point, but I personally find it a bit confusing to have the extension copy all .R files around before running them, just so they are easier to find.

Having them in the extension's directory would make it clear that they are not meant to be modified (currently, any modifications are overwritten anyways). A one-line R script in `~/.vscode-R` that points to the extension's installation directory would work as well, and avoids having two identical copies of the same code.

renkun-ken: > A one-line R script in ~/.vscode-R that points to the extension's installation directory would work as well, and avoids having two identical copies of the same code.

I like this idea. But in this script, it seems that it has to also set an environment variable of the init script folder path so that `init.R` script could know the paths of other scripts, which is needed by self-managed R sessions.

ManuelHentschel: I added a file that contains only `local(source("/PATH/TO/init.R", chdir = TRUE, local = TRUE))`, which temporarily changes the wd to the dir containing `init.R` to make sure that local path references work in `init.R`.

renkun-ken: Ok, that's a neat trick.

R/init.R
```diff
@@ -1,7 +1,23 @@
...
 
+# This file is executed with its containing directory as wd
+
+# Remember absolute path of vsc.R
+dir_init <- getwd()
+
+
+# This function is run at the beginning of R's startup sequence
+# Code that is meant to be run at the end of the startup should go in `initLast`
+initFirst <- function() {
```
renkun-ken, line 10: I guess we should stick to our existing snake_case naming convention in all newly introduced symbols?

ManuelHentschel: Thanks for fixing the script paths. Feel free to adjust naming/casing as you see fit. Considering the chaos of naming conventions in R itself, I have kind of given up on this topic.. :D

renkun-ken:
I checked rstudioapi and it works now.

renkun-ken (approved):
LGTM

R/init.R
```diff
@@ -14,706 +29,81 @@ if (interactive() &&
...
-              }
-            }
-          }, error = message)
-          TRUE
-        }
-
-        setHook("plot.new", new_plot, "replace")
-        setHook("grid.newpage", new_plot, "replace")
-
-        rebind(".External.graphics", function(...) {
```
renkun-ken, line 250: @ManuelHentschel   This change no longer forces the replacing of `base::.External.graphics` to a hacked version to trigger plot replay.

R/init.R
```diff
@@ -14,706 +29,81 @@ if (interactive() &&
...
+    .vsc.view <- .vsc$show_dataview
+    .vsc.browser <- .vsc$show_browser
+    .vsc.viewer <- .vsc$show_viewer
+    .vsc.page_viewer <- .vsc$show_page_viewer
 
-    .vsc.name <- "tools:vscode"
-    if (.vsc.name %in% search()) {
-      detach(.vsc.name, character.only = TRUE)
+    # assign functions that are optional:
+    for (funcName in c("View", ".External.graphics")) {
```
renkun-ken, line 69: This does not seem to work with `.External.graphics`.

R/init.R
ManuelHentschel: Should we just revert this to the old behavior or do you know a different fix? I tried to avoid messing with the `base` namespace itself, but if it's the only (feasible) way to achieve this, it of course makes sense to make an exception

renkun-ken: Yes, we rely on this replacement to make it possible for us to make all plot functions including e.g. `abline()` that adds graphics elements to the existing plot to trigger plot replay so that the plot is saved to the png file we watch in vscode.

renkun-ken: That's about the only way I know so that we don't have to write a full graphics device to know when the plot is updated. Therefore, I revert this via #625.

Issue #59, opened 2018-04-03
Provide R package-related tasks


jacob-long:
There are some things added in one of the recent releases that we should consider including as VS Code *tasks* as well.

In particular, being able to run tests and R CMD check is ideally-suited to tasks. 

I wasn't sure whether an extension can provide tasks in this way, but it appears to be possible. I don't know TypeScript well enough to implement, but here's an example of an extension that provides users with tasks: https://github.com/Microsoft/vscode-extension-samples/tree/master/task-provider-sample

I can help in terms of putting together the task definition JSONs since I'm using some in my own projects already.

Ikuyadeu:
@jacob-long Sorry my late reply.
Good idea! if you have a task JSON files, could you share it?

andycraig:
Hi @jacob-long, I’m following up on some old issues. Are tasks still something you’re keen to see added?

jacob-long:
I think these would be valuable additions to support R package developer workflows.

andycraig:
@jacob-long Great! It sounds like you might already have some task definition JSON files that you can share? If so, that would be very useful.

I structure most of my projects as packages and use a workflow based around `devtools::load_all()` and `devtools::test()`, but I’ve never submitted a package to CRAN, used R CMD, or used VSCode tasks. Would you be able to provide a list of desired tasks and what they should do? I’m happy to help out on the implementation side of things. 

Thank you!

crsh:
Sorry, to barge in. I'm only just getting started using VS Code, but in case it helps, my set up looks as follows:

~~~json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Build package",
            "type": "shell",
            "command": "RScript -e 'devtools::install(upgrade = \"never\")'",
            "presentation": {
                "reveal": "always"
            },
            "problemMatcher": [],
            "group": {
                "kind": "build",
                "isDefault": true
            }
        },
        {
            "label": "Run unit tests",
            "type": "shell",
            "command": "RScript -e 'devtools::test()'",
            "presentation": {
                "reveal": "always"
            },
            "problemMatcher": [],
            "group": {
                "kind": "test",
                "isDefault": true
            }
        },
        {
            "label": "Run R CMD checks",
            "type": "shell",
            "command": "RScript -e 'devtools::check()'",
            "presentation": {
                "reveal": "always"
            },
            "problemMatcher": [],
            "group": {
                "kind": "test",
                "isDefault": false
            }
        }
    ]
[...]

PR #603, opened 2021-04-03
Add tasks Check, Document, Install, Test

andycraig:
Relevant to #59 

Adds some Tasks (Document, Install, Test).

I'm hoping that having these in the code will encourage others to add tasks they think would be useful, since it should be pretty clear how these can be copy-pasted to add new tasks.

Run command `Tasks: Run Task`. Select `R`. Observe `R: Document`, `R: Install`, `R: Check`. Select one of them. Select `Continue without scanning the task output`. Observe that the Rscript command is sent to a new terminal. E.g., for `R: Install`, the following appears in the terminal:

`Executing task: Rscript -e "devtools::install()"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -997,7 +997,12 @@
           "description": "Remove hidden items when clearing workspace."
         }
       }
-    }
+    },
+    "taskDefinitions": [
+      {
+        "type": "R"
+      }
+    ]
   },
   "scripts": {
     "vscode:prepublish": "webpack --mode production",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -139,6 +139,25 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
     vscode.languages.registerCompletionItemProvider('r', new completions.StaticCompletionItemProvider(), '@');
 
 
+    // register task provider
+    const type = 'R';
+    vscode.tasks.registerTaskProvider(type, {
+        provideTasks() {
+            return [
+                new vscode.Task({type: type}, vscode.TaskScope.Workspace, 'Document', 'R',
+                    new vscode.ShellExecution('Rscript -e "devtools::document()"')),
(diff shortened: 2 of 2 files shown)


renkun-ken:
Looks like `R: Check` is missing. Would you like to also add it?

renkun-ken:
When I choose to `Tasks: Configure Tasks`, the following appears in the `task.json`:

```json
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "R",
			"problemMatcher": [],
			"label": "R: Document"
		}
	]
}
```

I don't see a way to customize the command or arguments like configuring other self-created tasks.

andycraig:
> Looks like `R: Check` is missing. Would you like to also add it?

Done.

andycraig:
Re: `Tasks: Configure Tasks`, I've never used that command so I'm not 100% sure what the correct behaviour would be here. I ran `Tasks: Configure Tasks` and selected `R: Install`, then added a `"presentation"` field as shown below. When I tried running `R: Install` again, the `Task - R: Install` terminal did indeed stay in the background as expected. It appears that the tasks can be customised in this way. So, perhaps this is working okay already?

```json
{
	"tasks": [
		{
			"type": "R",
			"problemMatcher": [],
			"label": "R: Install",
			"presentation": {
                                "reveal": "never"
                        }
		}
	]
}
```

renkun-ken:
Looks like the configuration is meant to customize other properties of the task except the command to run.

renkun-ken (approved):
LGTM

andycraig:
@renkun-ken Thanks for the review! I will merge.

ManuelHentschel:
Would there be problems with using `r.rpath.xxx` for these tasks? From my experience, the normal `R -e ...` (rather than `Rscript`) works as well and the user would be able to configure the R path used to run the task.

PR #608, opened 2021-04-06
Add knit task

krlmlr:
Expand list of tasks so that documents kan be rendered in a fresh session using tasks. Reference: https://github.com/Ikuyadeu/vscode-R/issues/59#issuecomment-813690674.

Open an `.Rmd` file, run the "R: Knit" task.

I wonder:

- How can we expand the task so that the rendered document is opened upon completion? Should we add an `rstudioapi::xxx()` call to the command line?
- Do we need "Knit as HTML", "Knit as PDF", "Knit all"?
- Should we eventually remap the Ctrl + Shift + K shortcut to this task
    - and enable it also for `.R` files
    - and also remap other shortcuts?

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -152,6 +152,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
                     new vscode.ShellExecution('Rscript -e "devtools::install()"')),
                 new vscode.Task({type: type}, vscode.TaskScope.Workspace, 'Test', 'R',
                     new vscode.ShellExecution('Rscript -e "devtools::test()"')),
+                new vscode.Task({type: type}, vscode.TaskScope.Workspace, 'Knit', 'R',
+                    new vscode.ShellExecution('Rscript -e "rmarkdown::render(\'${file}\')"')),
             ];
         },
         resolveTask(task: vscode.Task) {


malcook:
thought as an often emacs org-mode ESS (emacs speaks statistics) user and sometime rstudio user who would love to jump to VSC when possible:

"Knit to xxx" is good

"Knit All" is good and should knit to all formats in YAML header, of html if none.

Ctrl + Shift + K is what RStudio uses and is good unless otherwise claimed.

The YAML header of an rmarkdown document can specify one of more formats to which the document is expected to be knit, but a call to knit  can override.

jooyoungseo:
In RStudio IDE, the `knit` button (ctrl+shift+K) behaves differently depending on how you define knit  field within the YAML frontmatter of your document.

This has been described in [Customize the Knit button](https://bookdown.org/yihui/rmarkdown-cookbook/custom-knit.html).

For example, the following calls `bookdown::render_book()` instead of the default `rmarkdown::render()`:

`index.Rmd`:

``` yaml
---
title: Bookdown Rmd
knit: "bookdown::render_book"
output: bookdown::gitbook
---
```

It would be ideal if the knit task could pick up the `knit` metadata field within current document yaml frontmatter, and use the defined render function towards the opened file as in RStudio IDE.

andycraig:
@krlmlr Thanks for this PR! I'm happy to review it whenever you'd like. Adding knit HTML/PDF/all variations might be worthwhile now as they should be straightforward. You're welcome to take on other features like automatically opening the knitted document but it seems like it might be easiest to merge the basic features now and add on the more involved features in subsequent PRs.

gowerc:
@krlmlr,

Can I just double check what you were trying to achieve here. The current `Ctrl + Shift + K` shortcut already runs the Rmd in a fresh R session. I'm just wondering what the benefit of mapping it over to tasks instead would be.

To @jooyoungseo's point it also looks like the existing knit all / `Ctrl + Shift + K` also respects the knit frontmatter

image


Theres also quite a bit of chat here around different potential features we could implement is it worth extracting these to separate issues ? I would propose closing this PR considering its now outdated/incompatible with the current changes to `tasks.ts`.  I started putting together a new version that is compatible with the new changes but am unsure now of its value given what I've put above. @renkun-ken  any opinions ?

jooyoungseo:
@gowerc I believe the customizable knit (at least my suggestion) has already been addressed in the latest version, so this is somewhat outdated.

Issue #605, opened 2021-04-05
Launch Rstudio add-in gives "Error: RStudio not running"


albert-ying:
Hi, I did what is suggested in the document to activate Rstudio addin support.

After I call the command `R: launch Rstudio addin`, I can see a list of addins in the command palette.

<img width="858" alt="image" src="https://user-images.githubusercontent.com/59846322/113540421-1e4ad180-95ae-11eb-8dcd-d2bcb0d92437.png">

But after I select it, it gives me an error in the terminal

```
[ins] r$> xaringan:::inf_mr()                                                                             
Error: RStudio not running
```

I also tried other addins listed, but they give me the same error...

Did I miss anything here?

Thank you!

MilesMcBain:
This addin works for me.

The error suggests the `rstudioapi` functions have not been loaded by VSCode, however the fact that you are able to get to the addin menu suggests they have.

I can reproduce your behaviour by doing the following:

1. Start an R terminal with the VSCode extension using either command `R: Attach active terminal` or `R: Create new terminal`.
2. Create a new default terminal (running bash or powershell) and run the R executable with the command `R` or `R.exe`
3. With the terminal from step 2. active, run the addin from the menu and get your error.

The addin picker just passes a command to be run in the active terminal. If that active terminal is not one that has been initialised (attached) by the VSCode-R extension, RStudio addins, and many other VSCode-R extension features, will not work as expected.

Could running an R session not initialised by the vscode-R extension be a contributing factor in your case?

albert-ying:
Thank you @MilesMcBain 

I tried again, and it was not the scenario you described, as esquisser works

```
[ins] r$> .vsc.attach()                                                                                      

[ins] r$> xaringan:::inf_mr()                                                                                
Error: RStudio not running

[ins] r$> esquisse:::esquisser()                                                                             
Loading required package: shiny

Listening on http://127.0.0.1:5829
Browsing http://127.0.0.1:5829
```

Could it be something related to my .Rprofile or VS-code setting? Also, you mentioned this addin works for you, did you use the newest version? As I just installed it yesterday, it might be something about it?
```
remotes::install_github('yihui/xaringan')
```

Thank you so much!

MilesMcBain:
If xaringan were doing something weird in the latest version I would expect a different error about rstudioapi functionality not supported by vscode.

I'm thinking it could be your .Rprofile. I've seen situations where things in the profile cause the attach process to terminate abnormally.

[...]

PR #610, opened 2021-04-07
shim the rstudioapi if it has already been loaded

MilesMcBain:
fixes bug reported in #605

If the user has something in their .Rprofile that refers to the `rstudioapi` namespace, then that causes the namespace to be loaded before we can attach our onLoad hook in the extension R session initialisation routine.

RStudio Addins can be browsed via the picker, but not run in this case.

If you want to verify the bug you can add `rstudioapi::is_available()` to your alongside `vsc.rstudioapi = TRUE` in your `.Rprofile` and then try to run

`rstudioapi::insert_text("test")`

You should see: `Error: RStudio not running`

With this PR in place, try the same sequence and the text will be inserted into the current text editor window.

### DIFF
--- a/R/vsc.R
+++ b/R/vsc.R
@@ -560,6 +560,14 @@ if (rstudioapi_enabled()) {
       rstudioapi_util_env$rstudioapi_patch_hook(rstudioapi_env)
     }
   )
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

Works as expected. Thanks!

PR #614, opened 2021-04-08
Proof of concept: Integrate httpgd

ManuelHentschel:
*Update: To make styling easier, I've removed some hardcoded styles in a [fork of httpgd](https://github.com/ManuelHentschel/httpgd/tree/vsc).*
*The viewer now contains a very rudimentary plot history, which highlights a plot on hover ~but does not do anything else~ and lets the user select a plot*

**What problem did you solve?**
This PR contains a very basic integration of [httpgd](https://github.com/nx10/httpgd) to display plots.
A big advantage of the SVG format used here is that it can be styled using the CSS variables provided by vscode.
Between the two screenshots below, only the color theme and font family (in vscode's settings) were changed, the plot updated automatically.
The added typescript code is mostly just copied from [here](https://github.com/nx10/httpgd/blob/master/inst/www/httpgd.ts).

**Screenshots:**
httpgd1
httpgd2
image

**How can I check this pull request?**
Install httpgd from [this branch](https://github.com/ManuelHentschel/httpgd/tree/vsc), copy the following code to your `.Rprofile` and call `plotTest()` in an R terminal created by the extension.
``` r
options(vsc.plot = FALSE)
options(device = function(...) {
  httpgd::hgd(
    silent = TRUE
  )
  .vsc$request('httpgd', url=httpgd::hgd_url())
})

plotTest <- function(){
  plot(sin(1:30/10))
  plot(sin(1:90/10))
  plot(sin(1:300/10))
}
```

cc: @nx10, @renkun-ken

PR #620, opened 2021-04-14
Integrate httpgd

nx10:
***Update:*** To check the PR, you need to have `httpgd` installed and set `options(vsc.use_httpgd = TRUE)` in your .Rprofile.

---

This PR contains a very basic integration of `httpgd` to display plots.
It is intended to be a WIP PR and a continuation of (https://github.com/Ikuyadeu/vscode-R/pull/614).

I added the implementation of a `httpgd` server connection backend drafted by @ManuelHentschel and removed the plot styling to make it compatible with all `httpgd` versions (current CRAN and GitHub master).

@ManuelHentschel will now implement a custom plot viewer that integrates with and uses vscode controls.
The development is also discussed in https://github.com/nx10/httpgd/issues/63.

(This Post will be updated with screenshots and setup instructions as the development continues. Currently these are identical to https://github.com/Ikuyadeu/vscode-R/pull/614 with the exception that all `httpgd` versions can be used.) 

---
*Edit:* The viewer now uses the new api and includes some buttons in the titlebar to control it. In particular, it's possible to toggle between the original svg and the vscode-themed svg using the "Toggle Style" button.

image
image


renkun-ken:
I'm testing with the following code from <http://r-statistics.co/Top50-Ggplot2-Visualizations-MasterList-R-Code.html#Scatterplot>:

```r
# install.packages("ggplot2")
# load package and data
options(scipen=999)  # turn-off scientific notation like 1e+48
library(ggplot2)
theme_set(theme_bw())  # pre-set the bw theme.
data("midwest", package = "ggplot2")
# midwest <- read.csv("http://goo.gl/G1K41K")  # bkup data source

# Scatterplot
gg <- ggplot(midwest, aes(x=area, y=poptotal)) + 
  geom_point(aes(col=state, size=popdensity)) + 
  geom_smooth(method="loess", se=F) + 
  xlim(c(0, 0.1)) + 
  ylim(c(0, 500000)) + 
  labs(subtitle="Area Vs Population", 
       y="Population", 
       x="Area", 
       title="Scatterplot", 
       caption = "Source: midwest")

plot(gg)
```

In original webview:

image

In new httpgd viewer:

image

Looks like the webview does not fit well vertically.

ManuelHentschel:
Thanks for testing the PR! Under the plot there should be a bar that you can drag to resize the plot vertically.

ManuelHentschel:
The last commit improves scrolling by allowing to scroll below the bottom of the page (similar to the behaviour when editing text) and fixing the resizing mechanism when scrolled down.

@renkun-ken, @nx10 & others, have you found any (breaking) problems or are you fine with merging the viewer?

renkun-ken:
Thanks! I'll do more testing soon.

renkun-ken:
I don't see the difference between "Hide plot" vs. "[Alt] Close plot" on the same buttom. In both cases, it removes the plot. Please elaborate.

renkun-ken:
The "Open httpgd url" button does seem to work immediately but requires a URL? The httpgd device seems to be open silently, I'm not sure if it makes sense to ask user for the URL?

renkun-ken:
> I don't see the difference between "Hide plot" vs. "[Alt] Close plot" on the same buttom. In both cases, it removes the plot. Please elaborate.

OK, I get it. "Hide plot" could hide the current plot and "Restore plots" could restore all hidden plots.

renkun-ken:
The theming feature looks interesting, but I'm not sure if it is useful yet. Following are some examples:

GitHub Light:

image

image

GitHub Dark:

image

image

Looks like the plot will not be much useful whenever it uses multiple colors.

renkun-ken:
I notice that both Zoom In and Out buttons have Alt to function as its counterpart. Is it redundant?

renkun-ken:
A nice part of httpgd native viewer I find quite convenient is its support of keyboard shortcuts, e.g. arrow keys to navigate. Is it possible to support (maybe a subset of) these keyboard shortcuts in the viewer?

ManuelHentschel:
> I don't quite get the purpose of "Open httpgd url" button. What is it supposed to do?

That command allows you to open a viewer for an external httpgd instance. E.g. run this code in an external R session:
``` r
> httpgd::hgd()
httpgd server running at:
  http://127.0.0.1:59008/live?token=xRCZd0ow
> plot(123)
```
Then copy the shown URL and click "Open httpgd url" in vscode to view the plot inside vscode.

I'm not sure in what scenarios this functionality is useful, but it was too easy to implement to omit it.

---
> OK, I get it. "Hide plot" could hide the current plot and "Restore plots" could restore all hidden plots.

An important difference is that "Hide Plot" is front-end only, i.e. the plot can still be viewed in other viewers of the same httpgd instance. "Close Plot" actually sends a close command to the backend, closing the plot for other viewers as well.

> It would definitely look better if the icons are next to each other.

I agree, I think the title bar menu needs some general cleanup. For now I just wanted all possible commands in plain view so that users find them. As the list of commands grows, it might make sense to put some of them in submenus or only make them accessible through the command palette.

(I'm currently working on vscode v1.53 due to some unresolved bug in the latest version, so my menu is all messed up anyways, which is why I didn't put too much thought into this yet)

---

> The theming feature looks interesting, but I'm not sure if it is useful yet. Following are some examples:
> [...]

When your plots rely on coloring to convey information, I agree that the original theming is more useful. However, when you just want to take a look at a single data column, or work with plot styles that don't rely on coloring, the adapted theme conveys all relevant information and integrates much nicer with the rest of vscode.

E.g. when just checking the strucure of a graph (as in the screenshots in the very first comment), you don't really lose any information, but replacing the large white areas is much easier on the eyes, in my opinion.

---

> I notice that both Zoom In and Out buttons have Alt to function as its counterpart. Is it redundant?

That's indeed somewhat redundant. The idea behind this is e.g. when you are looking for the "zoom sweet spot" you might want to zoom in/out repeatedly and then it's easier (at least for me) to just hold and release shift than to move the mouse around.
If we implement a better use for these alt-commands at some point, I wouldn't mind replacing them, though.

---

> A nice part of httpgd native viewer I find quite convenient is its support of keyboard shortcuts, e.g. arrow keys to navigate. Is it possible to support (maybe a subset of) these keyboard shortcuts in the viewer?

Since the viewer is controlled (almost) only through vscode commands, these can be mapped using the normal keybindings.json. Adding e.g.
``` json
{
      "key": "k",
      "when": "resourceScheme =~ /webview/ && r.httpgd.active",
      "command": "r.httpgd.prevPlot"
},
{
      "key": "j",
      "when": "resourceScheme =~ /webview/ && r.httpgd.active",
      "command": "r.httpgd.nextPlot"
}
```
allows you to go through plots using `k` and `j`. In the future, we could add some of these keybindings by default or just make list of suggested/possible mappings.

renkun-ken:
I'm testing with a session started in a tmux window. When I reload vscode, and attach the same session, I don't see an obvious way to also recover the plot viewer.

I have to run `httpgd::hgd_url()` and run command "R Httpgd: Open httpgd url" and put this url in the input box and then the plot viewer is recovered. It would be certainly nice if we make a command to show the plot viewer for the attached session, or whenever the session is attached and the httpgd device is open, show the plot viewer automatically.

renkun-ken:
Another thing I notice is that when a number of plots are produced, and I click one thumbnail to navigate to different historic plots, then there's a significant delay to show the plot (nearly 1-2 seconds).

.vscode/launch.json
```diff
@@ -10,7 +10,8 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
-        "--extensionDevelopmentPath=${workspaceFolder}"
+        "--extensionDevelopmentPath=${workspaceFolder}",
+        "${workspaceFolder}/../testHttpgd"
```
renkun-ken, line 14: Should we remove this?

package.json
```diff
@@ -620,6 +620,100 @@
         "title": "Open help for path",
         "category": "R",
         "command": "r.helpPanel.openForPath"
+      },
+      {
+        "title": "Toggle Style",
+        "category": "R Httpgd",
```
renkun-ken, line 626: I guess "R Plot" might be a better name for the category? `httpgd` is the package name and we rely on the package as the backend to provide R plot in vscode. For end users, they might only need to think of "plot" instead of the specific package we rely on?

.vscode/launch.json
ManuelHentschel: Yes

package.json
ManuelHentschel: Makes sense. For now I wanted to make sure that users know what's part of the httpgd-based viewer and what's part of the "old" viewer, but if the old viewer does not have any settings/commands that might lead to confusion, we can also rename as you suggested.

ManuelHentschel:
> 
> 
> I'm testing with a session started in a tmux window. When I reload vscode, and attach the same session, I don't see an obvious way to also recover the plot viewer.
> 
> I have to run `httpgd::hgd_url()` and run command "R Httpgd: Open httpgd url" and put this url in the input box and then the plot viewer is recovered. It would be certainly nice if we make a command to show the plot viewer for the attached session, or whenever the session is attached and the httpgd device is open, show the plot viewer automatically.

I added a request to `.vsc.attach` that launches a httpgd viewer.

There is some potential to automate this in the future, using webview serialization, but that might take some more effort.

ManuelHentschel:
> Another thing I notice is that when a number of plots are produced, and I click one thumbnail to navigate to different historic plots, then there's a significant delay to show the plot (nearly 1-2 seconds).

Have you changed the settings `r.httpgd.timeouts.refreshTimeout` and/or `r.httpgd.timeouts.resizeTimeout` at some point? These can be used to control how eagerly the viewer redraws the plots and might help here. If changing these to a different value (e.g. both 100) doesn't help, please post the code that produces the laggy plots.

package.json
renkun-ken: The old plot does not have any setting in vscode but an option in R `vsc.plot` to control which view column to show the png plot file or `FALSE` to disable.

I guess we could rename vscode settings from `r.httpgd.*` to `r.plot.*` without producing any confusion?

renkun-ken:
> > Another thing I notice is that when a number of plots are produced, and I click one thumbnail to navigate to different historic plots, then there's a significant delay to show the plot (nearly 1-2 seconds).
> 
> Have you changed the settings `r.httpgd.timeouts.refreshTimeout` and/or `r.httpgd.timeouts.resizeTimeout` at some point? These can be used to control how eagerly the viewer redraws the plots and might help here. If changing these to a different value (e.g. both 100) doesn't help, please post the code that produces the laggy plots.

Cannot reproduce the delay on macOS. It is probably caused by the plot is a quite comprehensive with too many elements.

renkun-ken:
With the latest commit, the resizing seems a bit problematic that if you constantly run the same plot (e.g. `plot(rnorm(100))`), then the main plot area is getting smaller and smaller.

ManuelHentschel:
> 
> 
> With the latest commit, the resizing seems a bit problematic that if you repeat running the same plot (e.g. `plot(rnorm(100))`), then the main plot area is getting smaller and smaller.

Thanks for noting!

package.json
ManuelHentschel: Sounds good

package.json
```diff
@@ -995,6 +1149,42 @@
           "type": "boolean",
           "default": false,
           "description": "Remove hidden items when clearing workspace."
+        },
+        "r.plot.defaults.colorTheme": {
```
renkun-ken, line 1153: Do we really need that many levels of config key?

`r.plot.defaultColorTheme`, `r.plot.defaultPlotPreviewLayout`, `r.plot.resizeTimeout`, `r.plot.refreshTimeout` seem to look good enough.

ManuelHentschel: My motivation for nesting config entries this way is to control how they are sorted in the settings UI. Currently, this isn't a big deal, but if we simplify the config keys and later add e.g. `r.plot.resizeOnEventXYZ`, it will be displayed between `r.plot.resizeTimeout` and `r.plot.refreshTimeout`, though it would make more sense to group the timeouts together.

renkun-ken: OK. I also guessed it is intended for sorting, then it is useful indeed.

BTW, the timeouts seem more like intervals? I saw other extensions use "timeout" to limit the amount of time to wait until success (e.g. connection timeout). "Interval" seems more accurate for the purpose?

ManuelHentschel: Makes sense

renkun-ken: Still using `r.plot.timeouts.*`?

ManuelHentschel: Yeah I figured somewhere in the name it should indicate that it's a time-interval (since "interval" is not used uniquely for time delays). But we can also group it as `r.plot.timing.*` etc. if that sounds better, I don't really mind :)

renkun-ken: `r.plot.timing.*` sounds ok.

package-lock.json
```diff
@@ -1,4636 +1,8 @@
 {
   "name": "r",
   "version": "1.6.6",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
renkun-ken, line 4: The change in lock file seems to generate much bigger diff than it really has. Anything we could do about it? Somebody else runs `npm install` next time and the lock file could probably change back and forth.

ManuelHentschel: Not really sure what's going on here. I think I merged a different lockfile from master at some point or accidently changed this at some point. I can just revert this change for now.

Long term it might make sense to switch to yarn anyways.

renkun-ken (approved):
LGTM. Thanks for the great work!

renkun-ken:
@ManuelHentschel Would you like to update `README.md` (https://github.com/Ikuyadeu/vscode-R#available-functions-and-options) so that user knows about the `vsc.use_httpgd` option?

ManuelHentschel:
Thanks for the thorough testing and constructive feedback @renkun-ken!

nx10:
Great work @ManuelHentschel!

psobolewskiPhD:
The new feature works really well!
Kudos to all involved!
image

I have a couple questions:
1) Is it possible to have the vscode theme just do color and not override my font choice (here Fira Sans).
2) Is it possible to show/hide the history/thumbnails?
3) the ... menu allows for export, but just svg. Is there a way to to have PNG like the original httpgd functionality?

ManuelHentschel:
Thanks for the feedback!

A quick way to achieve 1. and 2. is to e.g. comment out the line https://github.com/Ikuyadeu/vscode-R/blob/master/html/httpgd/styleOverwrites.css#L9 and change https://github.com/Ikuyadeu/vscode-R/blob/master/html/httpgd/style.css#L53 to `display: none;`. You can try this out by opening the installation directory of the extension (`F1 > Extensions: Open Extensions Folder`) and modifying the .css files directly.

I'll also add proper commands/options for this.

Regarding 3., in https://github.com/nx10/httpgd/issues/69 a modular renderering approach is being implemented that will allow exporting to a number of different formats. When this is merged in httpgd, we'll also add corresponding options in vscode-R.

psobolewskiPhD:
Thank you for the clear instructions!
Two more questions:
I'm not in love with the white outline text vs. just white.
image
I thought I'd comment out the stroke and add fill: white; but that didn't work.
Is that an easy fix?

And more importantly—to me anyways—I **hate** the boxes around the legend. How can I turn that off?!?

psobolewskiPhD:
I played around a bit more and got—to my eye—very nice results:
image

Editted styleOverwrites.css as follows:
image

nx10:
The dynamic styling is basically just a proof of concept at this point. I plan on adding a second SVG renderer to httpgd that allows for mich easier styling at the expense of slightly increased file sizes (https://github.com/nx10/httpgd/issues/65). This will be included in the next update with the modular rendering.

(It might take a couple weeks/months as I am currently busy writing my masters thesis.)

psobolewskiPhD:
Honestly, it works great and—with the tweaks I just made—for me it looks great!
I really like the ability to easily flip plots from light vs dark mode for display purposes—saving the plots uses the normal (dark on white) colors, which is perfect.
Wonderful feature.
Kudos to you @nx10 and @ManuelHentschel plus everyone else involved in this extension.

nx10:
Thanks! @ManuelHentschel did most of the work on the vscode-r integration and the theming is completely his idea.

PR #626, opened 2021-04-20
LiveShare Functionality

### LINKED ISSUE #555: Implement shared R workspace in VS Live Share

ElianHugh:
Closes #555 

# General functionality
Vscode-R functions related to the session watcher currently do not function in LiveShare sessions. This means that the workspace viewer, dataview, plots, etc., do not work for guests.

This PR implements a 'LiveSession Listener', which allows for the sharing of file content and the forwarding of commands from a guest session to the host terminal. Thus, dataview, plots, browsers, and the workspace viewer should now function for guest sessions.

image

# Settings & controls

This PR also contributes a liveshare control view, for tweaking guest access to the R workspace. This involves the following controls:
- command forwarding
    - Whether the host allows guests to forward commands to the host terminal. E.g. workspace viewer view, remove
- workspace sharing
     - Allows guests access to the R environment and related files
- share ports
     - Automatically share & close browser ports with the guest

image

Default state can be controlled through three contributed settings.

## Features
- [x] Attach terminal
- [x] Send commands to terminal 
- [x] Plots (images)
- [x] Dataview: file
- [x] Dataview: table
- [x] Dataview: list
- [x] Browser
- [x] Webview: I didn't change much but it seems to work regardless? Might need someone else to check this one
- [x] Help Pages
- [x] Control menu/view

**I've intentionally left addin functionality out of this PR, as it is either a bigger undertaking or does not function with guest sessions at all. When initially adding guest access to addins, there was significant desyncing between host and guest sessions, so I have decided to leave it out. Further complicating the issue, there is no way (at the moment) of distinguishing between a guest addin call and a host addin call.

# Extra things to note
* Adds vsls as a dependency
* The window has to be reloaded if liveshare is not enabled prior to starting an R session
* Both the host and the guest have to be running vscode-R with this PR
* the guest does *not* need R installed

# How can I check this pull request?

1. Have the LiveShare extension installed
2. Have session watcher enabled
3. Start an R LiveShare session, session watcher capabilities should be functional** for both host and guest


ElianHugh:
As an aside, I've tried to write this with minimal impact on vscode-R development in mind. For example, the readContent function in util.ts is a drop-in replacement for readFile, which delegates to the guest service when necessary.

Ideally, contributors shouldn't have to worry about making their feature work on LiveShare as well. If there's anything I can do on this PR to help with this, please let me know!

ElianHugh:
> @ElianHugh, are there additional steps required to try this PR beyond those listed [here](https://github.com/Ikuyadeu/vscode-R/wiki/Contributing)? I have not tried out development VS Code extensions before. I installed npm and Node.js with Homebrew, and can confirm that I have the R and Live Share extensions installed, and session watcher enabled. But when I press F5 to load the extension cloned from your [repo](https://github.com/ElianHugh/vscode-R), then open a folder with R files, I get the following error:
> 
> > Activating extension 'Ikuyadeu.r' failed: Cannot find module '~/vscode-R/dist/extension'
> Require stack: - /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/loader.js - /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-amd.js - /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-fork.js.

Thanks for checking out the PR! The install instructions should be all that you need. Just to double check, did you run npm install in the extension folder, prior to pressing F5?

Also, depending on your local set-up, building the extension can hang. E.g. sometimes the npm terminal will ask if you want to use webpack, but not actually focus the terminal

package.json
```diff
@@ -100,6 +108,10 @@
       {
         "view": "rHelpPages",
         "contents": "R Help Pages"
+      },
+      {
+        "view": "rLiveShare",
+        "contents": "R LiveShare not active."
```
gvelasq, line 114: Microsoft refers to [Live Share](https://visualstudio.microsoft.com/services/live-share/) with spaces as opposed to LiveShare. Should this be 'R Live Share not active.'?

gvelasq:
> Thanks for checking out the PR! The install instructions should be all that you need. Just to double check, did you run npm install in the extension folder, prior to pressing F5?
> 
> Also, depending on your local set-up, building the extension can hang. E.g. sometimes the npm terminal will ask if you want to use webpack, but not actually focus the terminal

Thanks for the tip. You were correct, I had missed the `npm install` step. I was able to get it working, and it looks great! A few notes:

1. Based on the description of the PR above I was not sure I fully understood how the 'Forward guest commands' menu item was supposed to work. I expected that enabling it as the host, without toggling anything in the Live Share extension, would allow any kind of guest (connecting from a desktop VS Code instance or a browser instance) to type in the R terminal and share commands (e.g., write to the shared R workspace). Instead, the shared R terminal behaved as other shared terminals do on Live Share even when 'Forward guest commands' was enabled for a guest connecting in the browser. When a guest connecting from a browser tries to write in the R terminal with this menu item enabled, they see the following:
> Cannot edit a read-only terminal.
Source: Live Share (Extension)

The host needed to explicitly enable read/write permissions on the shared R terminal by going to Live Share > Session Details > Shared Terminals > R Interactive (Read-only) > Right-click > Make Read/Write.

I'm not sure that I know the best answer, but perhaps there can be an additional menu item in the R extension 'Live Share Controls' menu that detects and displays the status of the shared R terminal in the Live Share extension (not shared, read-only, read-write), and which ideally could be used to toggle sharing permissions for the R terminal directly from the R extension without needing to navigate to the Live Share extension? Or, perhaps this is just an issue for guests connecting from a browser and not for guests connecting from a desktop VS Code application with the R and Live Share extensions installed?

2. This may be outside the scope of this PR, but the VS Code R extension cannot be installed when a guest connects via a browser, but the host is still able to share a Read/Write R terminal with guests connecting from a browser. The guest on a browser is able to interact with objects in the shared R workspace from the terminal, but they can't see the workspace itself without the R extension installed. Currently the error when trying to install the R extension from a guest browser is this:
>The 'R' extension is not available in VS Code Web. Click 'More Information' to learn more.

Is it possible to make the R extension available to guests connecting from a browser? Guest browsers are a special use case for Live Share so perhaps this last point could become a separate issue so this PR could be merged first.

Many thanks again for the effort, this is a great contribution!

ElianHugh:
> 1. Based on the description of the PR above I was not sure I fully understood how the 'Forward guest commands' menu item was supposed to work. I expected that enabling it as the host, without toggling anything in the Live Share extension, would allow any kind of guest (connecting from a desktop VS Code instance or a browser instance) to type in the R terminal and share commands (e.g., write to the shared R workspace). Instead, the shared R terminal behaved as other shared terminals do on Live Share even when 'Forward guest commands' was enabled for a guest connecting in the browser.

The intent of the "forward guest commands" was to fix the issue of non-invasive vscode-R commands from failing due to a terminal being read only. E.g. viewing a workspace viewer item would fail (something fairly innocuous). Toggling this allows the host to prevent or allow ALL users from using vscode-R commands (such as the workspace viewer commands). This is mainly handled this way due to the liveshare permissions API still being in development.

Having said that, I think what the "forward guest commands" button does could definitely be better communicated.

> 
> Is it possible to make the R extension available to guests connecting from a browser? Guest browsers are a special use case for Live Share so perhaps this last point could become a separate issue so this PR could be merged first.

Unfortunately I think you're on the money in terms of the difficulty in getting browsers to work with the R workspace. 

One issue is that there are still certain functions that require a file to parse for them to work. For instance, the (default) plot viewer requires an image file. The PR gets around this by getting the file content from the host, and then creating the file on the guest machine. The issue, then, is handling how the browser handles (and has access to) file systems vs. how an application handles file systems. It may be possible to handle this disparity, but I suspect this may prove to be difficult.

The other issue was to do with sharing terminals in browsers - last time I checked this was still in development, which meant I couldn't do much until it was fininalsed.

Thank you very much for the feedback!

src/rShareTree.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+class CommandNode extends ToggleNode {
+    toggle(treeProvider: LiveShareTreeProvider): void {
+        forwardCommands = !forwardCommands;
+        this.description = forwardCommands === true ? 'Enabled' : 'Disabled';
+        treeProvider.refresh();
+    }
+
+    public label: string = 'Forward guest commands';
```
gvelasq, line 126: Instead of `Forward guest commands`, consider making it clear that this is for non-invasive vscode-R commands instead of the R terminal. Possible alternatives are: `Allow guests to interact with R Workspace` or `Allow guests to interact with host R extension`.

src/rShareTree.ts
```diff
@@ -0,0 +1,153 @@
...
+    public tooltip: string = 'Whether R commands should be forwarded from the guest to the host terminal (bypasses terminal permissions)';
```
gvelasq, line 127: As above, to make it clear the commands are non-invasive vscode-R commands (instead of the R terminal permissions managed in the Live Share extension), a potential alternative is:

`Whether commands to interact with the R extension should be forwarded from the guest to the host (bypasses permissions); shared R terminal (command line) permissions can be toggled in the Live Share extension`

gvelasq:
> As an aside, I've tried to write this with minimal impact on vscode-R development in mind. For example, the readContent function in util.ts is a drop-in replacement for readFile, which delegates to the guest service when necessary.
> 
> Ideally, contributors shouldn't have to worry about making their feature work on LiveShare as well. If there's anything I can do on this PR to help with this, please let me know!

@renkun-ken, could you please review? It would be great to see this merged.

renkun-ken:
Thanks! I'll do more testing on this soon.

renkun-ken:
I didn't have experience using LiveShare before. I just play with this PR a bit, and it is astonishing. Everything seems to already work quite smoothly.

I'll take a closer look at the code and use it more thoroughly and see if I can find something to improve.

Thanks for the great work!

src/rShare.ts
```diff
@@ -0,0 +1,322 @@
...
+
+// random number to fake a UUID for differentiating between
+// host calls and guest calls (specifically for the workspace
+// viewer 'View' function)
+export const UUID = Math.floor(Math.random() * Date.now());
+
+/// state-tracking bools
+// Bool to check if live share is loaded and active
+export async function isLiveShare(): Promise<boolean> {
+    const shareStarted = (await vsls.getApi())?.session.id;
```
renkun-ken, line 31: When I test on macOS, this line will take forever and the extension will not be fully activated. Not sure what caused this while it does not seem to happen on Ubuntu.

ElianHugh: Very weird, I don't get this on WSL either... I'll try to get my hands on a mac to see if I can reproduce

renkun-ken: Is there anything we could improve by learning how vscode-jupyter does this?

https://github.com/microsoft/vscode-jupyter/blob/b1f70d0fc436c9bade9941406b2590b4034cf2bf/src/client/datascience/liveshare/liveshare.ts#L40-L62

ManuelHentschel: If I understood the code correctly, this command tries to activate the live share extension/api when it's called, but the description (and function name) suggest that it only checks if the api is active. Would it make sense to split this into two separate functions (e.g. `isLiveShare` and `activateLiveShare`)? Because if the extension isn't active, there can't be an active session anyways, and then we'd cause unnecessary overhead by activating the live share extension.

src/rShare.ts
```diff
@@ -0,0 +1,322 @@
...
+// service vars
+export const ShareProviderName = 'vscode-r';
+export let service: vsls.SharedServiceProxy | vsls.SharedService | null = undefined;
+
+// random number to fake a UUID for differentiating between
+// host calls and guest calls (specifically for the workspace
+// viewer 'View' function)
+export const UUID = Math.floor(Math.random() * Date.now());
+
+/// state-tracking bools
```
ManuelHentschel, line 28: To make the code of these helper functions easier to read, I'd suggest replacing the patterns
``` ts
// replace
if(CONDITION){
  return true;
} else{
  return false;
}

// with
return !!CONDITION;
```
``` ts
//replace
return A === B ? true : false;

// with
return A === B;
```
``` ts
// replace
if(CONDITION === true){}

// with
if(CONDITION){}
```

src/rShare.ts
```diff
@@ -0,0 +1,322 @@
...
+        this._isStarted = false;
+    }
+    /// Session Syncing ///
+    // These are called from the host in order to tell the guest session
+    // to update the env/request/plot
+    // This way, we don't have to re-create a guest version of the session
+    // watcher, and can rely on the host to tell when something needs to be
+    // updated
+    public notifyGlobalenv(file: string): void {
+        if (this.isStarted) {
```
ManuelHentschel, line 189: Is this supposed to be e.g. `if (this.isStarted())` or `if (this._isStarted)`?

src/rShare.ts
```diff
@@ -0,0 +1,322 @@
...
+                } else {
+                    console.error('[GuestService] failed to retrieve file content (not of type "Buffer")');
+                }
+            }
+        }
+    }
+
+    public async requestHelpContent(file: string): Promise<HelpFile> {
+        const content: string | null | unknown = await request(Callback.GetHelpFileContent, file);
+        if (content !== undefined) {
```
ManuelHentschel, line 293: I'd suggest using `if(content)`, in case the result is `null`

src/rShare.ts
ElianHugh: Thanks for checking out the PR @ManuelHentschel! The code is *intended* to check if there is an active live share session.

`vsls.getApi()).session.id` doesn't (or shouldn't) start a live share session on its own, it should be just returning the api for the live share SDK, which contains information on the current session. If liveshare is installed, `vsls.getApi()` will always return TRUE, which isn't especially useful - hence checking if there is a session id.

I've tried to pinpoint how the liveshare extension activates (though this is a little difficult given it being closed-source). From what I can tell, the extension is always active when installed (and enabled), so the overhead presumably wouldn't be any different to the user having the extension enabled. Admittedly, I'm not sure how to track extension performance...

Having said that, it isn't especially useful to grab the api twice, so I've changed it to rely on the liveSession variable, which is set during `initLiveShare`.

Thank you again!

---

@renkun-ken Hi Ken, thanks again for having a look at the PR. I had a look at the jupyter code, but I was a little confused as to why they re-implement the vsls API. It seems like they've done so to implement some custom commands, but otherwise the functionality seems to be identical to the vsls API. Am I wrong in this understanding?

I've based the PR on a few different liveshare repos:

- https://github.com/vsls-contrib/togezr/blob/7d814f1182efea6dd3d44945423a701e9e34a802/vscode-extension/src/branchBroadcast/liveshare.ts#L33
- https://github.com/vsls-contrib/pairit/blob/master/src/extension.ts#L41-L63
- https://github.com/lostintangent/whiteboard/blob/master/src/service/guestService.ts

ManuelHentschel: > I've tried to pinpoint how the liveshare extension activates


On the extension page you can check Feature Contributions > Activation Events to see when an extension activates. In this case, the list includes `*`, which means that the extension will always be activated on startup, so afaik you don't need to worry about accidentally activating the extension yourself.

ElianHugh:
> I just tested this again and it remains true that sharing terminals (e.g., zsh) in browsers is not supported as per this [tweet](https://twitter.com/FilishaShah/status/1357908285156773890). That said, if the host makes the R terminal read/write in the Live Share extension, it does work on a guest browser even though zsh does not! The R terminal is labeled as `Local` on the guest browser, and seems to bypass `.radian_profile` because on my host machine I have `options(radian.insert_new_line = FALSE)` but the guest browser R terminal inserts new lines between R prompts.

@gvelasq I've had a look at [this codespace FAQ](https://code.visualstudio.com/docs/remote/codespaces) and it looks like the more significant issue is that vscode-R *can't be installed in the browser*, and therefore custom functionality such as this PR isn't available in the browser :(

From the link above:

> #### Why is an extension not installable in the browser
> There are a small number of extensions that have built-in assumptions or need to run on the desktop. Examples are when an extension accesses files from the VS Code installation on the desktop or when an extension depends on a executable that must run in a desktop environment. When you try to install such an extension in the browser, you will be informed that the extension is not available.

gvelasq:
@ElianHugh thanks for looking into the Codespaces FAQ. It's good to know that it's not possible to install the `vscode-R` extension for Live Share guests connecting via a browser. Since Live Share in `vscode-R` will require guests to connect via their own VS Code desktop instance, that suggests to me that ongoing efforts to facilitate installation and configuration for `vscode-R` will also help Live Share uptake for this extension.

Thanks to @renkun-ken and @ManuelHentschel for their technical reviews — what remains to be done before this PR can be merged?

src/rShare.ts
renkun-ken: Let me try again and see if it works.

renkun-ken:
Now the LiveShare does not hang at startup on macOS. But it looks like the guest session cannot attach to the active terminal.

image

renkun-ken:
I tested on Ubuntu and both host and guest session works well with attached session (e.g. `View(mtcars)`.

However, when the host is Ubuntu and the guest is macOS, the guest cannot attach to active terminal. When the host is macOS and the guest is Ubuntu, the guest cannot attach to active terminal either.

ElianHugh:
> 
> 
> I tested on Ubuntu and both host and guest session works well with attached session (e.g. `View(mtcars)`).
> 
> However, when the host is Ubuntu and the guest is macOS, the guest cannot attach to active terminal. When the host is macOS and the guest is Ubuntu, the guest cannot attach to active terminal either.

Very weird, but that does help narrow down the issue a bit. I looked at Ubuntu host to Windows guest and that seems to work well too. I'll have access to a mac shortly, so should be able to pinpoint what the issue is!

ElianHugh:
Hi @renkun-ken, I tried this PR on a mac with a fresh install of vscode, and it seems to work okay for me. I did, however, have to enable session watcher prior to joining the session.

image

So, the source of the issue is likely to stem from the `doesTermExist` command:

```
export function doesTermExist(): boolean {
    const termNames = ['R', 'R Interactive'];
    for (const [i] of vscode.window.terminals.entries()) {
        const terminal = vscode.window.terminals[i];
        const terminalName = terminal.name;
        if (termNames.includes(terminalName)) {
            return true;
        }
    }
    return false;
}
```

Is it possible that the terminal doesn't have the names listed under `const termNames`? That would be the "easiest" answer to this problem I think

renkun-ken:
Thanks for pointing out the `termNames`. I always type `r` to start a radian in terminal (see https://github.com/Ikuyadeu/vscode-R/wiki/R-Session-watcher#advanced-usage-for-self-managed-r-sessions) but the automatically determined terminal name is `python3`).

I use `R: Create R Terminal` or manually start an `radian` and rename the terminal to a proper name `R`, then share the session, everything works well now on both Ubuntu and macOS.

I notice that if I don't have an active terminal before sharing the session and a guest joins, then starting an active terminal does not make the guest attach to the terminal. The only way to make it work is to attach to a session before a guest joins.

renkun-ken:
Do you think it is handy to support join before attach? If not, would you like to write more in the error message so that user knows what should be done?

ElianHugh:
> Do you think it is handy to support join before attach? If not, would you like to write more in the error message so that user knows what should be done?

Guest attach should definitely mirror host attach at all times (aside from when the workspace isn't shared). I'll have a commit to fix this soon

ElianHugh:
> 
> 
> Thanks for pointing out the `termNames`. I always type `r` to start a radian in terminal (see [Ikuyadeu/vscode-R/wiki/R-Session-watcher#advanced-usage-for-self-managed-r-sessions](https://github.com/Ikuyadeu/vscode-R/wiki/R-Session-watcher#advanced-usage-for-self-managed-r-sessions)) but the automatically determined terminal name is `python3`).
> 

To get around the name issue, I can make use of `!!pid` on the host side. However, the issue is that the pid isn't removed once the terminal is disposed, so this method could be a bit clunky in some cases

renkun-ken:
>To get around the name issue, I can make use of !!pid on the host side. However, the issue is that the pid isn't removed once the terminal is disposed, so this method could be a bit clunky in some cases

I wonder why we really need a check or filter here?

ElianHugh:
> 
> 
> > To get around the name issue, I can make use of !!pid on the host side. However, the issue is that the pid isn't removed once the terminal is disposed, so this method could be a bit clunky in some cases
> 
> I wonder why we really need a check or filter here?

You're probably right - previously it was to prevent undefined objects crashing the guest session, but the upcoming changes don't really warrant it

src/liveshare/share.ts
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// Listens for the activation of a LiveShare session
+export async function LiveSessionListener(): Promise<void> {
+    rHostService = new HostService;
+    rGuestService = new GuestService;
+
+    // Return out when the vsls extension isn't
+    // installed/available
+    const liveSessionStatus = await vsls.getApi();
```
renkun-ken, line 90: When I enabled LiveShare extension on macOS and start debugging the extension or have the extension built and installed, vscode-R could not be fully activated but still stuck here.

renkun-ken: Is it possible to use a timeout here?

renkun-ken: What about

```r
const liveSessionStatus = await Promise.race([
        vsls.getApi(),
        new Promise((res) => setTimeout(() => res(null), 1000))
    ]);
    if (!liveSessionStatus) { return; }
    liveSession = liveSessionStatus as vsls.LiveShare;
```

renkun-ken: Does the code in vscode-jupyter help? https://github.com/microsoft/vscode-jupyter/blob/b1f70d0fc436c9bade9941406b2590b4034cf2bf/src/client/datascience/liveshare/liveshare.ts#L47-L61

ElianHugh: I like the timeout solution you wrote above - it works nicely with a setting to tweak the timeout speed.

We could, like vscode-jupyter, have a bool to hard-toggle LiveShare functionality. I wonder if it might get confusing with overlap in functionality with the PR's tree view commands?

renkun-ken:
Thanks for the update. I works quite nicely now.

renkun-ken:
@ElianHugh Would you like to resolve all conflicts here?

ElianHugh:
> 
> 
> @ElianHugh Would you like to resolve all conflicts here?

Will do! Just fixing up some conflicts

ElianHugh:
Looks like I might have messed up the merge somewhere, will fix tomorrow

ElianHugh:
Alright, that should fix up the errors! One thing I can't seem to get to work is reopening guest browsers when they've been closed. E.g. if the guest closes a httpgd browser, calling another plot won't reopen the webview

package.json
```diff
@@ -1150,6 +1186,26 @@
           "default": false,
           "description": "Remove hidden items when clearing workspace."
         },
+        "r.liveShare.timeout": {
+          "type": "integer",
+          "default": 3000,
+          "description": "Time in milliseconds before aborting attempt to connect to Live Share API"
+        },
+        "r.liveShare.commandForwardDefault": {
```
renkun-ken, line 1194: I guess we could follow the naming style `r.liveShare.defaults.commandForward`, `r.liveShare.defaults.shareWorkspace`, `r.liveShare.defaults.shareBrowser`?

It is consistent with the defaults introduces in https://github.com/Ikuyadeu/vscode-R/pull/620/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R1153.

src/extension.ts
```diff
@@ -165,20 +172,21 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
...
-        // create status bar item that contains info about the session watcher
-        console.info('Create sessionStatusBarItem');
-        const sessionStatusBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Right, 1000);
-        sessionStatusBarItem.command = 'r.attachActive';
-        sessionStatusBarItem.text = 'R: (not attached)';
-        sessionStatusBarItem.tooltip = 'Attach Active Terminal';
-        sessionStatusBarItem.show();
-        context.subscriptions.push(sessionStatusBarItem);
-        session.startRequestWatcher(sessionStatusBarItem);
+        if (!(rShare.isGuest())) {
```
renkun-ken, line 176: Seems unnecessary to have that many parentheses.

ElianHugh:
Hey @ManuelHentschel, you're probably the best person to ask re: the plot viewer. I'm trying to share the httpgd server with guests, which currently only works if the viewer is started *after* starting a live share session. In order to get around this, I've been trying to hook into where the viewer is refreshed and share the server from there - but haven't had much luck. Mostly I'm unsure exactly which method to hook into that will ensure the server is shared.

ManuelHentschel:
[Here](https://github.com/ManuelHentschel/vscode-R/blob/77f982b8e96dad3e4e0f2b058a211f11409bcc81/src/plotViewer/index.ts#L331-L336) the viewer gets notified about changes to the plot and then requests the current state of the httpgd server in `HttpgdViewer.checkState()` and `HttpgdViewer.refreshPlots()`. Are these what you're looking for?

ElianHugh:
> [Here](https://github.com/ManuelHentschel/vscode-R/blob/77f982b8e96dad3e4e0f2b058a211f11409bcc81/src/plotViewer/index.ts#L331-L336) the viewer gets notified about changes to the plot and then requests the current state of the httpgd server in `HttpgdViewer.checkState()` and `HttpgdViewer.refreshPlots()`. Are these what you're looking for?

Thank you, exactly what I was looking for!

renkun-ken:
I'm testing the latest build of this PR and it seems the guest can no longer attach to host session however I tried.

ElianHugh:
> I'm testing the latest build of this PR and it seems the guest can no longer attach to host session however I tried.

I wonder if this is due to the API timeout implemented previously? I've noticed the same too, but usually restarting the extension fixes it

renkun-ken:
Looks like it is indeed caused by the default timeout 3000. I adjust it to 5000 and it works now.

renkun-ken:
Let me test again on macOS and take a closer look at the behavior of vsls.getApi which could take forever as I mentioned.

renkun-ken:
I tested on macOS and I have to increase the timeout to 8000 to make it work.

Let me take a closer look into this.

renkun-ken:
Looks like `vsls.getApi()` does not hang any more (not sure why). I guess we could set the timeout to 10000 which is quite safe for most cases.

renkun-ken:
I found a quirky behavior of guest httpd plot viewer. Make sure to put the following in the `~/.Rprofile` to enable native httpgd plot viewer:

```r
options(vsc.use_httpgd = TRUE)
```

In a live share session, start an R session and try the following code:

```r
plot(rnorm(100))
```

which opens up an httpgd plot viewer tab. Then create a shinyApp:

```r
shiny::runExample("01_hello")
```

A brower tab is shown. Put these two viewers in the same view column in vscode window and moving the slide in the shiny app will always reveal the httpgd plot viewer.

ElianHugh:
> A brower tab is shown. On both host and guest window, put these two viewers in the same editor view column. Moving the slide in the shiny app will always reveal the httpgd plot viewer on the guest side but not on the host side.


I could replicate this - looks like the `notifyGuestPlotManager` call was too eager in updating. I've moved this under the checkState method, and should now be fixed

renkun-ken:
Thanks @ElianHugh for the quick fix. It works well now.

R/vsc.R
```diff
@@ -323,7 +323,7 @@ if (show_view) {
     list(columns = columns, data = data)
   }
 
-  show_dataview <- function(x, title,
+  show_dataview <- function(x, title, .UUID = NULL,
```
renkun-ken, line 326: Is there a reason to use `.UUID` instead of `UUID` here? I don't see a potential conflict or misuse here.

Also, it looks a bit inconsistent with the lower-case argument code-style in the rest of R code.

Would you mind changing all `UUID` into `uuid`?

renkun-ken:
Once this PR is merged, I suggest that we add @ElianHugh as a collaborator since this is such a nice and important feature which needs good maintainence. @Ikuyadeu @andycraig @ManuelHentschel

R/vsc.R
ElianHugh: I was using `.UUID` in case of accidental argument matching, but it is unlikely that something would go wrong there. Lowercase makes sense too!

renkun-ken (approved):
LGTM Thanks for the nice work!

renkun-ken:
@gvelasq @ManuelHentschel Any more comments?

src/plotViewer/index.ts
```diff
@@ -46,13 +49,15 @@ export function initializeHttpgd(): HttpgdManager {
     return httpgdManager;
 }
 
+export let shareUrl = undefined;
```
ManuelHentschel, line 52: Is it necessary to export this?

src/plotViewer/index.ts
```diff
@@ -515,9 +522,13 @@ export class HttpgdViewer implements IHttpgdViewer {
         this.state = await this.api.getState();
         if(this.state.upid !== oldUpid){
             await this.refreshPlots();
+            if (isHost()) {
+                void shareBrowser(shareUrl, '[VSC-R] R Plot', true);
+                void rHostService.notifyGuestPlotManager(shareUrl);
```
ManuelHentschel, line 527: Instead of `shareUrl` can we use `this.host`? This would be safer, since there can be multiple Httpgd viewers at the same time and we'd avoid an (unnecessary?) global variable.

src/liveshare/shareCommands.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (isGuest()) {
+        // is guest service
+        (service as vsls.SharedServiceProxy).onNotify(name, command as vsls.NotifyHandler);
+    } else {
+        // is host service
+        (service as vsls.SharedService).onRequest(name, command as vsls.RequestHandler);
+    }
+}
+
+export function request(name: string, ...rest: unknown[]): unknown {
```
ManuelHentschel, line 147: To avoid conusion, should we maybe use a more descriptive name? E.g. `liveShareRequest` etc.

README.md
```diff
@@ -1,5 +1,7 @@
 # R support for Visual Studio Code
 
+[![Badge](https://aka.ms/vsls-badge)](https://aka.ms/vsls)
+
```
ManuelHentschel, line 4: So far, we don't have any other badges (e.g. version, cicd, license, installs, rating...) in the readme. Personally I'd suggest we either agree on a sensible set of badges or just not have any badges at all.

src/liveshare/share.ts
```diff
@@ -0,0 +1,336 @@
+import * as vscode from 'vscode';
```
ManuelHentschel, line 1: A design pattern I personally like (e.g. in the helpViewer files) is to name the main file `index.ts` and only import from this file elsewhere in the extension. If necessary, items can also be re-exported here.

This makes it easier to work with/on the liveshare related functions, since you have a single file from where you import these, and you only have to check the impact on the other liveshare files when changing liveshare-related functions that are not exported in `index.ts`.

I think that's just personal preference though, so feel free to mark as resolved if you don't want to change this!

ManuelHentschel (commented):
Nice work, thanks!

src/plotViewer/index.ts
ElianHugh: It seems that `this.host` isn't a valid url to share, but `http://${this.host}/live?token=${this.token}` works

src/liveshare/share.ts
ElianHugh: Ah, that's pretty elegant! Thank you for the tip.

README.md
ElianHugh: Makes sense to me - I'll remove the badge for now

renkun-ken: Putting a single badge is perfectly ok to me, just like <https://github.com/Microsoft/vscode-cpptools> 😺

ManuelHentschel: I guess this is just a matter of personal preference, so feel free to leave the badge in, if you agree on this :)

ManuelHentschel (approved):
LGTM (Please remember to fix the linter warnings before merging)

ElianHugh:
Finger slipped on the GitHub app :)

Thank you for the kind words, and the in-depth reviews! I will fix up the linter issues shortly

renkun-ken:
Looks like this PR is good to merge now.

gvelasq:
> Looks like this PR is good to merge now.

Thank you for all of your amazing work on this!

Ikuyadeu:
@ElianHugh
Great contribution!
Now, I send the collaborator invitation to you.
I'm happy if you have interested to join us.

As mentioned in # 98, the management of this repository will be moved from me within a year.
However, collaborators' access will remain the same.
I hope you will participate in the development for a long time.

MilesMcBain:
Congratulations all this is looking incredible!!

ElianHugh:
> @ElianHugh
> Great contribution!
> Now, I send the collaborator invitation to you.
> I'm happy if you have interested to join us.
> 
> As mentioned in # 98, the management of this repository will be moved from me within a year.
> However, collaborators' access will remain the same.
> I hope you will participate in the development for a long time.

Thank you for the kind words and the invitation! More than happy to help out :)

PR #641, opened 2021-05-17
Improve development workflow

ManuelHentschel:
This PR makes working on the extension a bit easier.

By default, I disabled webpack and replaced the prelaunch task with a tsc watch task, which is considerably faster (especially on subsequent launches) than the webpack task. The script "scripts/enableWebpack.js" enables webPack again and can be called e.g. before publishing the extension. I also added some tsc watch tasks to compile the help/httpgd related typescript, but did not include these in the normal watch task, since most of the time they won't be changed anyways. Other possible modifications include

- [x] Adjusting the release actions to keep using webpack
- [x] Using yarn instead of npm. This basically comes down to replacing the lock file. In summary, yarn is compatible with the current package.json, but significantly faster. It's a bit less common than npm, apart from that I'm not aware of any drawbacks. For me personally, the switch would be worth it, but I also see the point that it might make running the extension from code a bit harder for new contributors.
- [x] Using subdirectories/other names for the typescript source files. Currently, there are some files (e.g. the help related stuff, or the upcoming liveshare files) that might as well be in their own folder. And I guess in most cases, we could remove the `r` from the beginning of file names, since they're all related to R anyways.

For the future, a slightly bigger change would be to set `"strictNullChecks": true` in tsconfig.json. This requires developers to work more cautiously when dealing with possibly undefined variables and requires rather many adjustments to the code base, since it cannot be turned off on a per-file basis (as opposed to eslint rules). But considering the rather strict rules implemented in `.eslintrc.yml`, I think it would make sense to also be strict here. 

**How to test:**
Behavior of the installed extension shouldn't change.
When running the extension from code, startup should be significantly faster than with webpack.
When the package is published on Github (as build artifact, pre-release, release) or the vscode marketplace, webpack should be enabled again.


ManuelHentschel:
I guess using yarn and organizing files in subdirectories is optional anyways, so I wouldn't mind skpiping it here.

Ikuyadeu:
We used yarn for a while, but I think I switched back to npm due to compatibility issues with webpack or GitHub Action.
If yarn is available, I think it's best.

ManuelHentschel:
> We used yarn for a while, but I think I switched back to npm due to compatibility issues with webpack or GitHub Action.
> If yarn is available, I think it's best.

Using yarn instead of npm seems to work fine, both with and without webpack. If you find any problems with it, let me know.

Ikuyadeu:
> Using yarn instead of npm seems to work fine, both with and without webpack. If you find any problems with it, let me know.

OK, changing that is not so large work. And we can check bugs on this PR.

> Using subdirectories/other names for the typescript source files. Currently, there are some files (e.g. the help related stuff, or the upcoming liveshare files) that might as well be in their own folder. And I guess in most cases, we could remove the r from the beginning of file names, since they're all related to R anyways.

It sounds great, and I think, we can put them in one folder such as `help`, `http`.

ManuelHentschel:
I moved help and plot related code to their own folders, and using yarn over npm seems to work fine with Github actions/vsce

Ikuyadeu:
LGTM, Thank you.
I'll review more detail this week.

BTW, for release process, I am planning to use a draft release instead of the latest, that do not need to update the tag index.
https://github.com/release-drafter/release-drafter

Ikuyadeu:
@ManuelHentschel 
This point was that I should have reviewed before this PR, but `src / helpViewer / index.ts` and the other `helpviewer` modules are interdependent.
This design complicates the modifications when editing each other.
In my idea, one solution is to divide the `index.ts` functions into other files such as `util.ts`.

I want to merge this PR once and solve this problem with a new PR, is that okay? Or do you want to do it here?

Ikuyadeu:
For other parts, I think we can merge them.

ManuelHentschel:
> 
> I want to merge this PR once and solve this problem with a new PR, is that okay? Or do you want to do it here?

I agree that it's best to address that issue in a different PR and merge this PR for now.

PR #647, opened 2021-05-27
Update R syntax

### LINKED ISSUE #646: Native pipe operator `|>` support
**Describe the bug**
When executing a line using `R: Run Selection/Line` with a line feed after the native pipe operator `|>` added in R4.1.0, only the line with the cursor will be executed.

Is it because the native pipe operator is not registered here?

https://github.com/Ikuyadeu/vscode-R/blob/5925d08c5a26479249ad06b5b820d7439b81a950/syntax/r.json#L192-L193

**To Reproduce**
Place the cursor on each of the end of `head()` and try to run the line (Ctrl + Enter).

```r
mtcars |> head()
mtcars |>
    head()
mtcars %>%
    head()
```

**Expected behavior**
Two lines will be executed at the same time, as in magrittr's pipe operator `%>%`.

**Screenshots**
image

- OS: Ubuntu (Windows Docker Desktop)
- VSCode Version: 1.56.2
- R Version: 4.1.0
- vscode-R version: 1.6.6

renkun-ken:
# What problem did you solve?

Closes #646 

1. Put cursor on any point of the following code and send code to terminal should send the whole pipeline to terminal.

```r
1:10 |>
  sapply(\(x) x^2) |>
  sum()
```

### DIFF
--- a/src/lineCache.ts
+++ b/src/lineCache.ts
@@ -67,7 +67,7 @@ export function cleanLine(text: string): string {
 }
 
 function doesLineEndInOperator(text: string) {
-    const endingOperatorIndex = text.search(/(,|\+|!|\$|\^|&|\*|-|=|:|~|\||\/|\?|%.*%)(\s*|\s*#.*)$/);
+    const endingOperatorIndex = text.search(/(,|\+|!|\$|\^|&|\*|-|=|:|~|\||\/|\?|<|>|%.*%)(\s*|\s*#.*)$/);
     const spacesOnlyIndex = text.search(/^\s*$/);
 
     return ((endingOperatorIndex >= 0) || (spacesOnlyIndex >= 0));
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -189,6 +189,10 @@
                     "name": "keyword.operator.logical.r"
                 },
                 {
+                    "match": "(\\|>)",
+                    "name": "keyword.operator.pipe.r"
+                },
+                {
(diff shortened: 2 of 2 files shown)


renkun-ken:
TBH I'm not quite familiar with the syntax file. Could you please point out if I don't do it right? @andycraig @Ikuyadeu

Ikuyadeu:
OK, I'll check this weekend.
And, I'll try to make the test code for CI check.

renkun-ken:
Should we do anything about the pipe-bind operator (`=>`) which is disabled by default unless environment variable `_R_USE_PIPEBIND_` is set to a true value?

ElianHugh:
Maybe for future proofing? Depends on whether the pipe bind operator is going to stay or not - [this](https://r.789695.n4.nabble.com/Pipe-bind-restored-in-R-4-1-0-td4768620.html) would lead me to believe it is, but then [this twitter thread](https://twitter.com/LukeTierney4/status/1399158922402054144?s=20) seems to go against that.

Ikuyadeu:
@renkun-ken 
OK, I checked the code. Thank you!
But testing CI is a bit hard now. So, I merge this once.

And, `=>` should be let now, for confusing replacement. I think users should set that manually.

renkun-ken:
Do you encounter that most highlighting is lost with latest release installed?

renkun-ken:
Looks like https://github.com/Ikuyadeu/vscode-R/pull/647/files#diff-f6f05bbc187255f5cf06a39e3a4f7befa7288489c5fd44bb14586143076ffd1dR407-R447 is somehow problematic. I remove that code and the highlighting goes back to normal.

PR #649, opened 2021-05-27
Prefer rPath from PATH

ManuelHentschel:

This PR changes the order in which the extension tries to find an R path on windows.

I think it makes sense to prefer `PATH` over the registry since this is more transparent and makes it easier to switch between R versions systemwide (i.e. normal terminals and vscode-R).

Leave `r.rterm.windows` empty, have different r paths in the `PATH` environment variable and registry, and run `R: create R Terminal`

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -35,7 +35,9 @@ export async function getRpathFromSystem(): Promise<string> {
     let rpath = '';
     const platform: string = process.platform;
     
-    if ( platform === 'win32') {
+    rpath ||= getRfromEnvPath(platform);
+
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
`PATH` taking the highest precedence in the determination of R path makes good sense.

renkun-ken:
Are you interested in putting up a PR to also update [vscode-r-lsp util.ts](https://github.com/REditorSupport/vscode-r-lsp/blob/master/src/util.ts) so that they are consistent in finding R path?

PR #650, opened 2021-05-27
Fix Webview Identification Function

ManuelHentschel:
In the most recent insider-build of vscode, some webview commands are broken because the internal function `Webview.asWebviewUri()` changed its behavior.

Should fix the issue mentioned above.


renkun-ken (approved):
LGTM

PR #653, opened 2021-05-29
Don't run chunks with eval = FALSE. Fixes #651

### LINKED ISSUE #651: Don't run `eval = FALSE` R Markdown chunks with code lens commands

I sometimes put non-working/incomplete/only-for-demonstration code in R Markdown chunks and use `eval = FALSE` to have them not run when the document is knitted. When using the code lens commands in the extension, such as "Run Above", these chunks are still evaluated, which often causes errors.

In R Studio, using "Run All" or "Run All Chunks Above" will skip running chunks with `eval = FALSE`, which I think is the correct behavior. I am hoping that this extension will adopt the same behavior.

Since vscode-R is already discovering R markdown chunks with what I suppose is some kind of regular expression, my naive guess is that it should be straightforward to also look for `eval = FALSE` in the code chunk header. I will probably be a little more work to get the extension to respect global `knitr::opts_chunk` settings in combination with code chunk header options, but that would of course be optimal.

I haven't considered any alternatives here.

jolars:
# What problem did you solve?

When using the code lens actions in the extension, such as "Run Above", code chunks with `eval = FALSE` are evaluated although they probably should not be since they often contain pseudo-code or other things that are not intended to actually be run. The default behavior in R Studio, for instance, is not to run these code chunks. See #653 for more details.

The extension already collects information on the value of `eval`, so the fix is just to implement a few if-clauses to prevent evaluation when `eval = FALSE`. I purposely did not include actions `runNextChunk`, `runPreviousChunk`, `runCurrentChunk`, and `runCurrentAndBelowChunks` since I think they signal that the user explicitly wants the specific chunk to be run. This also conforms to the behavior in R Studio when using "Run Next Chunk" and "Run Current Chunk".

## Screenshot

Peek 2021-05-29 22-00

### DIFF
--- a/src/rmarkdown.ts
+++ b/src/rmarkdown.ts
@@ -314,7 +314,9 @@ export async function runAboveChunks(chunks: RMarkdownChunk[] = _getChunks(),
 
   for (let i = firstChunkId ; i <= previousChunkId ; i++) {
     const chunk = chunks.filter(e => e.id === i)[0];
-    codeRanges.push(chunk.codeRange);
+    if (chunk.eval) {
+      codeRanges.push(chunk.codeRange);
+    }
   }
   await runChunksInTerm(codeRanges);
 }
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

Thanks!

PR #658, opened 2021-05-31
Fix bugs in helpviewer

ManuelHentschel:
Fixes minor bugs in the help viewer

* Install a single package using the treeview
* Use `Open help for selection` on a function that is provided by two different packages, press ESC when prompted to select a topic (previously an error message would pop up)


renkun-ken (approved):
LGTM

Works well on Ubuntu.

PR #662, opened 2021-06-07
Code cells in .R files

### LINKED ISSUE #191: add a way to declare a cell and a command run cell
Please considering add a way to declare a cell like `#%%` so that users can run code from cell to cell without selecting multiple lines of code.

example:
https://code.visualstudio.com/docs/python/jupyter-support-py#_jupyter-code-cells
https://nteract.gitbooks.io/hydrogen/

ElianHugh:
Implements jupyter code block style decorators for .R files, piggybacking on the previous rmarkdown code lens implementation.

# What problem did you solve?
Closes #191 

## Screenshot

120947678-be001980-c72f-11eb-91c3-4dae45e42c6b

Open .R file, and make a code block like so:

```r
# %% I am a chunk
mtcars
# %%

# %% So am I
2 * 2
# %%

```

Markdown code fences should *not* work in .R files, and the new decorator should *not* work in rmd files

### DIFF
--- a/src/rmarkdown.ts
+++ b/src/rmarkdown.ts
@@ -7,7 +7,7 @@ import { runChunksInTerm } from './rTerminal';
 import { config } from './util';
 
 function isRDocument(document: TextDocument) {
-  return (languages.match('r', document) > 0);
+  return (document.languageId === 'r');
 }
 
 function isRChunkLine(text: string) {
--- a/src/rmarkdown.ts
+++ b/src/rmarkdown.ts
@@ -1,7 +1,7 @@
 import {
   CancellationToken, CodeLens, CodeLensProvider,
   CompletionItem, CompletionItemProvider,
-  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection, languages
+  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection
 } from 'vscode';
 import { runChunksInTerm } from './rTerminal';
 import { config } from './util';
--- a/src/rmarkdown.ts
+++ b/src/rmarkdown.ts
@@ -1,23 +1,33 @@
 import {
   CancellationToken, CodeLens, CodeLensProvider,
   CompletionItem, CompletionItemProvider,
-  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection
+  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection, languages
 } from 'vscode';
 import { runChunksInTerm } from './rTerminal';
 import { config } from './util';
 
-function isChunkStartLine(text: string) {
-  if (text.match(/^\s*```+\s*\{\w+\s*.*$/g)) {
-    return true;
-  }
-  return false;
+function isRDocument(document: TextDocument) {
+  return (languages.match('r', document) > 0);
 }
 
-function isChunkEndLine(text: string) {
-  if (text.match(/^\s*```+\s*$/g)) {
-    return true;
+function isRChunkLine(text: string) {
+  return (!!text.match(/^#+\s*%%/g));
+}
+
+function isChunkStartLine(text: string, isRDoc: boolean) {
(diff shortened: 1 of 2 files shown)


src/rmarkdown.ts
```diff
@@ -1,23 +1,33 @@
...
+  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection, languages
 } from 'vscode';
 import { runChunksInTerm } from './rTerminal';
 import { config } from './util';
 
-function isChunkStartLine(text: string) {
-  if (text.match(/^\s*```+\s*\{\w+\s*.*$/g)) {
-    return true;
+function isRDocument(document: TextDocument) {
+  return (languages.match('r', document) > 0);
```
renkun-ken, line 10: Is there a difference using `document.languageId` instead of `languages.match('r', document) > 0`>

ElianHugh: I haven't found any difference, thanks for picking that up! Not sure what the use case would be if languageId exists

src/rmarkdown.ts
```diff
@@ -1,23 +1,33 @@
 import {
   CancellationToken, CodeLens, CodeLensProvider,
   CompletionItem, CompletionItemProvider,
-  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection
+  Event, EventEmitter, Position, Range, TextDocument, TextEditorDecorationType, window, Selection, languages
```
renkun-ken, line 4: `languages` seems unnecessary now.

renkun-ken (approved):
LGTM

apoorvalal:
Thanks for implementing this! Was wondering if it wouldn't be too much trouble to make cell delimiters single `#%%`s instead of requiring both top and bottom delimiters.

For example, percent scripts produced by [jupytext](https://jupytext.readthedocs.io/en/latest/formats.html#the-percent-format) currently skip alternate cells because the `#%%` at the bottom of a the i-th cell isn't interpreted as the beginning of the i+1-th cell.
image
 
I'm assuming this is because the current implementation replaces the rmarkdown ``` delimiter with `#%%`, so it expects both top and bottom fences.

Also, seems like the shortcut to run chunk doesn't work for R scripts fenced by `#%%`; it runs the entire script when i use `Ctrl+Shift+Enter` (run chunk shortcut on linux). Clicking on `Run Chunk`, however, produces the desired behaviour (i.e. it runs just the current cell).

ElianHugh:
Thank you for the feedback! Just to be 100% clear on expected functionality, would it work like the following:

`````
# %% 
chunk 1

# %%
chunk 2




still chunk 2
``````

i.e., a chunk continues until a new chunk is defined via `# %%`?

apoorvalal:
Yep, exactly. Since there are no text blocks in scripts, there's no need for a separate bottom delimiter.

Yunuuuu:
Thanks for this excellent feature. Can the background between both `# %%` be customized?

Yunuuuu:
Thanks! renkun! I didn't realize this is the same with rmarkdown. In this case, the problem raised in #901, which indicated `Ctrl + Shift + ENTER` couldn't work  to run current chunk, is because of the `when` close:
image

renkun-ken:
Could you try changing the `when` clause to `editorTextFocus && (editorLangId == 'r' || editorLangId == 'rmd')` and see if it works?

Yunuuuu:
yeah, this can work. I have tried it last night. Just post here to let @ElianHugh know.

renkun-ken:
@ElianHugh Looks like we should allow all these commands to run in both `r` and `rmd`?

Yunuuuu:
@renkun-ken, thanks for your help. I find the paranthesis is not supported in when clause, as pointed in https://github.com/microsoft/vscode/issues/91473. so the when clause here should be `editorLangId == r && editorTextFocus || editorLangId == rmd && editorTextFocus`

ElianHugh:
Seems like a good change to me! I'd like to push this along with some of the jupyter code cell changes, but feel free to push a change sooner if needed

PR #663, opened 2021-06-08
rmarkdown bug squashing and minor changes

### LINKED ISSUE #473: `Run chunk` command  threw error on the first and last chunk
For the following rmd:

````
# Title

This is a test document.

## Section 1

```{r}
1
```

```{r}
2
```

```{r,eval=FALSE}
3
```

```{python}
def fun(x,y):
  return x+y
```

```{sql}
select * from table
```

```{R chunk-1}
4
```

## Section 2

```{r plot1}
5
```

```{r plot2}
6
```
````

The run chunks commands seem to ignore chunk language, i.e. they run chunks of all languages. Instead, they should only send R chunks to the terminal.

Also, some edge cases are not properly handled yet. For example, Run Above the first chunk, or Run Below the last chunk will cause errors like

image

_Originally posted by @renkun-ken in https://github.com/Ikuyadeu/vscode-R/issues/465#issuecomment-735230484_

### PR DESCRIPTION
**Pull request by ElianHugh**
Closes #473, #639, #633

1. Prevents vscode errors from showing when attempting to run chunks that do not exist. e.g.
   - `goToPrevious`, `runPrevious` & `runAbove` on chunk.id === 1
   - `goToNext`, `runNext` & `runBelow` on last chunk (chunk.id === chunks.length)
2. Change run and knit icons as per #639
3. `Go to chunk` now moves the screen to the cursor position, instead of leaving the cursor at a new off-screen chunk

With regard to [1], I'm not sure if we should have a more informative error message, or just run the current chunk, or just assume that the user will understand why nothing happened. At present, I've landed on option 3, but happy to change.

# Screenshot

**Knit icon**

image

**Source icon**

image

# How can I test?

In a .R file, attempt to move between chunks, and run above/below, etc.
```r
# %%
print("chunk 1")

# %%

# %%
print("chunk 2")

# %%

```


renkun-ken:
Thanks for working on this.

The go-to chunk commands work with rmd but not newly implemented R chunks here.

image

Am I missing something?

ElianHugh:
> Thanks for working on this.
> 
> The go-to chunk commands work with rmd but not newly implemented R chunks here.
> 
> image
> 
> Am I missing something?

Very weird, I can't replicate this on linux Arch, it should be working on both rmd and R chunks

renkun-ken:
Looks like this line 
https://github.com/Ikuyadeu/vscode-R/pull/663/files#diff-7241b38d1bf7bb9f45ef7397474b27e7ee5be116df6aa401ba2bf7e1fd3baea0R256 is problematic.

```
# %%
print("chunk 1")

# %%

# %%
print("chunk 2")
|
# %%
```

When cursor is at `|` in the above example, and I run "Go to previous chunk", and the error occurs with 

```typescript
const currentChunk = chunks.find(i => i.startLine <= line && i.endLine >= line);
```

where `line=8` which is not the start line of any chunk.

renkun-ken:
I did a quick fix in `getCurrentChunk` and also replaces most `chunks.filter(pred)[0]` with `chunks.find(pred)`.

ElianHugh:
> Looks like this line 
> https://github.com/Ikuyadeu/vscode-R/pull/663/files#diff-7241b38d1bf7bb9f45ef7397474b27e7ee5be116df6aa401ba2bf7e1fd3baea0R256 is problematic.
> 
> ```
> # %%
> print("chunk 1")
> 
> # %%
> 
> # %%
> print("chunk 2")
> |
> # %%
> ```
> 
> When cursor is at `|` in the above example, and I run "Go to previous chunk", and the error occurs with 
> 
> ```typescript
> const currentChunk = chunks.find(i => i.startLine <= line && i.endLine >= line);
> ```
> 
> where `line=8` which is not the start line of any chunk.

Ah, my bust, thanks for picking up on that!

renkun-ken (approved):
LGTM

package.json
```diff
@@ -320,19 +320,13 @@
...
-          "dark": "./images/DownloadFile_inverse.ico"
-        }
+        "icon": "$(run)"
       },
       {
         "title": "Knit Rmd",
         "category": "R",
         "command": "r.knitRmd",
-        "icon": {
-          "light": "./images/DownloadFile.ico",
```
renkun-ken, line 333: These ico files seem to be no longer used anywhere. Should we remove them?

Ikuyadeu: OK, thank you for you review!

Issue #668, opened 2021-06-13
Simplify README and update wiki


renkun-ken:
The README seems too long to be an effective introduction to the repo, and plenty of information presented in it is out-dated. I believe that in README, only the most important and up-to-date information should be presented.

I'd like to restructure (mostly simplify) the README to make everything brief and put links to wiki as the detailed documentation, i.e. the following points:

* Quick start
* Features
* Questions, bug reports, feature requests, and contribution

The original content should be mostly moved to wiki and I'll put links in "Features". Also, extensions we recommend users to install with vscode-R such as vscode-r-lsp and VSCode-R-Debugger should be introduced in README, so that user could get full R development experience including language service and debugging capabilities immediately.

ElianHugh:
I think this would go a long way to making vscode-R more palatable for newcomers (see also some comments at #496 ).

There are a few repos that also have some nice formatting for reference:
- [julia-vscode](https://github.com/julia-vscode/julia-vscode)
- [vscode-python](https://github.com/Microsoft/vscode-python)

renkun-ken:
Yes, those repos are exactly some good examples I'm looking at.

Ikuyadeu:
Good, so we can put only three things like these extensions.

* `Install` or `Quick Start`
* Listing the functions and these detail links
* Adding QA link and document link
* ~~Data and telemetry~~ (We do not collect)

renkun-ken:
Also, I suggest we turn on `r.sessionWatcher` by default and make it no longer labled as experimental.

PR #150, opened 2019-12-03
R session watcher

renkun-ken:
**What problem did you solve?**

This PR is an attempt to implement the file-based approach to communicating between vscode-R and a live R session suggested in #143, which also solves #142, #51, and will potentially provide nicer solution to #15, #22, #25, and all other session-related issues (if I'm missing some). Thanks @andycraig for the initial implementation of attaching to active session and showing hover on symbols in globalenv in active session.

The reason I suggest a file-based mechanism is to allow flexible management of R sessions. Under  such mechanism, user can easily work with any R sessions, no matter whether it is official R or radian, whether it is started by vscode-R or by user, whether it is in a tmux or screen window, whether there are multiple R sessions, whether it is local or remote. Such loose connection between vscode-R and live R sessions would not impose much constraints on the stability and availability of both sides, because there's no communication between two top-level user inputs.

An R script `R/init.R` in included in this extension to enable the file-based communication. On extension activation, it would be copy to `~/.vscode-R/init.R`. To enable R session watcher, user should add `source("~/.vscode-R/init.R")` to `~/.Rprofile` and then enable `r.sessionWatcher` experimental option so that the script is executed on R startup.

This PR implements the following features:

* Attach Active Terminal (by command or clicking status bar item)
* Auto attach on R session startup: if `init.R` is sourced in `.Rprofile`, starting an R session will notify vscode-R to automatically attach to it.
* Provide hover to global symbol in attached session
* Show plot file on the fly
* Show WebView to present htmlwidgets and shiny apps
* Show WebView for `data.frame` and `list` object when calling `View()`

This PR also enables the following features to be implemented in the future:

* Provide completion to global symbol in attached session
* Provide completion to list-like objects (list, environment, S4, R6, etc.) in attached session
* Plot history navigation
* WebView history navigation

** Screenshots **

Symbol hover: (first lines provided by latest languageserver, last line provided by vscode-R session watcher)

image

image

Basic plot and ggplot2:

image

image

Show htmlwidgets in WebView:

image

image

Show shiny app:

image

View data.frame:

image

View list:

image

Animation:

image

PR #670, opened 2021-06-14
Enable r.sessionWatcher by default

renkun-ken:
# What problem did you solve?

Since a file-based session watcher was introduced in #150 one and a half years ago, a lot of features has been implemented around it, and they work well in a wide range of scenarios. This PR changes the status of the session watcher from *experimental* to *normal* by enabling `r.sessionWatcher` by default in the setting as I suggested at <https://github.com/Ikuyadeu/vscode-R/issues/668#issuecomment-860214074>.

The setting of `r.sessionWatcher` is enabled by default.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -57,8 +57,8 @@ Yes / No
 // Use bracketed paste mode
 "r.bracketedPaste": false,
 
-// Enable R session watcher (experimental)
-"r.sessionWatcher": false,
+// Enable R session watcher
+"r.sessionWatcher": true,
 
 // Delay in milliseconds before sending each line to rterm (only applies if r.bracketedPaste is false)
 "r.rtermSendDelay": 8,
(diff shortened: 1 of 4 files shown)


ElianHugh:
Not sure if this fits the scope of the PR, but should we also also pass `r.previewDataframe` and `r.previewEnvironment` command s to session watcher by default?

renkun-ken:
> Not sure if this fits the scope of the PR, but should we also also pass `r.previewDataframe` and `r.previewEnvironment` commands to session watcher by default?

Yes, I think this PR should address these commands that are introduced before sesssion watcher. It might make sense to check on `r.sessionWatcher` and if it is enabled, then `r.previewDataframe` just sends `View(X)` and `r.previewEnvironment` sends `View(globalenv())` to terminal or otherwise keep the old behavior.

renkun-ken:
`previewEnvironment` and `previewDataframe` are updated so that if session watcher is enabled, then `View(globalenv())` and `View(symbol)` are sent to terminal, respectively; otherwise, fallback to the old behavior.

ManuelHentschel (approved):
LGTM, thanks!

PR #672, opened 2021-06-17
Minor workspace-related changes

### LINKED ISSUE #648: Option to disable workspace clear confirmation prompt
It is very handy that we can assign a keybinding to `r.workspaceViewer.clear`. However, it feels inconvenient to all the time have to answer the confirmation prompt (i.e., see below). I understand the need during analysis, but especially during development, this can feel distracting.

image

I think a boolean setting that enables users to choose whether they want the prompt or not would suffice. For instance `r.workspaceViewer.clearPrompt` or `r.workspaceViewer.clearConfirm`.

ElianHugh:
# What problem did you solve?
1. Prevent the str hover provider from shooting debug console errors when an object is not yet defined in the workspace

> At present, when hovering over a symbol that is undefined in the workspace, vscode shoots the error: `Cannot read property 'str' of undefined: TypeError: Cannot read property 'str' of undefined`. This prevents the error from occurring in the console.

2. **Allow environments with appropriate str to have their contents displayed in the workspace viewer**

> Presently, only list-type objects have their str output displayed in the workspace viewer. However, it is possible for environments to have relevant str outputs that mimic list-type objects. This change brings the workspace viewer in line with the current behaviour of the str hover provider.

image

3. Contribute setting to toggle "clear workspace" prompt (default is to show prompt)

> The prompt for clearing the workspace is helpful for users, but may be unwanted. This provides a setting to toggle whether a confirmation prompt is shown. (prompt enabled by default). Closes #648 

1. Check the console for an error when hovering over an undefined object in the workspace
2. Check if environments with str are displayed in the workspace viewer

```r
str.environment <- function(object, ...) {
    object <- as.list(object)
    invisible(
        NextMethod(
            "str", ...,
            no.list = FALSE,
            nest.lev = 0L
        )
    )
}

a <- new.env()
a$b <- 5
```

3. Check the workspaceViewer.clearPrompt setting is functioning appropriately

### DIFF
--- a/package.json
+++ b/package.json
@@ -1180,6 +1180,11 @@
           "default": false,
           "description": "Remove hidden items when clearing workspace."
         },
+        "r.workspaceViewer.clearPrompt": {
+          "type": "boolean",
+          "default": true,
+          "description": "Prompt the user for confirmation when clearing the workspace."
+        },
         "r.liveShare.timeout": {
           "type": "integer",
           "default": 10000,
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -35,7 +35,10 @@ export class HoverProvider implements vscode.HoverProvider {
         }
         const wordRange = document.getWordRangeAtPosition(position);
         const text = document.getText(wordRange);
-        if (!session.globalenv[text]?.str) {
+        // use juggling check here for both
+        // null and undefined
+        // eslint-disable-next-line eqeqeq
+        if (session.globalenv[text]?.str == null) {
             return null;
         }
         return new vscode.Hover(`\`\`\`\n${session.globalenv[text]?.str}\n\`\`\``);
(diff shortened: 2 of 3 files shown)


renkun-ken:
Please note that an environment might contain delayed assigned symbols or active bindings. `as.list(env)` will force the evaluation of delayed assignments and get the underlying functions from active bindings. I'm not sure if we should deal with these situations, therefore I left them as is without coercing them into a list.

renkun-ken:
Just notice that the changes are made to work with environments with proper `str` output like `list` rather than coercing all environments. Then it is totally fine.

ElianHugh:
Right, it only uses information that is already available to us in the globalenv.json file

renkun-ken (approved):
Everything works well.

LGTM

PR #678, opened 2021-06-20
Add customization options to plot viewer

ManuelHentschel:
Addresses the suggestions made by @psobolewskiPhD in https://github.com/Ikuyadeu/vscode-R/pull/620#issuecomment-846284374.

re 1.: It is now possible to set `r.plot.customStyleOverwrites` to a CSS file containing custom style overwrites. If specified, this file replaces [html/httpgd/styleOverwrites.css](https://github.com/Ikuyadeu/vscode-R/blob/master/html/httpgd/styleOverwrites.css). To check this, save e.g. the following css to a file and set `r.plot.customStyleOverwrites` to its absolute path:
``` css
.httpgd rect {
    stroke: none !important;
    fill: blue !important;
}

svg text {
    font-family: var(--vscode-editor-font-family) !important;
    color: blue !important;
}

.httpgd line, .httpgd polyline, .httpgd polygon, .httpgd path, .httpgd circle, .httpgd rect:not(:first-of-type) {
    stroke: blue !important;
}
```

re 2.: The command `r.plot.togglePreviewPlots` now cycles through mutlirow/scroll/hidden layouts. The setting `r.plot.defaults.plotPreviewLayout` can also be set accordingly.

### DIFF
--- a/html/httpgd/index.ejs
+++ b/html/httpgd/index.ejs
@@ -3,7 +3,7 @@
 
 <head>
     <link href="<%= asWebViewPath('style.css') %> " rel="stylesheet">
-    <link href="<%= asWebViewPath('styleOverwrites.css') %> " rel="stylesheet" class="overwrites" <%- overwriteStyles ? '' : 'disabled' %>>
+    <link href="<%= overwriteCSS %> " rel="stylesheet" class="overwrites" <%- overwriteStyles ? '' : 'disabled' %>>
 </head>
 
 <body>
--- a/html/httpgd/index.ejs
+++ b/html/httpgd/index.ejs
@@ -3,7 +3,7 @@
 
 <head>
     <link href="<%= asWebViewPath('style.css') %> " rel="stylesheet">
-    <link href="<%= overwriteCSS %> " rel="stylesheet" class="overwrites" <%- overwriteStyles ? '' : 'disabled' %>>
+    <link href="<%= overwriteCssPath %> " rel="stylesheet" class="overwrites" <%- overwriteStyles ? '' : 'disabled' %>>
 </head>
 
 <body>
--- a/html/httpgd/index.ejs
+++ b/html/httpgd/index.ejs
@@ -11,7 +11,7 @@
         <%- largePlot?.svg %>
     </div>
     <div id="handler"> </div>
-    <div id="smallPlots" class="<%- useMultirow ? 'multirow' : '' %>">
+    <div id="smallPlots" class="<%- previewPlotLayout%>">
         <% plots.forEach((plot)=> { %>
             <div class="wrapper">
             <%- include(asLocalPath('./smallPlot.ejs'), {plot: plot}) %>
--- a/html/httpgd/index.js
+++ b/html/httpgd/index.js
@@ -65,8 +65,8 @@ window.addEventListener('message', (ev) => {
     else if (msg.message === 'addPlot') {
         addPlot(msg.html);
     }
-    else if (msg.message === 'toggleMultirow') {
-        toggleMultirow(msg.useMultirow);
+    else if (msg.message === 'togglePreviewPlotLayout') {
+        togglePreviewPlotLayout(msg.style);
     }
 });
 function addPlot(html) {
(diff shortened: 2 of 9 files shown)


renkun-ken:
I test it on macOS and the plot viewer tab is always empty and viewing in external browser works.

I observe the following errors in the WebView developer tools:

image

ManuelHentschel:
Thanks for catching that!

renkun-ken:
Thanks for the quick fix. It works well now.

renkun-ken:
I put a css file like the following

```css
.httpgd rect {
    stroke: none !important;
    fill: none !important;
    color: blue;
}

svg text {
    font-family: var(--vscode-editor-font-family) !important;
    stroke: var(--vscode-foreground) !important;
    color: blue;
}

.httpgd line, .httpgd polyline, .httpgd polygon, .httpgd path, .httpgd circle, .httpgd rect:not(:first-of-type) {
    stroke: var(--vscode-foreground) !important;
    color: blue;
}
```

and use its path in `r.plot.customStyleOverwrites` but it seems nothing is changed. How could I check if it actually works?

ManuelHentschel:
For these elements, the colors are specified by the properties `stroke` and `fill`. In order to overwrite the values set inside the svg source, you need to append `!important` (as e.g. in the original styleOverwrites.css). The following example works for me:
``` css
.httpgd rect {
    stroke: none !important;
    fill: blue !important;
}

svg text {
    font-family: var(--vscode-editor-font-family) !important;
    color: blue !important;
}

.httpgd line, .httpgd polyline, .httpgd polygon, .httpgd path, .httpgd circle, .httpgd rect:not(:first-of-type) {
    stroke: blue !important;
}
```

renkun-ken:
Thanks for the example css. It works as designed.

renkun-ken (approved):
LGTM

psobolewskiPhD:
> re 2.: The command `r.plot.togglePreviewPlots` now cycles through mutlirow/scroll/hidden layouts. The setting `r.plot.defaults.plotPreviewLayout` can also be set accordingly.

The css override works perfectly for me. 
I also set this setting, because I never found the thumbnails too useful.
It works, in that they are hidden, but the empty area beneath the plot is still present?
image

I mean with some care, it can be made smaller, but I sort of assumed it would be hidden.

Also, I thought that the paint palette would toggle to the VS Code theme, which matches my OS--dark after dark, light during the day. 
image
I have vscode selected here:
image

I downgraded back to Release and it worked...

ManuelHentschel:
>  It works, in that they are hidden, but the empty area beneath the plot is still present?

That empty area is a feature, not a bug :D My intention was to mimic the scroll-behavior of text editor tabs, where you can scroll past the last line to position the thing you're looking at at the top of the screen. If you want to remove that space, you could set `display: none` for the div with id `"placeholder"` (I haven't tested this)



> Also, I thought that the paint palette would toggle to the VS Code theme, which matches my OS--dark after dark, light during the day.
> [...]
> I have vscode selected here:

That menu is a bit misleading if you're using the CSS overwrite. The file you specify there completely replaces the normal style overwrites, which includes the color theme matching. In order to keep these, you can start by copying https://github.com/Ikuyadeu/vscode-R/blob/master/html/httpgd/styleOverwrites.css and modify that file (as you previously did in the installation directory).

psobolewskiPhD:
Not sure I get the empty space bit, but I'll defer to your expertise 😄 

> That menu is a bit misleading if you're using the CSS overwrite. The file you specify there completely replaces the normal style overwrites, which includes the color theme matching. In order to keep these, you can start by copying https://github.com/Ikuyadeu/vscode-R/blob/master/html/httpgd/styleOverwrites.css and modify that file (as you previously did in the installation directory).

Hmm, I did copy that file and edit it—to get my Fira font preference and eliminate the outlines, etc. as I did before. So for svg text I still have     `stroke: var(--vscode-foreground) !important;`
I'll play around some more, maybe tomorrow.

ManuelHentschel:
> Hmm, I did copy that file and edit it—to get my Fira font preference and eliminate the outlines, etc. as I did before. So for svg text I still have `stroke: var(--vscode-foreground) !important;`
> I'll play around some more, maybe tomorrow.

I just realized that the proper way to color text is using the `fill` property, i.e.
``` css
svg text {
    fill: var(--vscode-foreground) !important;
}
```
in case that makes a difference here (the text looks way better this way than before, I always thought this was due to the dark color theme or somthing -.-). I'll make a new PR with this change, in which I also rename the config entry you mentioned above.

If you're not doing it already, make sure that the new CSS is loaded after you make changes (safest way to do this is to restart vscode, though toggling between styles should work as well). If you still cannot get it to work, please post an example of your CSS file.

If you find some CSS settings that look good for most plots, feel free to share them here. Even in case they're too "disruptive" for the default style, we can still add them as comments, to make it easier for other users to personalize the CSS file.

Issue #671, opened 2021-06-16
plots generated by httpgd do not appear automatically


jooyoungseo:
I was wondering if there would be any convenient way to find generated plot outputs with `httpgd`.

Nothing happens after creating a plot. I have to manually open a browser and go to the url specified within `httpgd::hgd_browse()`.

The following includes my options in `Rprofile`:

``` r
options(vsc.rstudioapi = TRUE)
options(vsc.use_httpgd = TRUE)
```

Of course, `httpgd` has been installed.

renkun-ken:
When `vsc.use_httpgd = TRUE` is used, any plot update should reveal the plot viewer automatically.

Does it work if you execute vscode command "R Plot: Show viewers"?

@ManuelHentschel Any idea on this?

jooyoungseo:
So weird, this issue doesn't occur with v1.6.8. I have found that "R: Luanch RStudio Addin" doesn't work, either with v2.0.0.

In sum, the newest version v2.0.0 seems to have some issues on Windows.

renkun-ken:
The major change from v1.6.8 to v2.0.0 is made in #626. Everything works well under Ubuntu and macOS. @ElianHugh It looks like you are using Windows. Could you reproduce the problem here?

ElianHugh:
I'm on Linux until tomorrow night, but haven't experienced any of the problems outlined here. I don't see anything from the 2.0 release that would cause this to happen, particularly when the rstudioapi was not touched at all

renkun-ken:
@jooyoungseo Could you go to "Help -> Toggle Developer Tools" and choose "Console" tab when you start an R session. If the extension is successfully activated and the session attached, the following log is expected:

```
[Extension Host] Start language server for file:///home/renkun/Workspaces/testR/httpgd.R
console.ts:137 [Extension Host] R binary: /usr/bin/R
console.ts:137 [Extension Host] LANG: en_US.UTF-8
console.ts:137 [Extension Host] R process connected
console.ts:137 [Extension Host] activating extension
console.ts:137 [Extension Host] starting language server
console.ts:137 [Extension Host] [LiveSessionListener] started
console.ts:137 [Extension Host] Initialize session watcher
console.ts:137 [Extension Host] [deploySessionWatcher] extensionPath: /home/renkun/.vscode/extensions/ikuyadeu.r-2.0.0
console.ts:137 [Extension Host] [deploySessionWatcher] watcherDir: /home/renkun/.vscode-R
[...]

PR #679, opened 2021-06-20
Catch LiveShare API errors

ElianHugh:
Primary change:
- Adds try/catch to the accessing of the vsls API, so as to hopefully avoid situations such as #671

Minor changes:
- Remove some superflous/dead code from previous iterations of the LiveShare PR
- Change the guest dataview file to have the same path as the host

Tested on Arch Linux


renkun-ken:
I test on macOS with a corrupted live share installation (some extension files are removed deliberately) and vscode-R is activated successfully. When live share API is ok, the features work well too.

renkun-ken (approved):
LGTM

ElianHugh:
Thank you for checking!

PR #681, opened 2021-06-21
Small Plot Viewer adjustments

ManuelHentschel:
Addresses some issues that came up in the conversation of #678:
* Fixes the CSS used to change font colors
* Clarifies the description of two config entries

Also:
* Removes some unnecessary logging output
* Makes resizing the plot more "snappy" by not requesting an updated plot, while the previous request is still being handled


renkun-ken (approved):
LGTM. Works quite well.

psobolewskiPhD:
With the snappier resizing I don't mind that empty area.

PR #684, opened 2021-06-27
Provide links to workspace files in terminal

### LINKED ISSUE #512: Create terminal link provider for {testthat} errors / warnings
When showing test reports like:

```
✔ |   7       | test data.table paste [0.3 s]
✖ |   0 1     | test RStudio preference API calls
─────────────────────────────────────────────────────────
test_rstudio.R:10: failure: readPreference() returns an integer
rstudioapi::readPreference("num_spaces_for_tab", 4) has type `double`, not `integer`.
─────────────────────────────────────────────────────────
✔ |  28       | test tribble paste [0.6 s]
✔ |  16       | test vector paste [0.3 s]

```

It would be convenient if the `test_rstudio.R:10` text was a clickable link to that file and line number. I am not sure if RStudio has this feature, but it is implemented in Emacs/ESS.

I have not researched it thoroughly yet, but it seems like the VSCode [TerminalLinkProvider API](https://code.visualstudio.com/api/references/vscode-api#TerminalLinkProvider) is designed for this purpose. I found some example code [here](https://code.visualstudio.com/updates/v1_49).

ElianHugh:
# What problem did you solve?

Closes #512 

E.g., provides terminal links for the following:
- text.r
- text.r:5
- text.r:5:6

You can technically test this by typing the file name in the terminal, and alt+clicking the link.

## Screenshot

image

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -305,7 +305,7 @@ export async function sendRangeToRepl(rng: vscode.Range): Promise<void> {
     editor.selections = sel0;
 }
 
-export class rTerminalLinkProvider implements vscode.TerminalLinkProvider {
+export class TerminalLinkProvider implements vscode.TerminalLinkProvider {
     provideTerminalLinks(context: vscode.TerminalLinkContext, token: vscode.CancellationToken): vscode.ProviderResult<vscode.TerminalLink[]> {
         const matches: vscode.TerminalLink[] = [];
         const regex = /[-_a-zA-Z]+\.r(:[0-9]+)?(:[0-9]+)?/gi;
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -331,11 +331,12 @@ export class rTerminalLinkProvider implements vscode.TerminalLinkProvider {
         return matches;
     }
     async handleTerminalLink(link: unknown): Promise<void> {
-        if (link['data']) {
-            const url = (link['data'] as string).replace(/:[0-9]+/g, '');
-            const range = (link['data'] as string).match(/(?<=:)[0-9]+/g);
+        const data: unknown = link['data'];
+        if (typeof data === 'string') {
+            const url = data.replace(/:[0-9]+/g, '');
             const loc = await vscode.workspace.findFiles(`**/${url}`, null, 1);
-            if (loc) {
+            if (loc?.length > 0) {
+                const range = (data.match(/(?<=:)[0-9]+/g))?.map(num => Number(num) - 1);
                 await vscode.commands.executeCommand(
                     'vscode.open',
                     vscode.Uri.file(loc[0].fsPath),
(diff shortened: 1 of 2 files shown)


src/rTerminal.ts
```diff
@@ -304,3 +304,64 @@ export async function sendRangeToRepl(rng: vscode.Range): Promise<void> {
     await vscode.commands.executeCommand('editor.debug.action.selectionToRepl');
     editor.selections = sel0;
 }
+
+export class TerminalLinkProvider implements vscode.TerminalLinkProvider {
+    provideTerminalLinks(context: vscode.TerminalLinkContext, token: vscode.CancellationToken): vscode.ProviderResult<vscode.TerminalLink[]> {
+        const matches: vscode.TerminalLink[] = [];
+        const regex = /[-_a-zA-Z]+\.r(:[0-9]+)?(:[0-9]+)?/gi;
```
renkun-ken, line 311: Is `/[-_a-zA-Z]+\.[Rr](:[0-9]+)?(:[0-9]+)?/g` clear enough so that there's no need to use ignore-case here?

renkun-ken: Also, what about the case where the file name contains spaces?

ElianHugh: Thank you for the review! The regex change works nicely, thank you.

The space one is tricky I think. I had the link handler handle searching for files so that the link provider was more performant, but that does mean its hard to tell if a file actually exists...

If we did it purely to *testthat* specifications, we could check purely between brackets, e.g. the brackets in:

```
Error: (test function.r:2:5): functions work properly
```

Alternatively, we could have a list of the workspace files, and parse the text against the list?

renkun-ken: I wonder if this terminal link provider overrides the default handler where full paths like `/home/ken/test.R:10:5` could be already handled, i.e. if we only handle `(file.R:line:col)` case, will the default cases still work?

renkun-ken: >If we did it purely to testthat specifications, we could check purely between brackets, e.g. the brackets in:
I know the following packages that could produce file links on some condition or error:

* testthat
* lintr
* Rcpp
* cpp11
* tinytest

It seems only testthat uses `(file:line:col)` notation. If we limit the regex to only capture this, it would make the terminal link provider much less useful.

ManuelHentschel:
> E.g., provides terminal links for the following:
> 
>     * text.r
> 
>     * text.r:5
> 
>     * text.r:5:6
> 
> 
> You can technically test this by typing the file name in the terminal, and alt+clicking the link.

For me, these examples all work without any additional terminal link provider, resulting in the same behaviour as pressing `ctrl+p` and typing the file name. Are there examples where this does not work?

renkun-ken:
So far, the default behavior does not work with testthat error and Rcpp error output where only relative paths are provided.

ManuelHentschel:
> 
> 
> So far, the default behavior does not work with testthat error and Rcpp error output where only relative paths are provided.

I'm not sure if I'm missing something, but e.g. in the following output
```
Error (test-transformation_functions.R:251:3): Gamma2chi_3D works
Error: could not find function "Gamma2chi_3D"
Backtrace:
```
I can hold `ctrl`, then click `test-transformation_functions.R:251:3`, then confirm the filename with `Enter` and it opens the correct file and puts the cursor at the correct position.

renkun-ken:
@ManuelHentschel Thanks for pointing out this. On Ubuntu, I could hold `alt` with the file link and then possible files are listed in the picker. Looks like everything already works nicely now, we just need to document this so that user knows about this amazing feature already built-in in VS Code.

renkun-ken:
@ElianHugh Looks like this PR is not needed any more.

ElianHugh:
Wow! I could have sworn this wasn't working when I last checked.... Thank you for picking up on that, very cool feature!

renkun-ken:
On macOS, it is `option` + click.

Issue #103, opened 2019-03-23
Preview of Rmarkdown


Ikuyadeu:
Currently, the Markdown file can get a preview.
I believe R markdown file can also get a preview by calling preview command on Markdown.
I put this issue as the TODO

andycraig:
By adding the below to `settings.json`, you can use `Markdown: Open Preview` on R markdown files. It doesn't do anything special with the R code blocks, though.

    "files.associations": {
         "*.Rmd": "markdown"
    }

From https://qiita.com/yutannihilation/items/1f71c0fedae6e4b8038c

Ikuyadeu:
@andycraig Thank you! It is working well.

Ikuyadeu:
I can not use R markdown syntax and inline execution by this change.
We should call this as the R markdown mode.

andycraig:
That's right, that change doesn't do anything special with R code blocks.

A better R markdown experience could be gained by adding a command that calls the `knitr` package, and then previews a resulting HTML or PDF file (maybe using https://marketplace.visualstudio.com/items?itemName=hdg.live-html-previewer and https://marketplace.visualstudio.com/items?itemName=tomoki1207.pdf). This could be a Good First Issue.

RStudio-like inline output from R code blocks would be tougher. I'm not sure if the VSCode editor supports non-text objects.

andycraig:
It looks like the neuron extension might be very helpful here:

https://marketplace.visualstudio.com/items?itemName=neuron.neuron-IPE

I haven't managed to get it running successfully yet.

andycraig:
Using VSNotebooks:

![notebook](https://user-images.githubusercontent.com/23294156/56659225-852f9280-66d7-11e9-8c55-5d20e25587ba.gif)


This setting was necessary to get VSNotebooks to recognise Rmd files:

```
"files.associations": {
    "*.Rmd": "r"
}
```

rjBadger:
Is this still being worked on? Rmarkdown support is one of the major reasons I don't switch to vscode.

andycraig:
Hi @rjBadger, as shown in the video above, the VSNotebooksBeta extension (a fork of the neuron extension) provides some support of R Markdown files, although there is a lot of room for improvement. I'd like to see work on R Markdown support for VSCode made on neuron or one of its forks, so that the improvements also benefit Julia etc. I'm not currently working on anything related to this, though.

andycraig:
[...]

PR #692, opened 2021-07-01
Preview R Markdown files via background process

ElianHugh:
# What problem did you solve?
Improves previewing rmd files, as per some discussion in #103. Not quite the same experience as vscode's markdown preview, but is a little less cumbersome than running rmarkdown::run manually.

~~The preview "engine" is controlled through a configuration setting, and currently supports `rmarkdown::run` or `xaringan::infinite_moon_reader`.~~

Some stuff that I think would be neat to have at some point:
1. scroll sync with rmd source 
~~2. get the document's http content (similar to how the help provider functions), meaning that the iframe isn't needed~~

## Screenshot
image

### DIFF
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -71,7 +71,7 @@ class RMarkdownPreview extends vscode.Disposable {
         const isHtml = !!re.exec(content);
 
         if (!isHtml) {
-            const html = escapeHtml(content);
+            const html: string = escapeHtml(content);
             content = `<html><head></head><body><pre>${html}</pre></body></html>`;
         }
 
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -71,7 +71,7 @@ class RMarkdownPreview extends vscode.Disposable {
         const isHtml = !!re.exec(content);
 
         if (!isHtml) {
-            const html: string = escapeHtml(content);
+            const html = escapeHtml(content);
             content = `<html><head></head><body><pre>${html}</pre></body></html>`;
         }
 
@@ -299,19 +299,23 @@ export class RMarkdownPreviewManager {
                 (data: Buffer) => {
                     const dat = data.toString('utf8');
                     this.rMarkdownOutput.appendLine(dat);
-                    const outputUrl = re.exec(dat)?.[0]?.replace(re, '$1');
-                    if (outputUrl) {
-                        if (viewer !== undefined) {
-                            void this.openPreview(
-                                vscode.Uri.parse(outputUrl),
-                                fileUri,
-                                fileName,
-                                childProcess,
-                                viewer,
-                                currentViewColumn
-                            );
-                        }
+                    if (token?.isCancellationRequested) {
                         resolve(childProcess);
+                    } else {
+                        const outputUrl = re.exec(dat)?.[0]?.replace(re, '$1');
+                        if (outputUrl) {
+                            if (viewer !== undefined) {
+                                void this.openPreview(
+                                    vscode.Uri.parse(outputUrl),
+                                    fileUri,
+                                    fileName,
+                                    childProcess,
+                                    viewer,
+                                    currentViewColumn
+                                );
+                            }
+                            resolve(childProcess);
+                        }
                     }
                 }
             );
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -1,89 +1,142 @@
 
 import * as cp from 'child_process';
-import * as kill from 'tree-kill';
 import * as vscode from 'vscode';
+
 import { getBrowserHtml } from '../session';
 import { closeBrowser, isHost, shareBrowser } from '../liveshare';
 import { config, doWithProgress, getRpath,  setContext } from '../util';
+import { extensionContext } from '../extension';
 
-interface IPreviewProcess {
-    cp: cp.ChildProcessWithoutNullStreams,
-    file: string,
-    panel: vscode.WebviewPanel,
+class RMarkdownChild extends vscode.Disposable {
+    title: string;
+    cp: cp.ChildProcessWithoutNullStreams;
+    panel: vscode.WebviewPanel;
+    resourceViewColumn: vscode.ViewColumn;
+    uri: vscode.Uri;
+    externalUri: vscode.Uri;
+
+    constructor(title: string, cp: cp.ChildProcessWithoutNullStreams, panel: vscode.WebviewPanel, resourceViewColumn: vscode.ViewColumn, uri: vscode.Uri, externalUri: vscode.Uri) {
+        super(() => {
+            this.cp.kill('SIGKILL');
+            this.panel?.dispose();
+        });
+
+        this.title = title;
+        this.cp = cp;
+        this.panel = panel;
+        this.resourceViewColumn = resourceViewColumn;
+        this.uri = uri;
+        this.externalUri = externalUri;
+    }
+}
+
+class RMarkdownChildStore extends vscode.Disposable {
+    private store: Set<RMarkdownChild> = new Set<RMarkdownChild>();
+
+    constructor() {
+        super((): void => {
+            for (const child of this.store) {
+                child.dispose();
+            }
+            this.store.clear();
+        });
+    }
+
+    public add(child: RMarkdownChild): void {
+        this.store.add(child);
+    }
+
+    public delete(child: RMarkdownChild): void {
+        child.dispose();
+        this.store.delete(child);
+    }
+
+    public get(uri: vscode.Uri) {
+        for (const child of this.store) {
+            if (child.uri === uri) {
+                return child;
+            }
+        }
+        return undefined;
+    }
+
+    public has(uri: vscode.Uri) {
+        for (const child of this.store) {
+            if (child.uri === uri) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    [Symbol.iterator](): Iterator<RMarkdownChild> {
+        return this.store[Symbol.iterator]();
+    }
 }
 
 export class RMarkdownPreviewManager {
-    private openProcesses: IPreviewProcess[] = [];
-    private activePreview: vscode.WebviewPanel;
-    private activeResource: vscode.Uri;
-    private activeExternalResource: vscode.Uri;
     private rPath: string;
+    private ChildStore: RMarkdownChildStore = new RMarkdownChildStore;
+    private activePreview: RMarkdownChild;
+    private rMarkdownOutput: vscode.OutputChannel = vscode.window.createOutputChannel('R Markdown');
 
     public async init(): Promise<void> {
         this.rPath = await getRpath(false);
+        extensionContext.subscriptions.push(this.ChildStore);
     }
 
     public async previewRmd(viewer: vscode.ViewColumn, uri?: vscode.Uri): Promise<void> {
         const fileUri = uri ?? vscode.window.activeTextEditor.document.uri;
         const fileName = fileUri.path.substring(fileUri.path.lastIndexOf('/') + 1);
         const previewEngine: string = config().get('rmarkdown.previewEngine');
+        const currentViewColumn: vscode.ViewColumn = vscode.window.activeTextEditor.viewColumn ?? vscode.ViewColumn.Active;
         const cmd = (
             `${this.rPath} --silent --slave --no-save --no-restore -e "${previewEngine}('${fileUri.path}')"`
         );
         const reg: RegExp = this.constructRegex(previewEngine);
-        let call = undefined;
 
-
-        if (this.openProcesses.some(e => e.file === fileName)) {
-            this.openProcesses.filter(e => e.file === fileName)[0].panel.reveal();
+        if (this.ChildStore.has(fileUri)) {
+            this.ChildStore.get(fileUri).panel.reveal();
         } else {
-            await doWithProgress(() => {
-                try {
-                    call = cp.spawn(cmd, null, { shell: true });
-                } catch (e) {
-                    console.warn((e as string));
-                }
-                (call as cp.ChildProcessWithoutNullStreams).stderr.on('data',
-                    (data: Buffer) => {
-                        const dat = data.toString('utf8');
-                        const match = reg.exec(dat)?.[0];
-                        const previewUrl = this.constructUrl(previewEngine, match, fileName);
-                        if (match) {
-                            void this.showPreview(previewUrl, fileName, call, viewer, fileUri);
-                        }
-                    });
+            await doWithProgress(async () => {
+                await this.spawnProcess(cmd, reg, previewEngine, fileName, viewer, fileUri, currentViewColumn)
+                    .catch((cp: cp.ChildProcessWithoutNullStreams) => {
+                        void vscode.window.showErrorMessage('There was an error in knitting the document. Please check the R Markdown output stream.');
+                        cp.kill('SIGKILL');
+                    }
+                    );
             },
                 vscode.ProgressLocation.Notification,
-                `Rendering ${fileName}...`
+                `Knitting ${fileName}...`
             );
         }
     }
 
     public refreshPanel(): void {
         if (this.activePreview) {
-            this.activePreview.webview.html = '';
-            this.activePreview.webview.html = getBrowserHtml(this.activeExternalResource);
+            this.activePreview.panel.webview.html = '';
+            this.activePreview.panel.webview.html = getBrowserHtml(this.activePreview.uri);
         }
     }
 
     public async showSource(): Promise<void> {
-        if (this.activeResource) {
-            const viewCol = vscode.window.visibleTextEditors.filter(e => e.document.uri === this.activeResource)[0]?.viewColumn;
-            await vscode.commands.executeCommand('vscode.open', this.activeResource, {
+        if (this.activePreview) {
+            // to fix, kind of buggy
+            await vscode.commands.executeCommand('vscode.open', this.activePreview.uri, {
                 preserveFocus: false,
                 preview: false,
-                viewColumn: viewCol ?? vscode.ViewColumn.Active
+                viewColumn: this.activePreview.resourceViewColumn ?? this.activePreview.panel.viewColumn ?? vscode.ViewColumn.Active
             });
         }
     }
 
     public openExternalBrowser(): void {
-        if (this.activeExternalResource) {
-            void vscode.env.openExternal(this.activeExternalResource);
+        if (this.activePreview) {
+            void vscode.env.openExternal(this.activePreview.externalUri);
         }
     }
 
-    private async showPreview(url: string, title: string, cp: cp.ChildProcessWithoutNullStreams, viewer: vscode.ViewColumn, fileUri: vscode.Uri): Promise<void> {
+    private async showPreview(url: string, title: string, cp: cp.ChildProcessWithoutNullStreams, viewer: vscode.ViewColumn, fileUri: vscode.Uri, resourceViewColumn: vscode.ViewColumn): Promise<void> {
         // construct webview and its related html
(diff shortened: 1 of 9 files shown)


renkun-ken:
I wonder if it makes sense to create a `R Markdown` output channel to show the stdout and stderr of the background R progress?

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,179 @@
...
+    public refreshPanel(): void {
+        if (this.activePreview) {
+            this.activePreview.webview.html = '';
+            this.activePreview.webview.html = getBrowserHtml(this.activeExternalResource);
+        }
+    }
+
+    public async showSource(): Promise<void> {
+        if (this.activeResource) {
+            const viewCol = vscode.window.visibleTextEditors.filter(e => e.document.uri === this.activeResource)[0]?.viewColumn;
```
renkun-ken, line 71: This seems not working as expected for me. When the source is on the first view column but not selected and the preview on the second view column, then `showSource()` will open a new tab in the second view column (seems to fallback to `Active`) so that the original source editor in the first view column is not revealed.

ElianHugh: I could replicate this, but only if I did the following:

1. Open rmd file
2. Preview doc
3. Close rmd file
4. Reopen rmd file
5. Go to source

I'll dig into this a bit

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,267 @@
...
+                    }
+                    );
+            },
+                vscode.ProgressLocation.Notification,
+                `Knitting ${fileName}...`
+            );
+        }
+    }
+
+    public refreshPanel(): void {
```
renkun-ken, line 115: This does not work for me anymore. Whenever I click "Refresh Preview", the WebView goes to blank.

renkun-ken: From the WebView developer tool, it looks like the iframe has the wrong url:
```html
<iframe src="file:///Users/ken/Workspaces/test/rtest/test.Rmd" width="100%" height="100%" frameborder="0">
</body>
</html>
</iframe>
```

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,267 @@
...
+        if (this.activePreview) {
+            this.activePreview.panel.webview.html = '';
+            this.activePreview.panel.webview.html = getBrowserHtml(this.activePreview.uri);
```
renkun-ken, line 118: Here it should be `this.activePreview.externalUri`.

src/rmarkdown/preview.ts
renkun-ken: https://github.com/REditorSupport/vscode-R/pull/692/files#r663525890 should resolve this.

ElianHugh: Thanks for picking up on this!

src/util.ts
```diff
@@ -230,11 +230,12 @@ export async function executeAsTask(name: string, command: string, args?: string
 // executes a callback and shows a 'busy' progress bar during the execution
 // synchronous callbacks are converted to async to properly render the progress bar
 // default location is in the help pages tree view
-export async function doWithProgress<T>(cb: () => T | Promise<T>, location: string | vscode.ProgressLocation = 'rHelpPages'): Promise<T> {
+export async function doWithProgress<T>(cb: () => T | Promise<T>, location: string | vscode.ProgressLocation = 'rHelpPages', title?: string): Promise<T> {
     const location2 = (typeof location === 'string' ? { viewId: location } : location);
     const options: vscode.ProgressOptions = {
         location: location2,
-        cancellable: false
+        cancellable: false,
```
renkun-ken, line 237: Is it possible to make it cancellable by killing the spawned process?

If a Rmd document contains time-consuming code, it might take a long time to render. A cancellable progress might make sense in this case.

ElianHugh: Good idea, just pushed an update to implement this

src/rmarkdown/preview.ts
```diff
@@ -96,24 +99,38 @@ export class RMarkdownPreviewManager {
...
-                        void vscode.window.showErrorMessage('There was an error in knitting the document. Please check the R Markdown output stream.');
+            this.busyUri.push(fileUri);
+            await doWithProgress(async (token: vscode.CancellationToken) => {
+                await this.spawnProcess(cmd, reg, previewEngine, fileName, viewer, fileUri, currentViewColumn, token)
+                    .catch((rejection: {
+                        cp: cp.ChildProcessWithoutNullStreams,
+                        wasCancelled?: boolean
+                    }) => {
+                        if (!rejection.wasCancelled) {
+                            void vscode.window.showErrorMessage('There was an error in knitting the document. Please check the R Markdown output stream.');
```
renkun-ken, line 115: Does it make sense to reveal `rMarkdownOutput` to show the output messages?

src/util.ts
renkun-ken: I test with a computing heavy Rmd and it looks like the progress only shows quite shortly during the spawning rather than rendering the Rmd. The shiny app already displays a "Loading" during the rendering. Then the cancel button is only useful if the spawning takes a long time.

ElianHugh: Right. I guess it is useful if people use the `xaringan::infinite_moon_runner`, as this seems to actually knit the documents, as opposed to using a shiny backend.

ElianHugh: Having said that, I'm not 100% sold on having two preview 'engines'. I put it in initially as `infinite_moon_runner` seemed to not run into the issues of `rmarkdown::run`, such as the loss of text decoration.

renkun-ken:
It looks like the shiny app started by `rmarkdown::run()` does copy the web resources to local (see https://github.com/rstudio/rmarkdown/blob/master/R/shiny.R#L264) and the urls are replaced to local urls whenever an auto refresh occurs. This violates the security policy of the WebView on accessing local resources.

renkun-ken:
Another approach to avoid this is to only use "rmarkdown::render()" or any other engine that is able to produce an HTML file. And we could use `fs.watch()` to watch the file and implement auto-render on file changes. But this will add much complexity.

ElianHugh:
I suppose using `rmarkdown::render` would also ensure that syntax highlighting is maintained, and that we would have more control over the document's styling. I wonder if there is much of a difference performance and compilation wise

renkun-ken:
> I suppose using rmarkdown::render would also ensure that syntax highlighting is maintained, and that we would have more control over the document's styling.

Exactly. Both mathjax and syntax highlighting should work after reload.

> I wonder if there is much of a difference performance and compilation wise

I think there should not be a significant performance difference compared with rendering from a shiny app.

ElianHugh:
I've added a super messy implementation of the document file watcher, which appears to function as expected. Will be cleaned up shortly

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,328 @@
...
+        this.busyUriStore.delete(fileUri);
+    }
+
+    private async spawnProcess(fileUri: vscode.Uri, fileName: string, token?: vscode.CancellationToken, viewer?: vscode.ViewColumn, currentViewColumn?: vscode.ViewColumn) {
+        return await new Promise<cp.ChildProcessWithoutNullStreams>((resolve, reject) => {
+            const lim = '---vsc---';
+            const re = new RegExp(`.*${lim}(.*)${lim}.*`, 'ms');
+            const cmd = (
+                `${this.rPath} --silent --slave --no-save --no-restore -e ` +
+                `"cat('${lim}', rmarkdown::render('${String(fileUri.path)}', output_dir = '${this.watcherDir}'), '${lim}', sep=''); while(TRUE) Sys.sleep(1)" `
```
renkun-ken, line 194: Using `~/.vscode-R` to store the output files might not be a good idea. If two Rmd files have the same name but in different folders, they will have the same output file.

I suggest using system temp dir (`os.tmpdir()`) instead. Since the R session will be killed each time before update, the R session `tempdir()` does not work with the next R session. Therefore, we might need to create a temp dir for vscode-R and each file has a unique output file (a random file name or a `fileUri` hash or something). Also, we might also better cleanup the temp files we created on some point (e.g. once the file is read?)

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,328 @@
...
+                fsTimeout = setTimeout(() => { fsTimeout = null; }, 1000);
+                void RMarkdownPreviewManager.updatePreview(this);
+            }
+        });
+        this.fileWatcher = docWatcher;
+    }
+}
+
+class RMarkdownChildStore extends vscode.Disposable {
+    private store: Set<RMarkdownChild> = new Set<RMarkdownChild>();
```
renkun-ken, line 51: Is there a reason we don't use a `Map` here?

src/rmarkdown/preview.ts
renkun-ken: Just notice that the output file will be removed.

Also, just remind that we use `~/.vscode-R` instead of `/tmp/vscode-R` to avoid user clash on the same directory.

renkun-ken:
I move `watcherDir` from `session.ts` to `extDir` in `extension.ts` as a global extension directory in user home. And temporary files are generated into `~/.vscode-R/tmp` as `tmpDir`. Now session watcher uses `extDir` and rmarkdown preview uses `tmpDir`.

src/rmarkdown/preview.ts
ElianHugh: Not particularly - were you thinking something like `Map<vscode.Uri, RMarkdownChild>();`, where the uri is the input path?

ElianHugh:
> I move `watcherDir` from `session.ts` to `extDir` in `extension.ts` as a global extension directory in user home. And temporary files are generated into `~/.vscode-R/tmp` as `tmpDir`. Now session watcher uses `extDir` and rmarkdown preview uses `tmpDir`.

Thanks! Works quite nicely - I've also removed references to `dir` because it will just end up being `tmpDir` anyway

src/rmarkdown/preview.ts
renkun-ken: Yes, exactly.

ElianHugh:
Latest push now has a 'toggle style' button similar to the plot viewer:

image

Refactor from set to map is ongoing - RMarkdownChild will have the input uri removed soon as its now superfluous

renkun-ken:
The dark theme does not work for me:

image

renkun-ken:
Maybe we could consider using the way of style tweaking in help pages:

https://github.com/REditorSupport/vscode-R/blob/ffe7ef58a264788c3d0efb48e1d6f861ad853f5c/src/helpViewer/index.ts#L500-L518

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,357 @@
...
+import * as cp from 'child_process';
+import * as vscode from 'vscode';
+import * as fs from 'fs-extra';
+
+import { config, doWithProgress, getRpath, readContent, setContext } from '../util';
+import { extensionContext, tmpDir } from '../extension';
+import path = require('path');
+import crypto = require('crypto');
+
+class RMarkdownChild extends vscode.Disposable {
```
renkun-ken, line 11: Does it make sense to rename `RMarkdownChild` to `RMarkdownPreview` and `RMarkdownChildStore` to `RMarkdownPreviewStore`.

It seems to look more natural to have `RMarkdownPreviewManager` manage a collection of `RMarkdownPreview` stored in `RMarkdownPreviewStore`.

ElianHugh:
Just noticed that syntax highlighting is seemingly lost, but only on my windows computer.

src/rmarkdown/preview.ts
```diff
@@ -147,22 +202,39 @@ export class RMarkdownPreviewManager {
...
                 preview: false,
-                viewColumn: this.activePreview.resourceViewColumn ?? this.activePreview.panel.viewColumn ?? vscode.ViewColumn.Active
+                viewColumn: this.activePreview?.preview?.resourceViewColumn ?? this.activePreview?.preview?.panel.viewColumn ?? vscode.ViewColumn.Active
             });
         }
     }
 
     public openExternalBrowser(): void {
         if (this.activePreview) {
             void vscode.env.openExternal(
```
renkun-ken, line 215: Opening a local html file end up as a new tab in vscode here. Not sure if there is an easy way to open it in default browser.

renkun-ken:
The dark theme looks good, but the output is almost invisible under many dark themes.

image

src/rmarkdown/preview.ts
ElianHugh: https://github.com/sindresorhus/open seems to get around this issue, and works as expected when replacing openExternal. I guess we could either go with that, or just remove openExternal

renkun-ken: We could definitely use it.

ElianHugh: Holding off using it just for the moment - I haven't found a good way of getting the default browser for file opening. It works if we hardcode it, but obviously not ideal to do so

ElianHugh:
> 
> 
> The dark theme looks good, but the output is almost invisible under many dark themes.
> 
> image

Don't see the problem here 😅 

Should be fixed following last push!

renkun-ken:
The theming looks good now. And `vscode.env.openExternal(this.activePreview?.preview?.outputUri)` works well on Ubuntu. It we use an open-in-browser package and then it won't work in remote development scenario where a http server must be created to serve the html page. In this case, we need to rely on something like [Live Preview](https://marketplace.visualstudio.com/items?itemName=ms-vscode.live-server).

src/rmarkdown/preview.ts
renkun-ken: I revert to using `vscode.env.openExternal` at the moment. To make it work on both local and remote, user might need [Live Preview](https://marketplace.visualstudio.com/items?itemName=ms-vscode.live-server) to serve the HTML. Let's keep it simeple first.

ElianHugh:
Makes sense to me! I had a quick look into scroll syncing and it gave me a headache. I think if we are happy with the current features then we can save that for a later date

renkun-ken (approved):
LGTM

renkun-ken:
There's one more thing? I think we need to add a setting to control whether to auto-render the document on file update. Real world documents could be long and heavy, we could make such a setting to control the default behavior and also put a button on the toolbar to toggle the behavior.

renkun-ken:
"Enable/Disable Auto Refresh" commands and `r.rmarkdown.preview.autoRefresh` are added.

image

User could control whether to auto-refresh for each preview now.

ElianHugh:
Awesome, thanks for the great work! I wonder if it's worth thinking about forcing [granular caching](https://yihui.org/knitr/demo/cache/), particularly one of the lower settings (i.e. `cache = 1` or `cache = 2`).

renkun-ken:
> Awesome, thanks for the great work! I wonder if it's worth thinking about forcing or having an option for [granular caching](https://yihui.org/knitr/demo/cache/), particularly one of the lower settings (i.e. `cache = 1` or `cache = 2`).

Should the rmd author control the cache level or should we have control on it?

ElianHugh:
Hmm, after some testing, it probably shouldn't be forced, even in a document preview. There are a few situations which can be confusing if caching is forced globally:

```{r, cache = 1}
b <- 5
```

```{r}
b + 5
```

Even if chunk 1 is changed later, chunk 2 will print 10.

renkun-ken:
@ElianHugh Do you have anything to add? If not, I'll merge this PR.

ElianHugh:
No - looks good to me!

ManuelHentschel:
The command `R: Open Preview` isn't working on windows. As far as I can tell, this is due to using the unqoted path [here](https://github.com/ManuelHentschel/vscode-R/blob/950bfae89a2541c6485631f77175a36c17759a5b/src/rmarkdown/preview.ts#L178) (which doesn't work for paths containing spaces), the linebreaks [here](https://github.com/ManuelHentschel/vscode-R/blob/950bfae89a2541c6485631f77175a36c17759a5b/src/rmarkdown/preview.ts#L305-L307), and the backslashes in the paths [here](https://github.com/ManuelHentschel/vscode-R/blob/950bfae89a2541c6485631f77175a36c17759a5b/src/rmarkdown/preview.ts#L306).

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,416 @@
...
+    autoRefresh: boolean;
+
+    constructor(title: string, cp: cp.ChildProcessWithoutNullStreams, panel: vscode.WebviewPanel,
+        resourceViewColumn: vscode.ViewColumn, outputUri: vscode.Uri, uri: vscode.Uri,
+        RMarkdownPreviewManager: RMarkdownPreviewManager, themeBool: boolean, autoRefresh: boolean) {
+        super(() => {
+            this.cp?.kill('SIGKILL');
+            this.panel?.dispose();
+            this.fileWatcher?.close();
+            fs.removeSync(this.outputUri.path);
```
ManuelHentschel, line 32: Also applies to some other places: `vscode.Uri` has a `fsPath` property which is more robust regarding windows drive letters, backslashes etc.

src/rmarkdown/preview.ts
```diff
@@ -0,0 +1,416 @@
...
+        this.cp = cp;
+        this.panel = panel;
+        this.resourceViewColumn = resourceViewColumn;
+        this.outputUri = outputUri;
+        this.autoRefresh = autoRefresh;
+        void this.refreshContent(themeBool);
+        this.startFileWatcher(RMarkdownPreviewManager, uri);
+    }
+
+    public styleHtml(themeBool: boolean) {
```
ManuelHentschel, line 45: E.g. `useDarkTheme` is more descriptive than `themeBool`

DonJayamanne:
@renkun-ken @ElianHugh Apologies if this is a very silly question.
Can we not try to use the VS Codes Notebook UI to render markdown files (instead of vscodes markdown preview).
Its my understanding that this current implementation just displays the rmd file as HTML.
Questions:
* Does this impelementation execute cells or just display the markdown content & code in the HTML preview?
If this is the case, would the new VS Code Notebook interface been beneficial? (we can have a readonly interface).
The benefit is it will be more resource intensive & significantly performant.

ElianHugh:
> 
> 
> @renkun-ken @ElianHugh Apologies if this is a very silly question.
> Can we not try to use the VS Codes Notebook UI to render markdown files (instead of vscodes markdown preview).
> Its my understanding that this current implementation just displays the rmd file as HTML.
> 
>     * Does this current implementation execute cells or just display the markdown content & code in the HTML preview?
>       If this is the case, would the new VS Code Notebook interface been beneficial? (we can have a readonly interface).
>       The benefit is it will be more resource intensive & significantly performant.

Hi @DonJayamanne, not a silly question at all :)

The current preview implementation executes all code cells, so that the output html file is essentially what the final .rmd output would be. 

E.g., here I have a .rmd file on the left, and the preview on the right, executed parts of the cells in poorly scribbled red:

image

DonJayamanne:
> The current preview implementation executes all code cells, so that the output html file is essentially what the final .rmd

Awesome.
Nothing to change here. I was hoping we weren't executing the code, but looks like we are doing both, executing & generating HTML output in one step (using the cli).
More info about the notebook ui can be found here https://code.visualstudio.com/api/extension-guides/notebook

renkun-ken:
> @renkun-ken @ElianHugh Apologies if this is a very silly question.
> Can we not try to use the VS Codes Notebook UI to render markdown files (instead of vscodes markdown preview).
> Its my understanding that this current implementation just displays the rmd file as HTML.

I don't quite understand the question. The rendering is all done by `rmarkdown::render("file.Rmd")`, we just call it in background the present the resulted html file in a WebView.

DonJayamanne:
> This is a just equivalent of RStudio feature.

Thats fine.
My assumption was that we weren't executing the code and we were using this tool to generate HTML to simply preview the RMD with math renderer (markdown) correctly & felt the VS Codes Notebook UI would achieve that.

But that's not the case, hence we can consider my question resolved. Thanks

Issue #472, opened 2020-11-28
Refactor Rmd commands and configure


Ikuyadeu:
Our extension supports extends to Rmd, not only R.

Like the relationship with Python and Jupyter, R and Rmd should be separated.
Such as `r.selectCurrentChunk` should be `rmd.selectCurrentChunk`
Also, commands and configure values can be the same naming convention as Jupyter and Python extension commands.

PR #699, opened 2021-07-10
RMD Preview fixes

ElianHugh:
Implements changes suggested by @ManuelHentschel for R Markdown preview. Should not have any major change in functionality.

Changes:
- path -> fsPath
- ~~'/' -> path.sep~~
- themeBool -> useDarkTheme
- command category 'R' -> 'R Markdown'
- getRPath(false) -> true
- extDir -> ~~extensionDirectory~~ homeExtDir
- ~~tmpDir -> temporaryDirectory~~
- Handle untitled documents by prompting the user to save the doc prior to rendering
- Less eager directory creation 
- check exit code rather than error message for knitting errors
- refresh now re-knits the doc, rather than refreshing the html

Tested on Linux, I don't currently have access to Windows.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -84,14 +84,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
         'r.goToNextChunk': rmarkdown.goToNextChunk,
         'r.runChunks': rTerminal.runChunksInTerm,
 
-        'r.markdown.showPreviewToSide': () => rMarkdownPreview.previewRmd(vscode.ViewColumn.Beside),
-        'r.markdown.showPreview': (uri: vscode.Uri) => rMarkdownPreview.previewRmd(vscode.ViewColumn.Active, uri),
-        'r.markdown.preview.refresh': () => rMarkdownPreview.updatePreview(),
-        'r.markdown.preview.openExternal': () => void rMarkdownPreview.openExternalBrowser(),
-        'r.markdown.preview.showSource': () => rMarkdownPreview.showSource(),
-        'r.markdown.preview.toggleStyle': () => rMarkdownPreview.toggleTheme(),
-        'r.markdown.preview.enableAutoRefresh': () => rMarkdownPreview.enableAutoRefresh(),
-        'r.markdown.preview.disableAutoRefresh': () => rMarkdownPreview.disableAutoRefresh(),
+        'r.rmarkdown.showPreviewToSide': () => rMarkdownPreview.previewRmd(vscode.ViewColumn.Beside),
+        'r.rmarkdown.showPreview': (uri: vscode.Uri) => rMarkdownPreview.previewRmd(vscode.ViewColumn.Active, uri),
+        'r.rmarkdown.preview.refresh': () => rMarkdownPreview.updatePreview(),
+        'r.rmarkdown.preview.openExternal': () => void rMarkdownPreview.openExternalBrowser(),
+        'r.rmarkdown.preview.showSource': () => rMarkdownPreview.showSource(),
+        'r.rmarkdown.preview.toggleStyle': () => rMarkdownPreview.toggleTheme(),
+        'r.rmarkdown.preview.enableAutoRefresh': () => rMarkdownPreview.enableAutoRefresh(),
+        'r.rmarkdown.preview.disableAutoRefresh': () => rMarkdownPreview.disableAutoRefresh(),
 
         // editor independent commands
         'r.createGitignore': rGitignore.createGitignore,
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -86,7 +86,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
 
         'r.markdown.showPreviewToSide': () => rMarkdownPreview.previewRmd(vscode.ViewColumn.Beside),
         'r.markdown.showPreview': (uri: vscode.Uri) => rMarkdownPreview.previewRmd(vscode.ViewColumn.Active, uri),
-        'r.markdown.preview.refresh': () => rMarkdownPreview.refreshPanel(),
+        'r.markdown.preview.refresh': () => rMarkdownPreview.updatePreview(),
         'r.markdown.preview.openExternal': () => void rMarkdownPreview.openExternalBrowser(),
         'r.markdown.preview.showSource': () => rMarkdownPreview.showSource(),
         'r.markdown.preview.toggleStyle': () => rMarkdownPreview.toggleTheme(),
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,7 +5,6 @@
 import * as vscode from 'vscode';
 import * as os from 'os';
 import path = require('path');
-import fs = require('fs');
 
 // functions etc. implemented in this extension
 import * as preview from './preview';
@@ -25,8 +24,8 @@ import * as httpgdViewer from './plotViewer';
 import { RMarkdownPreviewManager } from './rmarkdown/preview';
 
 // global objects used in other files
-export const extensionDirectory: string = path.join(os.homedir(), '.vscode-R');
-export const temporaryDirectory: string = path.join(extensionDirectory, 'tmp');
+export const extensionDirectory = (): string => util.getDir(path.join(os.homedir(), '.vscode-R'));
+export const temporaryDirectory = (): string => util.getDir(path.join(extensionDirectory(), 'tmp'));
 export let rWorkspace: workspaceViewer.WorkspaceDataProvider | undefined = undefined;
 export let globalRHelp: rHelp.RHelp | undefined = undefined;
 export let extensionContext: vscode.ExtensionContext;
@@ -36,14 +35,6 @@ export let rMarkdownPreview: RMarkdownPreviewManager | undefined = undefined;
 
 // Called (once) when the extension is activated
 export async function activate(context: vscode.ExtensionContext): Promise<apiImplementation.RExtensionImplementation> {
-    if (!fs.existsSync(extensionDirectory)) {
-        fs.mkdirSync(extensionDirectory);
-    }
-
-    if (!fs.existsSync(temporaryDirectory)) {
-        fs.mkdirSync(temporaryDirectory);
-    }
-
     // create a new instance of RExtensionImplementation
     // is used to export an interface to the help panel
     // this export is used e.g. by vscode-r-debugger to show the help panel from within debug sessions
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -24,8 +24,8 @@ import * as httpgdViewer from './plotViewer';
 import { RMarkdownPreviewManager } from './rmarkdown/preview';
 
 // global objects used in other files
-export const extensionDirectory = (): string => util.getDir(path.join(os.homedir(), '.vscode-R'));
-export const temporaryDirectory = (): string => util.getDir(path.join(extensionDirectory(), 'tmp'));
+export const homeExtDir = (): string => util.getDir(path.join(os.homedir(), '.vscode-R'));
+export const tmpDir = (): string => util.getDir(path.join(homeExtDir(), 'tmp'));
 export let rWorkspace: workspaceViewer.WorkspaceDataProvider | undefined = undefined;
 export let globalRHelp: rHelp.RHelp | undefined = undefined;
 export let extensionContext: vscode.ExtensionContext;
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -25,8 +25,8 @@ import * as httpgdViewer from './plotViewer';
 import { RMarkdownPreviewManager } from './rmarkdown/preview';
 
 // global objects used in other files
-export const extDir: string = path.join(os.homedir(), '.vscode-R');
-export const tmpDir: string = path.join(extDir, 'tmp');
+export const extensionDirectory: string = path.join(os.homedir(), '.vscode-R');
+export const temporaryDirectory: string = path.join(extensionDirectory, 'tmp');
 export let rWorkspace: workspaceViewer.WorkspaceDataProvider | undefined = undefined;
 export let globalRHelp: rHelp.RHelp | undefined = undefined;
 export let extensionContext: vscode.ExtensionContext;
@@ -36,12 +36,12 @@ export let rMarkdownPreview: RMarkdownPreviewManager | undefined = undefined;
 
 // Called (once) when the extension is activated
 export async function activate(context: vscode.ExtensionContext): Promise<apiImplementation.RExtensionImplementation> {
-    if (!fs.existsSync(extDir)) {
-        fs.mkdirSync(extDir);
+    if (!fs.existsSync(extensionDirectory)) {
+        fs.mkdirSync(extensionDirectory);
     }
 
-    if (!fs.existsSync(tmpDir)) {
-        fs.mkdirSync(tmpDir);
+    if (!fs.existsSync(temporaryDirectory)) {
+        fs.mkdirSync(temporaryDirectory);
     }
 
     // create a new instance of RExtensionImplementation
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -9,7 +9,7 @@ import crypto = require('crypto');
 
 
 import { config, doWithProgress, getRpath, readContent, setContext, escapeHtml } from '../util';
-import { extensionContext, tmpDir } from '../extension';
+import { extensionContext, temporaryDirectory } from '../extension';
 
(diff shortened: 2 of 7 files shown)


src/extension.ts
```diff
@@ -25,8 +25,8 @@ import * as httpgdViewer from './plotViewer';
 import { RMarkdownPreviewManager } from './rmarkdown/preview';
 
 // global objects used in other files
-export const extDir: string = path.join(os.homedir(), '.vscode-R');
-export const tmpDir: string = path.join(extDir, 'tmp');
+export const extensionDirectory: string = path.join(os.homedir(), '.vscode-R');
```
renkun-ken, line 28: I think both `extDir` or `extensionDirectory` are ambiguous because it could mean the installation directory, but I don't come up with a name that makes more sense. Any ideas?

src/extension.ts
```diff
@@ -25,8 +25,8 @@ import * as httpgdViewer from './plotViewer';
...
+export const temporaryDirectory: string = path.join(extensionDirectory, 'tmp');
```
renkun-ken, line 29: It looks like `tmp` or `temp` are already a convention to indicate "temporary` (e.g. `os.tmpdir()`). Not sure if it really looks better to call it in full.

renkun-ken: This is a bit ambiguous too, i.e. `temporaryDirectory` just looks like an alias of `os.tmpdir()`?

src/rmarkdown/preview.ts
```diff
@@ -24,12 +24,12 @@ class RMarkdownPreview extends vscode.Disposable {
...
     constructor(title: string, cp: cp.ChildProcessWithoutNullStreams, panel: vscode.WebviewPanel,
         resourceViewColumn: vscode.ViewColumn, outputUri: vscode.Uri, uri: vscode.Uri,
-        RMarkdownPreviewManager: RMarkdownPreviewManager, themeBool: boolean, autoRefresh: boolean) {
+        RMarkdownPreviewManager: RMarkdownPreviewManager, useDarkTheme: boolean, autoRefresh: boolean) {
         super(() => {
             this.cp?.kill('SIGKILL');
             this.panel?.dispose();
             this.fileWatcher?.close();
-            fs.removeSync(this.outputUri.path);
+            fs.removeSync(this.outputUri.fsPath);
```
renkun-ken, line 32: Would you like to try working with an untitled document with rmd language selected? How should we handle this case? Let user save the document to get a valid `fsPath` first, or we write the document to a temp file and render that file as if we render this untitled document?

src/rmarkdown/preview.ts
```diff
@@ -38,26 +38,26 @@ class RMarkdownPreview extends vscode.Disposable {
         this.resourceViewColumn = resourceViewColumn;
         this.outputUri = outputUri;
         this.autoRefresh = autoRefresh;
-        void this.refreshContent(themeBool);
+        void this.refreshContent(useDarkTheme);
```
renkun-ken, line 41: Not sure whether it works if we use a high-contrast theme.

ElianHugh: Maybe something like useCodeTheme?

src/extension.ts
ElianHugh: I agree. Hmm, what about... localDirectory? homeDirectory?

src/rmarkdown/preview.ts
```diff
@@ -299,12 +296,10 @@ export class RMarkdownPreviewManager {
...
             const lim = '---vsc---';
             const re = new RegExp(`.*${lim}(.*)${lim}.*`, 'ms');
-            const outputFile = path.join(tmpDir, crypto.createHash('sha256').update(fileUri.fsPath).digest('hex') + '.html');
+            const outputFile = path.join(temporaryDirectory, crypto.createHash('sha256').update(fileUri.fsPath).digest('hex') + '.html');
             const cmd = (
                 `${this.rPath} --silent --slave --no-save --no-restore -e ` +
-                `"cat('${lim}',
-                rmarkdown::render('${String(fileUri.path)}', output_format = rmarkdown::html_document(), output_file = '${outputFile}', intermediates_dir = '${tmpDir}'),
-                '${lim}', sep='')"`
+                `"cat('${lim}', rmarkdown::render('${fileUri.fsPath}', output_format = rmarkdown::html_document(), output_file = '${outputFile}', intermediates_dir = '${temporaryDirectory}'), '${lim}', sep='')"`
```
ManuelHentschel, line 302: The following seems to work here:
``` ts
            const cmd = (
                `${this.rPath} --silent --slave --no-save --no-restore -e ` +
                `"cat('${lim}', rmarkdown::render(` +
                `'${fileUri.fsPath.replace(/\\/g, '/')}',` +
                `output_format = rmarkdown::html_document(),` +
                `output_file = '${outputFile.replace(/\\/g, '/')}',` +
                `intermediates_dir = '${temporaryDirectory.replace(/\\/g, '/')}'), '${lim}',` +
                `sep='')"`
            );
```

src/extension.ts
ManuelHentschel: Would it be feasible to simply use `os.tmpdir()` then (if I recall correctly, there might be issues with user privacy on some platforms)?

And if not, why is this directory declared/created here? It doesn't seem to be used anywhere eles in the code, so wouldn't it be enough to put this in the markdown-preview code?

src/extension.ts
```diff
@@ -36,12 +36,12 @@ export let rMarkdownPreview: RMarkdownPreviewManager | undefined = undefined;
...
-        fs.mkdirSync(extDir);
+    if (!fs.existsSync(extensionDirectory)) {
+        fs.mkdirSync(extensionDirectory);
     }
 
-    if (!fs.existsSync(tmpDir)) {
-        fs.mkdirSync(tmpDir);
+    if (!fs.existsSync(temporaryDirectory)) {
+        fs.mkdirSync(temporaryDirectory);
     }
```
ManuelHentschel, line 45: This creates these directories quite eagerly, i.e. as soon as the extension activates. Wouldn't it be better to create them only if they're actually used?

This could e.g. be implemented by adding a function that creates the directory (if necessary) and then returns its path

ManuelHentschel:
This is working quite nicely now, thanks for the nice work! :)

src/extension.ts
renkun-ken: Using `os.tmpdir()` will have some user privacy problem, unless we create something like `/tmp/vscode-R/$USER` and make the folder have mode 700.

Just notice that https://github.com/REditorSupport/vscode-R/blob/master/src/helpViewer/helpProvider.ts#L289 uses `os.tmpdir()` which is probably subject to multi-user clash problem.

renkun-ken:
There seems an inconsistency that the commands are `r.markdown.*`, global context variables are `r.preview.*` and the settings are `r.rmarkdown.*`?

ElianHugh:
Maybe we make it uniform: all under r.markdown? 
R.rmarkdown is a bit redundant

renkun-ken:
> Maybe we make it uniform: all under r.markdown?
> R.rmarkdown is a bit redundant

I agree. I think if a put a note in the news, and it is okay to rename existing settings of `r.rmarkdown.*` to `r.markdown.*`.

eitsupi:
Thanks for the great work.
Does the naming convention of the command have anything to do with this issue #472?

src/rmarkdown/preview.ts
ElianHugh: It might be a bit tricky because we would have to watch for changes on the untitled rmd file in order to sync it with the temp rmd file. I'm not aware of methods for tracking changes on virtual docs

ManuelHentschel:
> > Maybe we make it uniform: all under r.markdown?
> > R.rmarkdown is a bit redundant
> 
> I agree. I think if we put a note of breaking changes in the news, and it is okay to rename existing settings of `r.rmarkdown.*` to `r.markdown.*`.

I wouldn't say `r.rmarkdown` is redundant. All our settings/commands start with `r.` since they are contributed by the R extension. Changing the setting names to `r.markdown` would make it look like they affect .md files rather than .rmd.

ElianHugh:
To me, I think of r.markdown in my head as 
"R: Markdown", and r.rmarkdown as "R: R Markdown". However, I get your point re the extension naming scheme

src/extension.ts
ElianHugh: Given it's created in the ~/ directory, I think something relating to 'home' is closer to the right name.

Just throwing ideas out there, because honestly I'm not sure either:
- homeDir
- homeExtDir
- extHomeDir

src/rmarkdown/preview.ts
renkun-ken: The error is like the following with untitled rmd:

```
[VSC-R] Untitled-1 process started
Error in abs_path(input) : The file 'Untitled-1' does not exist.
Calls: cat -> <Anonymous> -> setwd -> dirname -> abs_path
In addition: Warning message:
In normalizePath(path, winslash = winslash, mustWork = mustWork) :
  path[1]="Untitled-1": No such file or directory

Backtrace:

1: stop("The file '", x, "' does not exist.")

2: abs_path(input)
3: dirname(abs_path(input))
4: setwd(dirname(abs_path(input)))

5: rmarkdown::render("Untitled-1", output_format = rmarkdown::html_document(), 

6: cat("---vsc---", rmarkdown::render("Untitled-1", output_format = rmarkdown::html_document(), 

Exiting on error

[VSC-R] Untitled-1 process exited with exit code 1
```

ElianHugh: Right, that would be relatively straight forward to solve. Given the auto refresh is based on changes to a saved file, having it auto refresh would be the difficult part. ...and I'm not sure it would make sense since the file wouldn't be untitled anymore once saved.

src/extension.ts
ElianHugh: Should now only be created when the variable is referred to:

```js
export const extensionDirectory = (): string => util.getDir(path.join(os.homedir(), '.vscode-R'));
export const temporaryDirectory = (): string => util.getDir(path.join(extensionDirectory(), 'tmp'));
```

Names tbd :)

src/extension.ts
ElianHugh: Perhaps we should set all references of os.tmpdir (and future references) to temporaryDirectory*?

* barring issues with the name

src/extension.ts
renkun-ken: I like `homeExtDir`, which is almost the same as its path.

src/rmarkdown/preview.ts
renkun-ken: Looks like we should ask the user to save the file first. If the document is untitled, then we don't render it.

src/rmarkdown/preview.ts
```diff
@@ -182,6 +182,11 @@ export class RMarkdownPreviewManager {
         const fileUri = uri ?? vscode.window.activeTextEditor.document.uri;
         const fileName = fileUri.fsPath.substring(fileUri.fsPath.lastIndexOf(path.sep) + 1);
         const currentViewColumn: vscode.ViewColumn = vscode.window.activeTextEditor?.viewColumn ?? vscode.ViewColumn.Active ?? vscode.ViewColumn.One;
+        // abort if the file is untitled
+        if (!uri && vscode.window.activeTextEditor.document.isUntitled) {
+            void vscode.window.showWarningMessage('Cannot knit an untitled file. Please save the document and try again.');
```
renkun-ken, line 187: Instead of show warning message, maybe we should execute command `workbench.action.files.save`?

renkun-ken:
It looks like the progress only shows when executing the preview command. The refresh does not show up the progress, and thus there is no way to cancel the rendering. I wonder if it makes sense to also show the progress on each refresh?

ElianHugh:
> It looks like the progress only shows when executing the preview command. The refresh does not show up the progress, and thus there is no way to cancel the rendering. I wonder if it makes sense to also show the progress on each refresh?

Whoops, yes that should happen. I'll fix that tonight

src/rmarkdown/preview.ts
ManuelHentschel: Or disable the preview command when the current file is not an RMarkdown file? That's how the integrated markdown preview behaves.

Else, would it work to copy the file contents to a tempfile (in the background) and still render the preview for untitled files?

renkun-ken: An untitled document could be assigned a language. If you create an untitled document and choose `md` language, then the markdown preview will work.

renkun-ken: This reminds me that should we always save document before rendering?

BerriJ: Personally, I would prefer this. This would be also in line with other extensions like latex-workshop which saves at compilation.

ManuelHentschel: > An untitled document could be assigned a language.

Ah ok, didn't know that, thanks

renkun-ken:
When I save the document, it looks like the document is rendered twice.

ElianHugh:
Seems to be an issue where the uri passed from the file explorer context menu is not equivalent to `vscode.window.activeTextEditor.document.uri` and causes issues. Not sure why this would be the case, but I guess we could just store `uri.path` as the key rather than `vscode.Uri`.

image

renkun-ken:
> Seems to be an issue where the uri passed from the file explorer context menu is not equivalent to `vscode.window.activeTextEditor.document.uri` and causes issues. Not sure why this would be the case, but I guess we could just store `uri.path` as the key rather than `vscode.Uri`.

OK. Using `uri.path` looks better.

ElianHugh:
Ideally we would keep vscode.Uri so that we still have access to its attributes (e.g. fsPath). However, no matter how I create a uri, an fsPath is not supplied.

```ts
// no fsPath
vscode.Uri.parse(fileUri.path);
vscode.Uri.file(fileUri.path);

// no fsPath
const fragment = /([^/\\]+$)/g.exec(fileUri.path)[0];
await vscode.workspace.findFiles(`**/${fragment}`, null, 10)[0]);

// no fsPath
vscode.Uri.from(
            {
                scheme: 'file',
                path: fileUri.path
            }
        );
```

I have no idea what constitutes the creation of an fsPath.

src/rmarkdown/preview.ts
```diff
@@ -183,8 +183,8 @@ export class RMarkdownPreviewManager {
     }
 
     public async previewRmd(viewer: vscode.ViewColumn, uri?: vscode.Uri): Promise<void> {
-        const fileUri = uri ?? vscode.window.activeTextEditor.document.uri;
-        const fileName = fileUri.fsPath.substring(fileUri.fsPath.lastIndexOf(path.sep) + 1);
+        const filePath = uri ? uri.fsPath : vscode.window.activeTextEditor.document.uri.fsPath;
+        const fileName = filePath.substring(filePath.lastIndexOf(path.sep) + 1);
```
renkun-ken, line 187: Is there a reason not to use `path.basename()`?

src/rmarkdown/preview.ts
```diff
@@ -132,21 +132,21 @@ class RMarkdownPreviewStore extends vscode.Disposable {
         });
     }
 
-    public add(uri: vscode.Uri, preview: RMarkdownPreview): Map<vscode.Uri, RMarkdownPreview> {
-        return this.store.set(uri, preview);
+    public add(fsPath: string, preview: RMarkdownPreview): Map<string, RMarkdownPreview> {
```
renkun-ken, line 135: Do we have both `filePath` and `fsPath`?

renkun-ken:
If the render process exits with an error, the progress will never close.

ElianHugh:
> 
> 
> If the render process exits with an error (e.g. R syntax error in code chunks), the progress will never close.

I can't replicate this, this is what should be happening when an error occurs: 

EahsGSe8YW

It may be that we expand or change the error detection here:

```ts
 childProcess.stderr.on('data', (data: Buffer) => {
                const dat = data.toString('utf8');
                this.rMarkdownOutput.appendLine(dat);
                if (dat.includes('Execution halted')) {
                    reject({ cp: childProcess, wasCancelled: false });
                }
            });
```

renkun-ken:
Kapture 2021-07-13 at 20 52 07

Let me take a look at this.

renkun-ken:
It turns out to be caused by my customized error handling specified in `options(error = function(...))` where it does not print "Execution halted" on error but "Exiting on error".

renkun-ken:
Looks like we should not rely on the error message "Execution halted" (e.g. R in other languages) but the exit code of the process.

renkun-ken (approved):
LGTM

ManuelHentschel (approved):
Works nicely, thanks!

PR #706, opened 2021-07-15
Remove show plot history command

### LINKED ISSUE #686: Remove "R: Show plot history"
Since [httpgd webview and plot viewer](https://github.com/Ikuyadeu/vscode-R/wiki/Plot-viewer) provide good support of plot history navigation, I think we could deprecate and remove the current "R: Show plot history" command and feature and thus we no longer depend on `"fotorama": "^4.6.4"`.

renkun-ken:
# What problem did you solve?

Closes #686 

Since the old plot viewer based on png replays is quite limited and in some cases it does not produce desired result (e.g. plot in a for loop, multi-page plots), I suggest we remove this feature in favor of the httpgd approach.

This PR removes the command "R: Show plot history" and related code that copies replayed png files in a history folder and removes the dependency on `fotorama`.

For better plot history support, use [httpgd](https://github.com/nx10/httpgd) instead as introduced in the [wiki](https://github.com/REditorSupport/vscode-R/wiki/Plot-viewer).

Disable httpgd plot viewer.

```r
plot(rnorm(100))
```

should correctly produce a png file which is open in a new tab. The "R: Show plot history" command is gone.

### DIFF
--- a/R/vsc.R
+++ b/R/vsc.R
@@ -176,13 +176,10 @@ if (use_httpgd && "httpgd" %in% .packages(all.available = TRUE)) {
 } else if (use_httpgd) {
   message("Install package `httpgd` to use vscode-R with httpgd!")
 } else if (show_plot) {
-  dir_plot_history <- file.path(dir_session, "images")
-  dir.create(dir_plot_history, showWarnings = FALSE, recursive = TRUE)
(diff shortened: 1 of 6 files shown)


ElianHugh (approved):
LGTM, no unexpected issues while testing. Thanks!

PR #707, opened 2021-07-15
Update r.rmarkdown.codeLensCommands

ElianHugh:
# What problem did you solve?
The current r.rmarkdown.codeLensCommands setting is prone to typos. This PR uses an enum to restrict the input. As this does not change the setting's input type, user settings should be unaffected.

image


renkun-ken (approved):
Works nicely. LGTM.

PR #708, opened 2021-07-15
Use ag-grid in data viewer

### LINKED ISSUE #506: Enhance the data viewer
Currently, we are using [DataTables](https://datatables.net/) in the WebView as the data viewer with session watcher enabled. It has the following capabilities (but we are not using all of them at the moment):

* Sorting a column
* Always show column header on page top
* Different background color of odd/even rows
* Highlight row on hover
* Highlight rows on click
* Search text and filter rows over the entire table
* Paging (not used yet)

There are some alternative data viewer widgets that could provide more power features such as column filtering by value or range (like in RStudio), multi-column ordering. I guess we could investigate into this and see if we could provide a more powerful data viewer.

More powerful alternatives:

* [Handsontable](https://handsontable.com/) (maybe there's potential license issue)

renkun-ken:
# What problem did you solve?

Closes #506

This PR replaces [DataTables](https://datatables.net) with [ag-grid](https://www.ag-grid.com/) in the data viewer.

## (If you have)Screenshot

https://user-images.githubusercontent.com/4662568/125810901-cd8f28d4-a35b-4399-903e-4ae3d03b4e4c.mp4

View any data.frame or matrix and the data viewer will use `ag-grid` to present the table with the following features:

* Sortable columns (hold shift for multi-sort)
* Filterable columns (filter number, date, text by conditions)
* Resizable columns
* Movable columns
* Pagination
* Great performance
* Dark theme


danielbasso:
Looking great! By any chance "Pagination" means dynamic loading? It would be perfect.

renkun-ken:
> Looking great! By any chance "Pagination" means dynamic loading? It would be perfect.

It's not dynamic loading here.

ElianHugh:
Is there another setup requirement in addition to running `yarn`? I'm getting several errors due to missing files, e.g.: 

```ts
(Error: Unable to resolve non-existing file 'vscode-remote://wsl+ubuntu-20.04/vscode-resource?{"requestResourcePath":"/home/elian/coding/VS/vscode-R/dist/resources/ag-grid-community.min.noStyle.js"}')
```

renkun-ken:
We need those js css resources to be copied from `node_modules` to `dist/resources` as specified in `webpack.config.js`. You might need to run `node .guthub/scripts/enableWebpack.js` and run.

src/session.ts
```diff
@@ -354,44 +354,88 @@ export async function getTableHtml(webview: Webview, file: string): Promise<stri
...
+      gridOptions.columnApi.getAllColumns().forEach(function (column) {
+        allColumnIds.push(column.colId);
       });
+      gridOptions.columnApi.autoSizeColumns(allColumnIds, skipHeader);
+    }
+    const gridOptions = {
+      defaultColDef: {
+        sortable: true,
+        resizable: true,
+        editable: true,
```
ElianHugh, line 404: Given that editing the file doesn't effect the R environment, would it be worth changing editable to `false`?

renkun-ken: If `editable: false`, then I find that we cannot copy the text in the WebView.

ElianHugh: Adding `enableCellTextSelection: true` under gridOptions appears to allow selection when `editable: false`

renkun-ken: Thanks for pointing to the right option!

src/session.ts
```diff
@@ -412,12 +456,15 @@ export async function getListHtml(webview: Webview, file: string): Promise<strin
   <link href="${String(webview.asWebviewUri(Uri.file(path.join(resDir, 'jquery.json-viewer.css'))))}" rel="stylesheet">
   <style type="text/css">
     body {
-        color: black;
-        background-color: white;
+        color: var(--vscode-editor-foreground);;
+        background-color: var(--vscode-editor-background);;
     }
     pre#json-renderer {
       border: 1px solid #aaa;
```
ElianHugh, line 464: Adding the following seems to make the list view nicer for other themes:

```css
pre#json-renderer {
      border: 1px solid var(--vscode-editorLineNumber-foreground);
}

ul.json-dict, ol.json-array {
        border-left: 1px dotted var(--vscode-editorLineNumber-foreground);
}
```

E.g.:

image

image

image

renkun-ken: Definitely! Thanks!

ElianHugh:
Lists and data.frames work nicely both locally and on live share! Not a must, but do you think it would be worth using some custom css to better fit the vscode theme? I threw a little css together for an example:

image
image

Otherwise, LGTM

ElianHugh (approved):

renkun-ken:
>Lists and data.frames work nicely both locally and on live share! Not a must, but do you think it would be worth using some custom css to better fit the vscode theme? I threw a little css together for an example:

Looks great! Would you like to provide a better css here?

src/session.ts
```diff
@@ -375,6 +374,71 @@ export async function getTableHtml(webview: Webview, file: string): Promise<stri
...
+        background-color: var(--vscode-sideBar-background);
+    }
+
+    [class*="vscode"] div.ag-header-cell[aria-sort="ascending"], div.ag-header-cell[aria-sort="descending"] {
+        color: var(--vscode-textLink-activeForeground);
+    }
+
+    /* Styling for rows and cells */
+
+    [class*="vscode"] .ag-row-selected {
```
renkun-ken, line 394: Is `.ag-row-hover` missing? The theme has a hover item.

src/session.ts
```diff
@@ -375,6 +374,71 @@ export async function getTableHtml(webview: Webview, file: string): Promise<stri
         padding: 0;
         overflow: auto;
     }
+
+    /* Styling for wrapper and header */
+
+    [class*="vscode"] div.ag-root-wrapper {
```
renkun-ken, line 380: Is `[class*="vscode"]` necessary here?

ElianHugh: This was mostly to have higher priority over the base theme. An alternative would be !important?

src/session.ts
ElianHugh: Do you think sharing the `.ag-row-selected` styling would be okay? Or should it be a unique colour?

renkun-ken: `.ag-row` should appear not highlighted, `.ag.row-hover` slightly highlighted, `.ag-row-selected` more highlighted. So I think it is different from `.ag-row-selected` or otherwise user cannot distinguish hovered rows and selected rows.

src/session.ts
renkun-ken: OK. Makes sense. Maybe a comment on this could be better to clarify the intention.

src/session.ts
renkun-ken: I have added a `.ag-row-hover`.

renkun-ken:
I've done some changes to make it also work with high-contrast theme. Now everything looks good.

ElianHugh:
Looks good and works well, thanks!

PR #738, opened 2021-07-31
Rename liveshare folder to liveShare

renkun-ken:
This PR renames `src/liveshare` to `src/liveShare` to be consistent with other folders.


ElianHugh:
Seems reasonable to me! Would this mean the rmarkdown folder should be rMarkdown? Or is it treated as one word?

renkun-ken:
I prefer treating rmarkdown or RMarkdown as one word.

PR #743, opened 2021-08-06
*Proof of concept* accessing VSC settings in R

ElianHugh:
# **proof of concept** 

This PR allows loading R-related vsc settings in R. This reduces the reliance on 
setting options in the .Rprofile, and means that you don't need to have settings that have no effect outside of VSC-R. This also allows maintainers to restrict invalid inputs.

This could also be used for increased customisation E.g., as outlined by [@renkun-ken](https://github.com/REditorSupport/vscode-R/issues/584#issuecomment-802927563_), this could be extended to modify .Rprofile execution order

_Caveats_:

- The default value of a setting *must* be identical to the R option default, as VSCode settings are not written to the settings.kson unless the default value is changed.
- May require an environment term to locate the settings.json in some cases
- Could have issues with malformed settings.json files

## Screenshot

image

---

Note: this PR is a proof-of-concept, and is a little rough around the edges. I'd love to hear whether this is something that is useful, and any ideas/issues that you think may exist/occur

### DIFF
--- a/R/getSettings.R
+++ b/R/getSettings.R
@@ -18,12 +18,24 @@ fileCon <- if (file.exists("~/.config/Code/User/settings.json")) {
 vsc_settings <- jsonlite::fromJSON(paste(readLines(fileCon), collapse = ""))
 ops <- vsc_settings[grep("r.rOptions", names(vsc_settings))]$r.rOptions
 
+get_val <- function(x) {
+    switch(EXPR = x,
+        "Two" = "Two",
+        "Active" = "Active",
+        "Beside" = "Beside",
+        "FALSE" = FALSE,
+        "TRUE" = TRUE,
+        "0" = 0,
+        "2" = 2,
+        x
+    )
+}
 
-lapply(names(ops), function(x) {
-    val <- ops[[x]]
-    switch(x,
+suppressMessages(lapply(names(ops), function(x) {
+    val <- get_val(ops[[x]])
+    switch(EXPR = x,
         "vsc.plot" = options("vsc.plot" = val),
         "vsc.globalenv" = options("vsc.globalenv" = val)
         # etc.
     )
-})
+}))
--- a/R/getSettings.R
+++ b/R/getSettings.R
@@ -1,16 +1,29 @@
-fileCon <- file("~/.config/Code/User/settings.json")
+fileCon <- if (file.exists("~/.config/Code/User/settings.json")) {
+    # windows
+    file("~/.config/Code/User/settings.json")
+} else if (file.exists("~/Library/Application Support/Code/User/settings.json")) {
+    # mac
+    file("~/.config/Code/User/settings.json")
+} else if (file.exists("~/.config/Code/User/settings.json")) {
+    # linux
+    file("~/.config/Code/User/settings.json")
+} else if (file.exists(paste0("/mnt///c/Users/", Sys.getenv("USER"), "/AppData/Roaming/Code/User/settings.json"))) {
+    # WSL
+    file(paste0("/mnt///c/Users/", Sys.getenv("USER"), "/AppData/Roaming/Code/User/settings.json"))
+} else {
+    NULL
+}
+
+
 vsc_settings <- jsonlite::fromJSON(paste(readLines(fileCon), collapse = ""))
 ops <- vsc_settings[grep("r.rOptions", names(vsc_settings))]$r.rOptions
 
 
-.vsc.setOption <- function(x) {
+lapply(names(ops), function(x) {
     val <- ops[[x]]
-    switch(
-        x,
+    switch(x,
         "vsc.plot" = options("vsc.plot" = val),
         "vsc.globalenv" = options("vsc.globalenv" = val)
         # etc.
     )
-}
-
-lapply(names(ops), .vsc.setOption)
+})
--- a/R/getSettings.R
+++ b/R/getSettings.R
@@ -1,0 +1,16 @@
+fileCon <- file("~/.config/Code/User/settings.json")
+vsc_settings <- jsonlite::fromJSON(paste(readLines(fileCon), collapse = ""))
+ops <- vsc_settings[grep("r.rOptions", names(vsc_settings))]$r.rOptions
+
+
+.vsc.setOption <- function(x) {
+    val <- ops[[x]]
+    switch(
+        x,
+        "vsc.plot" = options("vsc.plot" = val),
+        "vsc.globalenv" = options("vsc.globalenv" = val)
+        # etc.
+    )
+}
+
+lapply(names(ops), .vsc.setOption)
--- a/R/getSettings.R
+++ b/R/getSettings.R
@@ -1,3 +1,6 @@
+# try to find the VSC settings file.
+# can be tricky for specific platforms, such as WSL,
+# where the settings are located on the window's side
 fileCon <- if (file.exists("~/.config/Code/User/settings.json")) {
     # windows
     file("~/.config/Code/User/settings.json")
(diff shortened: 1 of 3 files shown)


R/getSettings.R
```diff
@@ -0,0 +1,66 @@
+# try to find the VSC settings file.
+# can be tricky for specific platforms, such as WSL,
+# where the settings are located on the window's side
+fileCon <- if (file.exists("~/.config/Code/User/settings.json")) {
```
renkun-ken, line 4: The problem mostly resides here as I'm not sure if there is a clean way to determine the path of the vscode config file under multiple scenarios: local instance, remote development, system installation, user installation, portable, etc.

renkun-ken: I think the best way is not let the R session to find the vscode settings file but vscode-R writes a JSON file in `~/.vscode-R/settings.json` where only `vsc.*` related settings are included. On configuration update of these settings, the file should be updated too.

Then `.vsc.attach()` should read `~/.vscode-R/settings.json` to apply the latest settings.

ElianHugh: Thanks, that's a great idea :)

ManuelHentschel:
Is this meant to completely replace/overwrite the options set in .Rprofile?

R/getSettings.R
```diff
@@ -1,66 +1,60 @@
...
-        "vsc.useHttpgd" = options("vsc.use_httpgd" = val),
-        "vsc.globalEnv" = options("vsc.globalenv" = val),
-        "vsc.showObjectSize" = options("vsc.show_object_size" = val)
-    )
-})
+        # lhs = vscode setting name
+        # rhs name = R options name
+        switch(EXPR = x,
+            # arrays
+            "vsc.plot" = options("vsc.plot" = val),
```
renkun-ken, line 44: Such camelCase to snake_case translation looks quite redundant and confusing. Would it be helpful if we keep the original names in R?

ElianHugh: I agree - the names were mostly a placeholder for the time being. I think ideally the names would be more descriptive, so that someone that hasn't come across the wiki before could understand what they do

ElianHugh:
> 
> 
> Is this meant to completely replace/overwrite the options set in .Rprofile?

I think ideally this would just be a more user-friendly way of setting options, and would be overwritten by the user's .Rprofile options

R/.Rprofile
```diff
@@ -1,3 +1,69 @@
+# Source VSCode-R options
```
renkun-ken, line 1: `.Rprofile` only applies to session started from `R: Create R terminal`. User managed sessions do not respect this profile. Maybe we need to put it back to `init.R` so that in both cases the settings are used.

ElianHugh: Thanks, I wasn't aware of this - I've added it back and put in a check for already-set options

R/getSettings.R
ElianHugh: More broadly, are there any opinions/ideas for how these options should:

1. be organised? At the moment it's under a fairly non-descriptive "bools" and "arrays", but this clumps together options that may only be tangentially related. Also, would these options better fall under already-defined categories, such as `r.rterm.rOptions`?

2. be named? One idea I had was have descriptive names, and then in the description state the option's name. E.g., 

**Show Object Size** (showObjectSize)
Show object size when hovering over a workspace viewer item.
Sets the option `vsc.show_object_size` in R.

ElianHugh: Re: categorising the settings, another option would be to split these under already-existing related categories. E.g., plot options all moved to `r.plots`, viewer options moved under session, etc.

renkun-ken: Now that all these options are only useful for vscode-R, there might be a point to only keep them in vscode settings and deprecate these options in R. Then it is much easier to put each setting in the appropriate category.

ElianHugh: I agree going forward that it's better to keep this on the VSCode side, but  for backwards compatibility it might be a good idea to keep the options around and deprecate them a bit further down the line

renkun-ken: > I agree going forward that it's better to keep this on the VSCode side, but for backwards compatibility it might be a good idea to keep the options around and deprecate them a bit further down the line

Yes, exactly. Just go ahead with the most consistent names and categories in vscode. Maybe it is useful to put the corresponding R option in the settings description to make it easier to search.

R/init.R
```diff
@@ -117,11 +117,27 @@ init_last <- function() {
     }
 
     if (!is.null(fileCon)) {
+      rSettings <- c(
+        "plot.useHttpgd",
+        "workspaceViewer.showObjectSize",
+        "session.emulateRStudioAPI",
+        "session.levelOfObjectDetail",
+        "session.watchGlobalEnvironment",
+        "session.viewers.Plot Column",
```
renkun-ken, line 126: Could we eliminate the space here? Looks a bit odd.

ElianHugh: I agree, the issue I was having is that array items don't use the same naming scheme as other configuration names, so `plotColumn` does not display as `Plot Column` in an array.

ElianHugh: Maybe the viewers could be instead grouped under `session.viewers.viewColumns.x`, which also leaves space for more configurations in the future

renkun-ken: I agree. Looks good to me if we just put them as ordinary settings like others.

ElianHugh: do you think camel case is better for the two-word options, e.g.

image

Or perhaps:

image

These would be seen in R as `session.viewers.viewColumn.camelCase` or `session.viewers.viewColumn.snake_case`

renkun-ken: `session.viewers.viewColumn.camelCase` is more consistent with other settings.

R/init.R
```diff
@@ -100,9 +100,99 @@ init_last <- function() {
     rm(".First.sys", envir = globalenv())
   )
 
+
   # Attach to vscode
   exports$.vsc.attach()
 
+  # Import VSC-R settings
```
renkun-ken, line 107: Some settings (e.g. `vsc.plot`, `vsv.use_httpgd`, `vsc.rstudioapi`) must be set before the `init.R` is sourced.

renkun-ken:
I move loading settings from `init.R` after attach to `vsc.R` before any option is used (e.g. `vsc.plot`, `vsc.use_httpgd`).

Also, each time `attach()` is called, the settings are loaded again so that some settings could take effect immediately (e.g. `vsc.str.max.level`, and those view column settings).

R/vsc.R
```diff
@@ -7,6 +7,91 @@ homedir <- Sys.getenv(
...
+    "session.viewers.viewColumn.viewer",
+    "session.viewers.viewColumn.pageViewer",
+    "session.viewers.viewColumn.view",
+    "session.viewers.viewColumn.helpPanel "
+  )
+
+  # settings.json can result in errors if read via fromJSON
+  vsc_settings <- suppressWarnings(
+    unlist(
+      jsonlite::fromJSON(paste(readLines(settings_file), collapse = ""))
```
renkun-ken, line 35: Why not use `jsonlite::read_json()`? Could you provide a case where "settings.json can result in errors if read via fromJSON"?

renkun-ken: OK. Let me refine the code a bit so that naming conventions are more consistent.

ElianHugh:
Thank you for updating the code @renkun-ken, it's working quite nicely now!

ManuelHentschel (approved):

ManuelHentschel:
Works nicely, thanks!

R/vsc.R
```diff
@@ -7,6 +7,64 @@ homedir <- Sys.getenv(
...
+
+  vsc_settings <- tryCatch(jsonlite::read_json(settings_file), error = function(e) {
+    message("Error occurs when reading VS Code settings: ", conditionMessage(e))
+  })
+
+  if (is.null(vsc_settings)) {
+    return(FALSE)
+  }
+
+  vsc_settings <- unlist(vsc_settings)
```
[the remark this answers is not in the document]
renkun-ken: It is already a nested list. What we need of `unlist` here is to convert `list(a=list(b=list(c="x")))` to `c(a.b.c = "x")`  so that it is easier to do the mapping below.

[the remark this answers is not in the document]
renkun-ken: If we need a list-valued item to work in R at some point, then I guess we need to avoid `unlist` here.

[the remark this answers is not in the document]
ManuelHentschel: Ah ok, thanks for the clarification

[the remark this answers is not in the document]
renkun-ken: I rewrite the `mapping` using `alist` with expressions so that `unlist` is no longer needed and list value could be supported too.

Issue #120, opened 2019-08-23
R tokenizing glitch with c() in function args


alexr00:
From @Hong-Revo in https://github.com/microsoft/vscode/issues/79559

- VSCode Version: 1.37.1 (system)
- OS Version: Windows 10 Pro 1903

There seems to be a (rather specific) bug in how character vectors are tokenized. When you have `c("a", "b", ...)` as a default function argument, the strings except for the first one are not treated as text:

![image](https://user-images.githubusercontent.com/11328890/63419106-8be0d300-c447-11e9-943c-9ccc12e542cf.png)

It only appears to happen when the `c` function is used; you can see that `list` doesn't have this problem. Similarly, if you have a character vector anywhere other than as a default function arg, it's coloured correctly.

Steps to reproduce

1. Type `f <- function(a=c("a", "b", "c")) {}` in an R editing window.
2. Note the colour of the strings.

The first string has the following scopes:
string.quoted.double.r
meta.function.parameters.r
meta.function.r
meta.function.r
source.r

The subsequent strings have the following scopes:
variable.parameter.r
meta.function.parameters.r
meta.function.r
meta.function.r
source.r

PR #751, opened 2021-08-13
Fix  issues with c() in function args

### LINKED ISSUE #713: Syntax highlighting fails when there's a vector argument
I came here by following the link in `[...]/code/resources/app/extensions/r/syntaxes/r.tmLanguage.json` so I hope I'm in the right place.

**Describe the bug**

The following code breaks the syntax highlighting; everything after the last quote is considered part of an (unfinished) string

```r
fun <- function(a, c = c("a", "()"), d) {
    function body, not a quoted string
}
```

syntax

It seems to view the close bracket within the quoted string as the end of the function args.

**To Reproduce**

1. Open an R file
2. Copy or type that code
3. Observe the results

## Do you want to fix by self? (We hope your help!)

I have tried to, but I'm not familiar with textmate grammars... so... maybe? I gather I've got to mess with function-parameters so that it no longer views the close bracket as the end of the function args, probably by allowing arguments to have default values that are the evaluation of function calls

## (If yes,) what kind of help do you want? (e.g. Which file should I fix, Survey (related documents)

**Expected behavior**
Pretty clear I think

**Screenshots**
Above

- OS: Linux
- VSCode Version: [e.g. 1.58.2]
- not important
- also unimportant, but I'm testing with the current language definition in this repo

ElianHugh:
Uses the match pattern for functions from the textmate R grammar in place of the previous function pattern. I don't have experience with using textmate grammar, but from testing on my code I haven't seen any issues thus far. If anyone is more knowledgeable with textmate and would like to chime in, that would be greatly appreciated :)

The grammar's [repo](https://github.com/textmate/r.tmbundle) indicates that there shouldn't be any licensing issues either.

# What problem did you solve?
Fixes #713 and incidentally #120.

The following should not have issues with text decoration when using the PR.

```r
# issue 713
fun <- function(a, c = c("a", "()"), d) {
    function body, not a quoted string
}

# issue 120
f <- function(a=c("a", "b", "c")) {}

```

### DIFF
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -1,9 +1,4 @@
 {
-    "information_for_contributors": [
-        "This file has been converted from https://github.com/randy3k/R-Box/blob/master/syntax/R%20Extended.sublime-syntax",
-        "If you want to provide a fix or improvement, please create a pull request against the original repository.",
-        "Once accepted there, we are happy to receive an update request."
-    ],
     "fileTypes": [
         "R",
         "r",
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -340,8 +340,8 @@
         "function-declarations": {
             "patterns": [
                 {
-                    "begin": "^\\s*([a-zA-Z._][\\w.:]*)\\s*(<<?-|=)\\s*(?=function\\s*\\()",
-                    "beginCaptures": {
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
Works nicely! Thanks!

eitsupi:
As I wrote in the comments https://github.com/REditorSupport/vscode-R/issues/713#issuecomment-885423686, do you need to remove the link to the R-Box which is already archived?
Anyone who comes to this repository looking at the vscode source code can be confused.

ElianHugh:
Thanks for the reminder! I've removed the statement, as modifications to the syntax should be made here anyway

PR #765, opened 2021-08-26
R Markdown Enhancements (Knit Manager)

ElianHugh:
Enhances rendering R Markdown documents. Specifically: 

- background knitting (with progress bar)
- set knit working directory
- Smart knit button, use custom knit functions + rmarkdown::render_site where appropriate

## Screenshots

image

image

image

#### Custom Knit

image

image

#### R Markdown Site

image

image


renkun-ken:
Always fail to start debugging. The following error occurs on extension activation:

```
Activating extension 'Ikuyadeu.r' failed: rmarkdown.RMarkdownPreviewManager is not a constructor.
```

src/rmarkdown/manager.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export abstract class RMarkdownManager {
+	protected rPath: string;
+	protected rMarkdownOutput: vscode.OutputChannel = rMarkdownOutput;
+	// uri that are in the process of knitting
+	// so that we can't spam the knit/preview button
+	protected busyUriStore: Set<string> = new Set<string>();
+
+	public async init(): Promise<void> {
+		this.rPath = await util.getRpath(true);
```
renkun-ken, line 31: If `rPath` is determined in `init()`, then chaning R paths in settings will not have effect before restart. Maybe we should call `util.getRpath()` each time we use it?

ElianHugh:
> 
> 
> Always fail to start debugging. The following error occurs on extension activation:
> 
> ```
> Activating extension 'Ikuyadeu.r' failed: rmarkdown.RMarkdownPreviewManager is not a constructor.
> ```

I suspect this is some kind of bug with TypeScript, as I had this happen until I deleted my `out` folder. It could also be the version of TS installed? I've updated to 4.3.5. In any case, I can't see why a wildstar import would be invalid for classes, but explicit exports would be ok

renkun-ken:
> as I had this happen until I deleted my out folder

This actually works nicely! Thanks!

src/rmarkdown/knit.ts
```diff
@@ -149,6 +149,7 @@ export class RMarkdownKnitManager extends RMarkdownManager {
 
 	// alters the working directory for evaluating chunks
 	public setKnitDir(): void {
+		const currentDocumentWorkspace = vscode.workspace.getWorkspaceFolder(vscode.window.activeTextEditor.document.uri);
```
renkun-ken, line 152: If the document is untitled but set to `rmd` language, then the workspace folder is `undefined`, if I remember correctly.

ElianHugh: Ah, yes that would cause some trouble. Given that a document can't be knit without being saved, I think it would make sense to have the following fallbacks:

Workspace folder > document folder > NULL

src/rmarkdown/knit.ts
```diff
@@ -149,20 +149,26 @@ export class RMarkdownKnitManager extends RMarkdownManager {
 
 	// alters the working directory for evaluating chunks
 	public setKnitDir(): void {
-		const currentDocumentWorkspace = vscode.workspace.getWorkspaceFolder(vscode.window.activeTextEditor.document.uri);
+		let currentDocumentWorkspacePath: string = null;
```
renkun-ken, line 152: I should have made it more clear: untitled document and document outside any workspace folder will return `undefined` workspace folder. We should handle both cases separately.

renkun-ken: Maybe we could provide QuickPickItem accordingly:

* If it has a workspace folder, then add a QuickPickItem for it.
* If it has a filesystem folder, then add a QuickPickItem for it.

renkun-ken (approved):
Works nicely!

LGTM

ManuelHentschel:
On windows, if I set the Knit directory to the document directory, I get an error:
```
[VSC-R] test.Rmd process started
Erreur : '\\' inattendu(e) dans "knitr::opts_knit[['set']](root.dir = c:\\"
Ex�cution arr�t�e

[VSC-R] test.Rmd process exited with exit code 1
```
In the workspace root this error does not occur.

ElianHugh:
> On windows, if I set the Knit directory to the document directory, I get an error:
> 
> ```
> [VSC-R] test.Rmd process started
> Erreur : '\\' inattendu(e) dans "knitr::opts_knit[['set']](root.dir = c:\\"
> Ex�cution arr�t�e
> 
> [VSC-R] test.Rmd process exited with exit code 1
> ```
> 
> In the workspace root this error does not occur.

Argh, the path differences are the bane of my existence!! Thanks for catching that, *hopefully* the most recent commit fixes this?

package.json
```diff
@@ -1262,7 +1299,37 @@
...
+          "markdownDescription": "Should the R Markdown output channel be focused when knitting?\n\nRequires `#r.rmarkdown.knit.useBackgroundProcess#` to be set to `true`."
+        },
+        "r.rmarkdown.knit.openOutputFile": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Should the output file be opened automatically when using knit?\n\nRequires `#r.rmarkdown.knit.useBackgroundProcess#` to be set to `true`."
+        },
+        "r.rmarkdown.knit.defaults.knitWorkingDirectory": {
+          "type": "string",
+          "default": "document",
```
ManuelHentschel, line 1321: This should probably be `document directory`

src/rmarkdown/knit.ts
```diff
@@ -0,0 +1,250 @@
+import * as util from '../util';
+import * as vscode from 'vscode';
+import * as fs from 'fs-extra';
+import path = require('path');
+import yaml = require('js-yaml');
+
+import { RMarkdownManager, KnitWorkingDirectory } from './manager';
+import { runTextInTerm } from '../rTerminal';
+
+export let knitDir: string = util.config().get<string>('rmarkdown.knit.defaults.knitWorkingDirectory') ?? undefined;
```
ManuelHentschel, line 10: Preferrably this should be typed as `KnitWorkingDirectory` (to make clear that it does not contain e.g. a literal path)

src/rmarkdown/knit.ts
```diff
@@ -0,0 +1,250 @@
...
+
+		if (items.length > 0) {
+			void vscode.window.showQuickPick(
+				items,
+				{
+					title: 'Set knit working directory',
+					canPickMany: false,
+					onDidSelectItem: (item: IKnitQuickPickItem) => {
+						knitDir = item.value;
+					}
```
ManuelHentschel, line 192: `onDidSelectItem` triggers everytime you change the selection in the quickpick, e.g. using the arrow keys.
This leads to wrong behaviour, e.g. when the user highlights an option and then presses escape.

A better way is to use the return value of `showQuickPick` (which needs to be checked for `undefined` in case the user presses escape)

src/rmarkdown/knit.ts
```diff
@@ -0,0 +1,250 @@
...
+				}
+			);
+		} else {
+			void vscode.window.showInformationMessage('Cannot set knit directory for untitled documents.');
+		}
+
```
ManuelHentschel, line 198: If the user changed the knit directory, it might be useful to trigger a refresh here

src/rmarkdown/index.ts
```diff
@@ -6,6 +6,10 @@ import {
 import { runChunksInTerm } from '../rTerminal';
 import { config } from '../util';
```
ManuelHentschel, line 8: I can't add the comment at the corresponding lines (since it's not actually part of this PR), but I'd definitely prefer to use a `*` import for `vscode` here

ManuelHentschel:
> Argh, the path differences are the bane of my existence!! Thanks for catching that, _hopefully_ the most recent commit fixes this?

Can confirm that I'm not getting the error anymore

ManuelHentschel (approved):
I'm not very experienced with Rmd, but the simple examples I tried worked nicely.

Thanks for the nice work!

PR #773, opened 2021-09-06
(Refactoring) Simplify RMD child process disposal

ElianHugh:
Minor refactor to the R Markdown preview/knit processes, cleaning up the disposal of child processes. 

Changes:
* Contributes a helper function for creating disposables from objects
* Simplifies disposal of child processes
* Do not print lines to output when the process has already been terminated

Both knitting and previewing should function as before

### DIFF
--- a/src/rmarkdown/knit.ts
+++ b/src/rmarkdown/knit.ts
@@ -26,12 +26,12 @@ interface IYamlFrontmatter {
 }
 
 export class RMarkdownKnitManager extends RMarkdownManager {
-	private async renderDocument(rPath: string, docPath: string, docName: string, yamlParams: IYamlFrontmatter, outputFormat?: string): Promise<DisposableProcess> {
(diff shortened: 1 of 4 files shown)


renkun-ken:
I try the following Rmd

````
```{r}
x <- 1
Sys.sleep(100)
```
````

The kniting will take 100s to finish. When I click "Cancel", the progress window is closed but the R process is still running until it is finished. Shouldn't the process be killed once it is cancelled?

**Comment by ElianHugh** [comment]
Does the process not receive an exit signal in your output stream? This is what I see when I try it:

image

**Comment by renkun-ken** [comment]
It does receive a `SIGKILL` signal

```

[VSC-R] terminating R process
[VSC-R] doc4.Rmd process exited from signal 'SIGKILL'
```

but I'm monitoring all R processes and the process does not actually stops.

ElianHugh:
I believe this is caused by the use of exec rather than spawn, as the shell appears to not close until completion

ElianHugh:
Last push appears to have fixed the termination issue on my end

renkun-ken:
Thanks for the quick fix. It works on my end too.

renkun-ken (approved):
LGTM

PR #778, opened 2021-09-08
Add object length limit

### LINKED ISSUE #174: Reduce overhead of session watcher task callback
Currently, we use `utils::capture.output(utils::str(obj, max.level = 0, give.attr = FALSE))` to get the string representation of each object in global environment. The overhead will become more significant as there are more symbols in it.

Suppose the global environment contains as many symbols as base environment (more than 1200 visible symbols), the overhead of getting object metadata will be more than 300ms, i.e., on each user input, there will be an extra 300ms to wait for the metadata of objects to be gathered.

To reduce the overhead, we could avoid using `capture.output` and `str` and instead create a simple string that contains simple information such as class, type, length, dim to represent the object.

```r
system.time(objs <- eapply(baseenv(), function(obj) {
  utils::capture.output(utils::str(obj, max.level = 0, give.attr = FALSE))
}))
#>    user  system elapsed 
#>   0.335   0.004   0.339
system.time(objs <- eapply(baseenv(), function(obj) {
  list(
    type = typeof(obj),
    length = length(obj),
    class = class(obj)
  )
}))
#>    user  system elapsed 
#>   0.013   0.001   0.013
```

If the time of writing files are included,

```r
system.time({
  objs <- eapply(baseenv(), function(obj) {
    list(
      type = typeof(obj),
      length = length(obj),
      class = class(obj),
      str = trimws(utils::capture.output(utils::str(obj, max.level = 0, give.attr = FALSE))),
      names = names(obj)
    )
  })
  jsonlite::write_json(objs, "test.json", auto_unbox = TRUE, pretty = FALSE)
})
#>    user  system elapsed 
#>   0.557   0.011   0.570
system.time({
  objs <- eapply(baseenv(), function(obj) {
    list(
      type = typeof(obj),
      length = length(obj),
      class = class(obj),
      names = names(obj)
    )
  })
  jsonlite::write_json(objs, "test.json", auto_unbox = TRUE, pretty = FALSE)
})
#>    user  system elapsed 
#>   0.113   0.000   0.113
```

renkun-ken:
# What problem did you solve?

Closes #174 

This PR introduces a new setting `r.session.objectLengthLimit` corresponding to R option `vsc.object_length_limit` to limit the length of objects to provide names in session symbol completion, hover, and workspace viewer item.

Then large named objects with many elements will not be called `names()`. The overhead of calling `names()` (on environment) and writing the names to `globalenv.json` should be reduced.

For the following code, before this PR, the delay will be more and more significant if more objects with many elements are created in the global environment.

```r
n1 <- seq_len(1000)
lst1 <- lapply(n1, function(i) i)
names(lst1) <- paste0("name", n1)

1 + 1 # no delay

n2 <- seq_len(5000)
lst2 <- lapply(n2, function(i) i)
names(lst2) <- paste0("name", n2)

1 + 1 # slight delay

n3 <- seq_len(100000)
lst3 <- lapply(n3, function(i) i)
names(lst3) <- paste0("name", n3)

1 + 1 # significant delay
```

With this PR being effective, creating `lst2` and `lst3` should no longer cause a permanent delay on all subsequent calculations. One should also observe that the auto-completion works with `lst1` but not `lst2` or `lst3` due to object length limit, and the workspace viewer item of `lst1` should show object details while `lst2` and `lst3` do not if `r.session.levelOfObjectDetail` is set to `Detailed`. Adjusting `r.session.objectLengthLimit` should work accordingly.

### DIFF
--- a/R/vsc.R
+++ b/R/vsc.R
@@ -134,6 +134,8 @@ inspect_env <- function(env, cache) {
   is_promise <- rlang::env_binding_are_lazy(env, all_names)
   is_active <- rlang::env_binding_are_active(env, all_names)
   show_object_size <- getOption("vsc.show_object_size", FALSE)
+  object_length_limit <- getOption("vsc.object_length_limit", 2000)
+  str_max_level <- getOption("vsc.str.max.level", 0)
   objs <- lapply(all_names, function(name) {
     if (is_promise[[name]]) {
       info <- list(
@@ -171,12 +173,10 @@ inspect_env <- function(env, cache) {
         info$size <- scalar(cobj$size)
       }
 
-      length_limit <- getOption("vsc.object_length_limit", 2000)
-      if (length(obj) > length_limit) {
+      if (length(obj) > object_length_limit) {
         info$str <- scalar(trimws(capture_str(obj, 0)))
       } else {
-        max_level <- getOption("vsc.str.max.level", 0)
-        info$str <- scalar(trimws(capture_str(obj, max_level)))
+        info$str <- scalar(trimws(capture_str(obj, str_max_level)))
         obj_names <- if (is.object(obj)) {
           .DollarNames(obj, pattern = "")
         } else if (is.recursive(obj)) {
--- a/R/vsc.R
+++ b/R/vsc.R
@@ -20,6 +20,7 @@ load_settings <- function() {
     vsc.show_object_size = workspaceViewer$showObjectSize,
     vsc.rstudioapi = session$emulateRStudioAPI,
     vsc.str.max.level = session$levelOfObjectDetail,
+    vsc.object_length_limit = session$objectLengthLimit,
     vsc.globalenv = session$watchGlobalEnvironment,
     vsc.plot = session$viewers$viewColumn$plot,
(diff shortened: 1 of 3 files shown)


ElianHugh:
Thanks for the quick fix! I've had a look over the commit and it looks good. I'll have a chance to try the PR in an hour or so

ElianHugh:
Thank you for the good work, it all works quite nicely! The only (very minor) issue I've had is the console complaining about undefined properties:

```js
Error Cannot read property 'map' of undefined: TypeError: Cannot read property 'map' of undefined
	at LiveCompletionItemProvider.provideCompletionItems (/home/elian/Documents/Coding/vscode-R/out/completions.js:111:22)
	at $.provideCompletionItems (/var/lib/snapd/snap/code/73/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:176938)
```

renkun-ken:
I made some changes to the completions to achieve the following:

* The errors should be gone since checks are added.
* The order of object elements are preserved.

ElianHugh (approved):
LGTM! Thanks again!

PR #780, opened 2021-09-09
Write NA as string

### LINKED ISSUE #776: Inf is unable to be displayed in the vscode-r table window
If you have a R dataframe `dat` in vscode-r, you can display the dataframe in an pop-up/separated window by typing `View(dat)`. However, I found that the table window is unable to display the infinity `Inf` normally.

Figure 1. Inf not shown in displayed table

Figure 2. We can see the Inf in the terminal outputs

**Environment:**

- OS: Windows 10 x64
- VSCode Version: 1.60.0
- R Version: 4.1.0
- vscode-R version: 2.2.0

renkun-ken:
# What problem did you solve?

Closes #776 

Since JSON does not support `NA`, `NaN`, `Inf` values, they can only be written as strings in the following way:

```r
jsonlite::write_json(x, file, na = "string", null = "null", auto_unbox = TRUE)
```

One remaining issue is that these values will not work properly when filtering is used in the data viewer, i.e. if one filters a column by "Greater than 0", `Inf` is not regarded as Inf > 0 but ignored.

## (If you have)Screenshot

image

The following code should open a data viewer where `Inf` and `NA` values are correctly displayed in the data grid.

```r
dt <- data.frame(
  id = 1:6,
  x = c(1, 2, NA, 3, Inf, -Inf),
  y = c("A", "B", NA, NA, NA, "D")
)

View(dt)
```

### DIFF
--- a/R/vsc.R
+++ b/R/vsc.R
@@ -418,13 +418,13 @@ if (show_view) {
     if (is.data.frame(x) || is.matrix(x)) {
       data <- dataview_table(x)
       file <- tempfile(tmpdir = tempdir, fileext = ".json")
-      jsonlite::write_json(data, file, auto_unbox = TRUE)
+      jsonlite::write_json(data, file, na = "string", null = "null", auto_unbox = TRUE)
       request("dataview", source = "table", type = "json",
         title = title, file = file, viewer = viewer, uuid = uuid)
     } else if (is.list(x)) {
       tryCatch({
         file <- tempfile(tmpdir = tempdir, fileext = ".json")
-        jsonlite::write_json(x, file, auto_unbox = TRUE)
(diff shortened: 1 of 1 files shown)


ElianHugh (approved):
LGTM! In the future we could probably look at comparators for sorting the values properly

renkun-ken:
It looks like JavaScript has support for [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/NaN) and [infinite values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/POSITIVE_INFINITY) but JSON has no way to represent them. We are only using jsonlite to write the structure to be inserted as part of the HTML, not really reading it as json anyway. In the future, we might have a better way to simply produce the correct JavaScript object directly.

PR #783, opened 2021-09-10
Use R files for background process

### LINKED ISSUE #781: Consider storing background-process R commands in files
Many of VSCode-R's features use a background R process to perform tasks (e.g., help providers, previewing and knitting documents). We currently define the calls to R in-line, e.g.

```ts
      const cmd = (
            `knitr::opts_knit[['set']](root.dir = ${knitWorkingDirText});` +
            `cat('${lim}', rmarkdown::render(` +
            `'${filePath.replace(/\\/g, '/')}',` +
            `output_format = rmarkdown::html_document(),` +
            `output_file = '${outputFile.replace(/\\/g, '/')}',` +
            `intermediates_dir = '${tmpDir().replace(/\\/g, '/')}'), '${lim}',` +
            `sep='')`
        );
```

I think that long-term this could get hard to maintain. Also, it can be difficult to wrap your head around all the quoting!

I think that it might be neater and easier to maintain if we instead stuck these calls in their own R files, and then sourced them. For instance:

```ts
cp.spawn(`R -f test.R`) 
```

We can use the [`commandArgs`](https://rdrr.io/r/base/commandArgs.html) method to pass values to a given file. E.g.

```r
# test.r
print(commandArgs(trailingOnly = TRUE))
```

```bash
R --silent --slave --no-save -f test.r --args "arg1" "arg2" "arg3"
#> [1] "arg1" "arg2" "arg3"
```

ElianHugh:
Resolves #781 

- Moves R process commands from inline text to R scripts
- Use environment variables to pass values to the scripts (as per @renkun-ken's suggestion)
- No user-facing changes

## Why?

In general, easier to maintain:

* Easier to read
* Easier to test!
* Less pain re: quoting 
* More extensible

## Changes
* Move R-related session files to `R/session`
* Move `getAliases.R` to `R/help`
* New file `helpServer.R` in `R/help`
* New file `knit.R` in `R/rmarkdown`
* New file `preview.R` in `R/rmarkdown`

---

* I looked at other uses of childProcesses, but I'm not 100% at what point we should use a file vs. inline text.
* Tested on WSL (Ubuntu) and Arch Linux


renkun-ken:
Should we also make it for the R help server at

https://github.com/REditorSupport/vscode-R/blob/4d6a6635d0d442dc1c410f108de99c99a09c1924/src/helpViewer/helpProvider.ts#L42-L45

ElianHugh:
> 
> 
> Should we also make it for the R help server at
> 
> https://github.com/REditorSupport/vscode-R/blob/4d6a6635d0d442dc1c410f108de99c99a09c1924/src/helpViewer/helpProvider.ts#L42-L45

I wasn't sure at what point/complexity we should create files. Do we think all background processes?

renkun-ken:
I think we should put a script if the command line has multiple expressions or needs quoting.

renkun-ken:
Also, do we really need the `background` folder? Will the following look better or more confusing?

* `R/help/getAliases.R`
* `R/help/server.R`
* `R/rmarkdown/knit.R`
* `R/rmarkdown/preview.R`

ElianHugh:
> 
> 
> Also, do we really need the `background` folder? Will the following look better or more confusing?
> 
>     * `R/help/getAliases.R`
> 
>     * `R/help/server.R`
> 
>     * `R/rmarkdown/knit.R`
> 
>     * `R/rmarkdown/preview.R`

I was thinking the background folder, because I wanted to distinguish it from the session-watcher related files. Would it make more sense to have the following?

* `R/rmarkdown/`
* `R/help/`
* `R/session/`

renkun-ken:
> Would it make more sense to have the following?

Makes good sense.

R/background/rmarkdown/knit.R
```diff
@@ -0,0 +1,22 @@
+# requires rmarkdown package to run (and knitr)
+if (!require(rmarkdown, quietly = TRUE)) {
```
renkun-ken, line 2: Should we use `requireNamespace("rmarkdown", quietly = TRUE)` here or we do want to attach the package?

ElianHugh: Thanks for catching that!

ElianHugh:
I've moved the session files to `R/session`! Everything should work as before -- just double checking if I caught everything

renkun-ken (approved):
Everything works well.
LGTM

PR #785, opened 2021-09-10
Respect preview output format 

### LINKED ISSUE #779: Cross-reference does not work in preview when specify output as `bookdown::html_document2`
Cross-referencing for images or tables via `\@ref(tab:xxx)` looks like working only under, say,  `bookdown::html_document2`. 
However, previewing the output by clicking "Open Preview to the Site" does not render cross-reference properly. 

For example:
```
---
title: "R Notebook"
output:
  bookdown::html_document2: default
---
{r table1}
knitr::kable(mtcars[1:5, 1:5], caption = "The mtcars data.")

`mtcar` data is shown in Table \@ref(tab:table1).

```
The vscode-R preview results in 
image

The correct one with `rmarkdown::render` or `Ctrl + Shift + K` results in
image

The current keybinding to open preview is `Ctrl + Shift + V`, is it possible to change it to render and open the preview with `Ctrl + Shift + K`, just like in RStudio?

ElianHugh:
Closes #779 

At the moment, previewing R Markdown documents forces the `rmarkdown::html_document()` format. However, this overwrites all output formats, even if they can result in a .html file. 

This PR uses the object `pandoc$to` from `rmarkdown:::create_output_format` to check if the document's output format == html. This way, we only override the output format if `pandoc$to` does not equal html.

## Screenshot

image

The following renders incorrectly on master, which is fixed in this PR:

````r
---
title: "R Notebook"
output:
  bookdown::html_document2: default
---

```{r table1}
knitr::kable(mtcars[1:5, 1:5], caption = "The mtcars data.")
```

`mtcar` data is shown in Table \@ref(tab:table1).
````

### DIFF
--- a/R/rmarkdown/knit.R
+++ b/R/rmarkdown/knit.R
@@ -1,5 +1,5 @@
 # requires rmarkdown package to run (and knitr)
-if (!requireNamespace(rmarkdown, quietly = TRUE)) {
+if (!requireNamespace("rmarkdown", quietly = TRUE)) {
     stop("Knitting requires the {rmarkdown} package.")
 }
 
--- a/R/rmarkdown/preview.R
+++ b/R/rmarkdown/preview.R
@@ -1,15 +1,31 @@
 # requires rmarkdown package to run (and knitr)
-if (!requireNamespace(rmarkdown, quietly = TRUE)) {
+if (!requireNamespace("rmarkdown", quietly = TRUE)) {
     stop("Previewing documents requires the {rmarkdown} package.")
 }
 
 # get values from extension-set env values
-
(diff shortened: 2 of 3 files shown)


renkun-ken:
The rendering works now.

But it seems the link to table 1 does not work: it will open in external browser and go to a non-working URL.

ElianHugh:
Thanks for picking that up - it looks like this is something to do with how webviews convert links, as the anchors work in external browsers.

renkun-ken:
The anchor tags work nicely now.

renkun-ken (approved):
LGTM

PR #771, opened 2021-09-03
Use unsafe-inline for script-src

renkun-ken:
# What problem did you solve?

Closes #770

The current content security policy seems too strict for flextable to work since it contains inline scripts. Relaxing the CSP will make it work as expected.

## (If you have)Screenshot

image

## (If you do not have screenshot) How can I check this pull request?

```r
flextable::flextable(iris)
```

PR #805, opened 2021-09-29
Reenable 'unsafe-eval' in script-src CSP

juba:
# What problem did you solve?

PR #771 replaced `unsafe-eval` with `unsafe-inline` in script-src CSP to make `flextable` work, but disabling `unsafe-eval` broke some of my package code which uses `htmlwidgets::JS`.

I think that putting both `unsafe-eval` and `unsafe-inline` in CSP allows to make both code work :

```
script-src https: data: filesystem: 'unsafe-inline' 'unsafe-eval';
```

The following code should produce a line plot with a red "Hello, world!" text.

```r
remotes::install_github("juba/obsplot")
library(obsplot)
library(htmlwidgets)
obsplot(aapl) |>
  mark_line(x = "Date", y = "Close") |>
  mark_function(JS("() => svg`<text x=20% y=20% fill=red>Hello, world!</text>`"))
```

Many thanks !

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -632,7 +632,7 @@ export async function getWebviewHtml(webview: Webview, file: string, title: stri
         upgrade-insecure-requests;
         default-src https: data: filesystem:;
         style-src https: data: filesystem: 'unsafe-inline';
-        script-src https: data: filesystem: 'unsafe-inline';
+        script-src https: data: filesystem: 'unsafe-inline' 'unsafe-eval';
     `;
 
     return `


renkun-ken:
Thanks for reporting and fixing it. I'm no expert on CSP, but it looks good according to <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src>.

renkun-ken (approved):
LGTM

Issue #661, opened 2021-06-04
`R: Open help for selection` fails for some functions


jolars:
**Describe the bug**

`r.helpPanel.openForSelection` (R: Open help for selection) sometimes fail to work although documentation-on-hovering does.

See the example below where I cannot use the command to open the documentation for `glmnet::glmnet()`, although the documentation shows when hovering (via the language server), and also opens up correctly when I execute `?glmnet`. This also happens for instance with the function `bag()` from the caret package, so this is not a problem only with functions named exactly as the package. For the built-in function `lm()` and also for other functions such as `ggplo2::geom_point()` the command works fine.

![help-shortcut-fail](https://user-images.githubusercontent.com/13087841/120826866-2a75ef80-c55b-11eb-85c1-611134c46c7d.gif)

**To Reproduce**

1. Open an `.R` file.
2. Write the name of a function in the file.
3. Highlight the function name and select `R: Open help for selection` from the command tool box.

## Do you want to fix by self? (We hope your help!)

Yes, I'll take a look at it.

`setting.json`:

```json
{
    "r.rterm.linux": <path-to-radian>,
    "r.sessionWatcher": true,
    "r.rmarkdown.codeLensCommands": [
        "r.runCurrentChunk",
        "r.runAboveChunks",
        "r.runAllChunks"
    ],
    "r.bracketedPaste": true
}
```

**Expected behavior**

I expected the documentation to open up.

**Environment (please complete the following information):**

- OS: Linux
- VSCode Version: 1.56.2
- R Version: 4.0.5
- vscode-R version: 1.6.8

renkun-ken:
I cannot reproduce what you demonstrate in your screencast. Everything works well here. Not sure what caused this. @ManuelHentschel any idea?

eitsupi:
I had the same experience the other day. (Sorry I haven't reported it)
At that time, the problem occurred with a package installed from a different repository than the one set as default.

ManuelHentschel:
Does calling "Clear Cached Index Files & Restart Help Server" in the R sidebar do anything?

eitsupi:
> Does calling "Clear Cached Index Files & Restart Help Server" in the R sidebar do anything?

In my case, this fixed the problem!

jolars:
> Does calling "Clear Cached Index Files & Restart Help Server" in the R sidebar do anything?

Yup, this works for me too! Thanks!

ManuelHentschel:
[...]

PR #819, opened 2021-10-11
Change help cache default

ManuelHentschel:
re #661 

Changes the default behavior for the caching of parsed help index files between sessions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1349,7 +1349,7 @@
             "Store on a per workspace basis",
             "Store globally"
           ],
-          "default": "Global"
+          "default": "None"
         },
         "r.helpPanel.rpath": {
           "type": "string",


renkun-ken:
I remove `glmnet` package, reload window, and hover/open help for `glmnet::glmnet`, and the help does not exist.

Then I install `glmnet`, the languageserver could instantly provide the documentation on hover but vscode-R help hover and open help for selection still does not work in this case.

ManuelHentschel:
> Then I install `glmnet`, the languageserver could instantly provide the documentation on hover but vscode-R help hover and open help for selection still does not work in this case.

- Does reloading vscode fix this?
- Is the behavior the same if you install `glmnet` using the command `Install CRAN Package`?

renkun-ken:
Thanks for the details.

PR #823, opened 2021-10-16
Use httpgd NPM package

nx10:
# What problem did you solve?

This is a first WIP attempt at replacing the client side API and connection code with the new [httpgd NPM package](https://www.npmjs.com/package/httpgd) to unify and simplify development.

This is also discussed in https://github.com/nx10/httpgd/issues/79 

**This PR is a very early prototype**, basic functionality is implemented, but there are still a lot of bugs. This is intended as a starting point for further development with @ManuelHentschel 

Changes:

* Replaced type definitions with imported types
* Replaced API and connection handling code
* Removed client side zooming (this is now properly handled server side)
* Removed unique ID hack (portable SVG renderer is now used)

Bugs:

* ~~There seems to be a problem when more than one plot is drawn (duplicated invisible plots in history)~~
* ~~Plot view very small when opened~~
* ~~Text border is colored instead of the text itself when using theming~~


ManuelHentschel:
Thanks for the PR! I started working on the "Export plot as..." functionality but got stuck at the point where I actually need to get the content of the exported plot ([here](https://github.com/ManuelHentschel/vscode-R/blob/518d884f4fd2a0dea6aed0f0f59592349c22116d/src/plotViewer/index.ts#L774)).


I've seen that you do this kind of "manually" [here](https://github.com/ManuelHentschel/vscode-R/blob/518d884f4fd2a0dea6aed0f0f59592349c22116d/src/plotViewer/index.ts#L615-L629). Should I also do something like that, or is this supposed to be part of the httpgd-js api?

ManuelHentschel:
The latest push depends on the httpgd version from https://github.com/nx10/httpgd-js/pull/1. To test this, I temporarily removed the httpgd dependency and use [yarn link](https://classic.yarnpkg.com/lang/en/docs/cli/link/) to import a local version of the package.

nx10:
I just pushed a commit that streams the plot directly into the exported file. This should be more memory efficient and might be a bit faster. There seems to be a problem with eslint not recognizing the `Response` object correctly (eslint thinks this is the browser Response object and not the node-fetch one. This might be a `cross-fetch` problem or config problem in `vscode-r` I am not sure.

> > ```
> > * Text border is colored instead of the text itself when using theming
> > ```
> 
> 
> Can you please give an example for this?

Nevermind, I think I was testing with an old version of vscode-R when I noticed this. Text looks good now.

Good job on the other two fixes as well.

src/plotViewer/index.ts
```diff
@@ -763,7 +763,29 @@ export class HttpgdViewer implements IHttpgdViewer {
         }
         // make sure outFile is valid or return:
         if (!outFile) {
+            // Suggest a file extension:
+            const renderer = this.api.getRenderers().find(r => r.id === rendererId);
+            
+            const defaultUri = vscode.workspace.workspaceFolders[0]?.uri;
```
renkun-ken, line 769: We might need to keep in mind that we support multi-root workspace. So it would be best if we could actually use a proper workspace folder that makes sense here. Need to cover the case of no workspace folder, single workspace folder, multi-root workspace.

Maybe we should use the workspace folder of the active document?

ManuelHentschel: Since the export dialogue will usually be triggered while the plot webview is active, the active document is `undefined` when this code runs.

Alternatively, we could start with the current `defaultUri`, but remember the save location and use it as default the next time the save dialogue is opened?

renkun-ken: Each plot viewer is started from an interactive R session. I don't know if each plot viewer remembers the working directory of that session.

Maybe we should use the working directory of the R session that starts the plot viewer as the default folder?

If the plot viewer do not know it then we might use the working directory of the currently attached session?

ManuelHentschel: > Each plot viewer is started from an interactive R session. 

An external httpgd-url can also be started using the commadn `r.plot.openUrl`. Using the wd of the connected R process might be the cleanest solution here. @nx10 would you consider adding the wd to the httpgd api or are there problems with this approach?

nx10: Do you mean adding an API that returns the working directory path of the R session? I am not sure if this is the best approach as the server runs not necessarily on the same system as the client (remote connections). 
I just added the `/info` / `hgd_info()` API to both httpgd and httpgd-js, which reports version information (which may be used for feature detection) and a (universally) unique server/device ID. Could the ID be used to work around this problem?

ManuelHentschel: > Do you mean adding an API that returns the working directory path of the R session?

Yes. But I get the concern that the R session might be running behind a remote connection, and that it wouldn't make sense to reveal a file path in that case.

But then I would rather not try to guess a directory based on the server/device ID or wd of the attached R process (this might get messy, and choosing one of the workspace folders is at least a very predictable behaviour). Remembering the last used export path also feels quite natural as this is done in many other programs as well.

renkun-ken: One question is that downloading the plot will put the file in the client's local filesystem right? Under remote development, the workspace folder is still remote rather than local.

ManuelHentschel: > One question is that downloading the plot will put the file in the client's local filesystem right?

Not sure what you mean here. Exporting the plot will save it to the remote file system, as should be the expected behavior.

renkun-ken: I thought "Export to PNG/SVG/..." always saves the image to the client file system, at least it does in the browser.

I would be totally fine to start with the first workspace folder and remember user choice.

ManuelHentschel:
Anything else that needs to be implemented/fixed, @nx10 @renkun-ken ?

nx10:
I think the restructuring went very well, thank you for your work. 

At some point if somebody volunteered to add an article for the httpgd documentation would be great: https://nx10.github.io/httpgd/articles/b01_vscode.html

renkun-ken:
Thanks for the nice work!

As I test the httpgd plot viewer commands, it looks like "Toggle Preview Plot Layout" will randomly make the previews disappear rather than rearrange the layout. Not sure if you could reproduce it?

ManuelHentschel:
> As I test the httpgd plot viewer commands, it looks like "Toggle Preview Plot Layout" will randomly make the previews disappear rather than rearrange the layout. Not sure if you could reproduce it?

That command cycles through three different layouts: horizontal scrolling, multi-row and hidden (i.e. the previews disappear)

renkun-ken:
>That command cycles through three different layouts: horizontal scrolling, multi-row and hidden (i.e. the previews disappear)

OK. The hidden one seems not quite useful for me, as it is almost the same with full window mode.

ManuelHentschel:
> OK. The hidden one seems not quite useful for me, as it is almost the same with full window mode.

I think this option was requested at some point, and it might be useful e.g. if your webview pane is rather tall and narrow, but you still don't want to see all your previous plots.

renkun-ken (approved):
LGTM

PR #851, opened 2021-10-31
Use new terminal API

ManuelHentschel:
This PR adds a terminal profile for R using the API discussed e.g. in https://github.com/microsoft/vscode/issues/120369.

The existing "Create R Terminal" command is not affected. Creating a terminal using the terminal profile should behave the same as using the command.

The API to add a custom icon for the terminal profile is still being discussed (https://github.com/microsoft/vscode/issues/127607), so I just left the default icon for now.

**To test:**
Call the command `Terminal: Create new terminal (with profile)` and select `R Terminal` or use the dropdown next to the new terminal button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,10 +61,19 @@
     "onCommand:r.helpPanel.back",
     "onCommand:r.helpPanel.forward",
     "onCommand:r.helpPanel.openForSelection",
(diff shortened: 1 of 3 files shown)


renkun-ken (approved):
Works nicely.

LGTM

PR #852, opened 2021-11-01
Show httpgd plot on attach

### LINKED ISSUE #832: Session watcher cannot attach to session with httpgd device open
**Describe the bug**

If an httpgd device is open, then session watcher cannot attach to the session if `r.plot.useHttpgd` is enabled.

**To Reproduce**

Ensure httpgd is installed and `r.plot.useHttpgd` is enabled.

1. Open a workspace folder.
2. Create R terminal and session watcher attaches to the R session.
3. Run `plot(rnorm(100))` and close the httpgd plot viewer.
4. Create another R terminal and session watcher attaches to the new R session.
5. Go to the first R terminal and run "Attach active terminal" or click the status bar item.
6. The plot created in the first R terminal shows but the session watcher does not attach to it as indicated by the pid shown in the status bar item.
7. Use `dev.off()` to close the device in the first terminal, and then it could be attached.

It is caused by that attaching the first R terminal makes it write two consecutive requests to `~/.vscode-R/request.log`, the request to reveal the plot viewer overwrites the attach request.

renkun-ken:
# What problem did you solve?

Close #832 

This PR moves the separate `httpgd` request url into the `attach` request. If a `httpgd` device is open, then its url is also sent to session watcher on attach.

This PR also removes the code to send `vsc.plot` back to session watcher as it is already defined as `r.session.viewers.viewColumn.plot` in vscode settings.

1. Open a workspace folder.
2. Create R terminal and session watcher attaches to the R session.
3. Run `plot(rnorm(100))` and close the httpgd plot viewer.
4. Create another R terminal and session watcher attaches to the new R session.
5. Go to the first R terminal and run "Attach active terminal" or click the status bar item.
6. The first session should be attached and the httpgd plot viewer should also appear.

### DIFF
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -480,16 +480,13 @@ attach <- function() {
   request("attach",
     version = sprintf("%s.%s", R.version$major, R.version$minor),
     tempdir = tempdir,
-    plot = getOption("vsc.plot", "Two"),
     info = list(
       command = commandArgs()[[1L]],
       version = R.version.string,
       start_time = format(file.info(tempdir)$ctime)
-    )
+    ),
+    plot_url = if (identical(names(dev.cur()), "httpgd")) httpgd::hgd_url()
   )
-  if (identical(names(dev.cur()), "httpgd")) {
-    .vsc$request("httpgd", url = httpgd::hgd_url())
-  }
 }
 
 path_to_uri <- function(path) {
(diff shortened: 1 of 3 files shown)


ElianHugh:
Thanks for this - it works as expected on my laptop. I'd like to double check the interaction with Live Share, but won't have access until tomorrow

renkun-ken:
In a live share session, it looks like the guest does not handle `httpgd` request in the first place. What's the expected behavior?

ElianHugh (approved):
Thank you for your patience -  LGTM! Guests should have the host's httpgd session shared with them, but technically it is the host that is doing the attaching.

PR #859, opened 2021-11-07
Add backtick to list of quote characters for syntax highlighting.

jan-imbi:
# What problem did you solve?
Backticks are a legal way to quote strings in R. Added backticks to list of quote characters for syntax highlighting.

## Screenshot
Before:
before
After:
after

### DIFF
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -262,11 +262,11 @@
                             "name": "punctuation.definition.string.end.r"
                         }
                     },
-                    "name": "string.quoted.single.r",
+                    "name": "variable.parameter.r",
                     "patterns": [
                         {
                             "match": "\\\\.",
-                            "name": "constant.character.escape.r"
+                            "name": "variable.parameter.r"
                         }
                     ]
                 }
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -248,6 +248,27 @@
                             "name": "constant.character.escape.r"
                         }
                     ]
+                },
+                {
+                    "begin": "`",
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
Works nicely! LGTM.

Thanks!

ManuelHentschel:
If I'm not mistaken, backticks don't delimit strings but rather non-standard variable names. So maybe this should be highlighted as some kind of variable name (e.g. `"general-variables"`) instead?

jan-imbi:
> If I'm not mistaken, backticks don't delimit strings but rather non-standard variable names. So maybe this should be highlighted as some kind of variable name (e.g. `"general-variables"`) instead?

I think you are correct, from the documentation of ?Quotes:

```
## Backticks are used for non-standard variable names.
## (See make.names and ?Reserved for what counts as
## non-standard.)
```

I *think* the following should work:

```
  {
      "begin": "`",
      "beginCaptures": {
          "0": {
              "name": "punctuation.definition.string.begin.r"
          }
      },
      "end": "`",
      "endCaptures": {
          "0": {
              "name": "punctuation.definition.string.end.r"
          }
      },
      "name": "variable.parameter.r",
      "patterns": [
          {
              "match": "\\\\.",
              "name": "variable.parameter.r"
          }
      ]
  }
```

renkun-ken:
Yes, it should be "general-variables". The last commit looks good now.

ManuelHentschel (approved):
Thanks!

PR #863, opened 2021-11-09
Auto refresh help

### LINKED ISSUE #854: Auto-refresh help topics
Currently, it looks like help topics could be properly shown if `getAliases.R` is run with the latest installed packages.

As reported in #661, #853, if a new package is installed but `getAliases.R` is not run, the cache is out of date and thus both "Open help for selection" and "hover" will not work with the newly installed packages. Manually clicking "Clear Cached Index Files & Restart Help Server" could refresh the aliases but it seems possible to make it more automatic.

@ManuelHentschel what if we make some changes to `helpServer.R` to periodically detect changes of installed packages and write the aliases json once there are changes, and `getAliases.R` will be no longer needed.

```r
# get values from extension-set env values
lim <- Sys.getenv("VSCR_LIM")

cat(
    lim,
    tools::startDynamicHelp(),
    lim,
    sep = ""
)

pkgs <- character()
while (TRUE) {
    cur_pkgs <- .packages(all.available = TRUE)
    if (!identical(pkgs, cur_pkgs)) {
        ip <- installed.packages()

        ord <- order(ip[, "Priority"])
        ip <- ip[ord, c("LibPath", "Package")]

        ret <- lapply(rownames(ip), function(row) {
            libPath <- ip[row, "LibPath"]
            pkg <- ip[row, "Package"]
            filename <- file.path(libPath, pkg, "help", "aliases.rds")
            info <- list(
                package = pkg,
                libPath = libPath,
                aliasFile = filename,
                aliases = NULL
            )
            if (file.exists(filename)) {
                info[["aliases"]] <- as.list(readRDS(filename))
            }
            info
        })
        names(ret) <- rownames(ip)
        json <- jsonlite::toJSON(ret, auto_unbox = TRUE)
        cat(lim, json, lim, "\n", sep = "")
        pkgs <- cur_pkgs
    }
    Sys.sleep(1)
}
```

Whenever a new package installed, the latest aliases will be written to stdout, or otherwise it is checking every second, which should not be expensive.

And the help manager should handle both output accordingly. One is help server port message, and the other is aliases json message.

ManuelHentschel:
Fix #854

Ho to check:
 - Have vscode open, install a new package in a separate process, use R help viewer. The Package should show up in the corresponding lists.
 - Open the treeview `R > Help Pages > Help Topics by Package`, install/remove a package in a separate process. The treeview should automatically refresh to reflect the change.

*Edit:*
I also refactored a bit of help-related code to make implementing this a bit easier. The most relevant changes are in [helpServer.R](https://github.com/ManuelHentschel/vscode-R/blob/d0154b35cfee01b8082027dd1683d609bbe5becc/R/help/helpServer.R#L19-L23) and [helpProvider.ts](https://github.com/ManuelHentschel/vscode-R/blob/d0154b35cfee01b8082027dd1683d609bbe5becc/src/helpViewer/helpProvider.ts#L74-L77).

### DIFF
--- a/R/help/helpServer.R
+++ b/R/help/helpServer.R
@@ -1,8 +1,6 @@
 # get values from extension-set env values
 lim <- Sys.getenv("VSCR_LIM")
 
-SLEEP_DURATION <- 1
-
 NEW_PACKAGE_STRING <- "NEW_PACKAGES"
 
 cat(
@@ -12,13 +10,15 @@ cat(
     sep = ""
 )
 
-currentPackages <- .packages(all.available = TRUE)
+currentPackages <- NULL
 
 while (TRUE) {
-    Sys.sleep(1)
-    newPackages <- .packages(all.available = TRUE)
+    newPackages <- installed.packages(fields = "Packaged")[, c("Version", "Packaged")]
     if (!identical(currentPackages, newPackages)) {
-        cat(NEW_PACKAGE_STRING, "\n")
+        if (!is.null(currentPackages)) {
+            cat(NEW_PACKAGE_STRING, "\n")
+        }
         currentPackages <- newPackages
     }
+    Sys.sleep(1)
 }
--- a/R/help/helpServer.R
+++ b/R/help/helpServer.R
@@ -1,6 +1,10 @@
 # get values from extension-set env values
 lim <- Sys.getenv("VSCR_LIM")
 
+SLEEP_DURATION <- 1
+
+NEW_PACKAGE_STRING <- "NEW_PACKAGES"
+
 cat(
     lim,
     tools::startDynamicHelp(),
@@ -8,4 +12,13 @@ cat(
     sep = ""
 )
 
-while (TRUE) Sys.sleep(1)
+currentPackages <- .packages(all.available = TRUE)
+
+while (TRUE) {
+    Sys.sleep(1)
+    newPackages <- .packages(all.available = TRUE)
+    if (!identical(currentPackages, newPackages)) {
+        cat(NEW_PACKAGE_STRING, "\n")
+        currentPackages <- newPackages
+    }
+}
--- a/src/helpViewer/index.ts
+++ b/src/helpViewer/index.ts
@@ -175,7 +175,7 @@ export class RHelp implements api.HelpPanel, vscode.WebviewPanelSerializer<strin
 		this.webviewScriptFile = vscode.Uri.file(options.webviewScriptPath);
 		this.webviewStyleFile = vscode.Uri.file(options.webviewStylePath);
 		const pkgListener = () => {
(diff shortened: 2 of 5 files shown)


src/helpViewer/index.ts
```diff
@@ -174,7 +174,11 @@ export class RHelp implements api.HelpPanel, vscode.WebviewPanelSerializer<strin
 	constructor(options: HelpOptions){
 		this.webviewScriptFile = vscode.Uri.file(options.webviewScriptPath);
 		this.webviewStyleFile = vscode.Uri.file(options.webviewStylePath);
-		this.helpProvider = new HelpProvider(options);
+		const pkgListener = () => {
+			void vscode.window.showInformationMessage('Restarting Help Server...');
```
renkun-ken, line 178: When a number of packages are installed, this message appears, but it lasts too long and becomes annoying. I suggest that we only write logging instead of showing a message here.

R/help/helpServer.R
```diff
@@ -1,11 +1,24 @@
 # get values from extension-set env values
 lim <- Sys.getenv("VSCR_LIM")
 
+SLEEP_DURATION <- 1
```
renkun-ken, line 4: `SLEEP_DURATION` is not really used below.

R/help/helpServer.R
```diff
@@ -1,11 +1,24 @@
...
     sep = ""
 )
 
-while (TRUE) Sys.sleep(1)
+currentPackages <- .packages(all.available = TRUE)
+
+while (TRUE) {
+    Sys.sleep(1)
+    newPackages <- .packages(all.available = TRUE)
+    if (!identical(currentPackages, newPackages)) {
```
renkun-ken, line 20: It works well with changes in installed packages. Sometimes, the names of installed package do not change, but a new version is installed. I guess it might make some sense to use `newPackages <- installed.packages()` here so that if any bit of the information of all packages change, the aliases should be reloaded.

However, it could not handle installing the same version number from GitHub instead of CRAN.

renkun-ken: Looks like we could use `newPackages <- installed.packages(fields = "Packaged")` and compare `currentPackages` with `newPackages[, c("Version", "Packaged")]` should be enough to capture all these changes?

src/helpViewer/index.ts
renkun-ken: Also, when we update all packages, the library is frequently updated and the help server is restarted many times. The message could be annoying in this case too.

renkun-ken (approved):
LGTM

R/help/helpServer.R
ManuelHentschel: Seems to be working nicely, thanks!

PR #867, opened 2021-11-16
Fix rmd preview chunk colouring

### LINKED ISSUE #866: Rmd preview fails if chunk-background setting is empty
With the setting `"r.rmarkdown.chunkBackgroundColor": ""`, the Rmd live preview fails in dark mode.

This can be solved by adding `const regOut = chunkCol.match(colReg) || [];` in [rmarkdown/preview.ts](https://github.com/ManuelHentschel/vscode-R/blob/b47e93b751896b743e269fd9a95189ae74af9f5e/src/rmarkdown/preview.ts#L89-L92), but then the code blocks are rendered with a white background and somewhat difficult to read.

ElianHugh:
# What problem did you solve?

Resolves #866. If a chunk background colour is not defined, fall back to a vscode theme colour

``````
```
A non-code chunk
```

```{r}
1:100
```
``````

### DIFF
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -91,10 +91,10 @@ class RMarkdownPreview extends vscode.Disposable {
         if (chunkCol) {
             const colReg = /[0-9.]+/g;
             const regOut = chunkCol.match(colReg);
-            outCol = `rgba(${regOut[0] ?? 100}, ${regOut[1] ?? 100}, ${regOut[2] ?? 100}, ${Number(regOut[3]) + 0.05 ?? .5})`;
+            outCol = `rgba(${regOut[0] ?? 128}, ${regOut[1] ?? 128}, ${regOut[2] ?? 128}, ${Math.max(0, Number(regOut[3]) - 0.05) ?? 0.05})`;
         } else {
-            chunkCol = 'var(--vscode-selection-background)';
-            outCol = 'var(--vscode-selection-background)';
+            chunkCol = 'rgba(128, 128, 128, 0.1)';
+            outCol = 'rgba(128, 128, 128, 0.05)';
         }
 
         const style =
--- a/src/rmarkdown/preview.ts
+++ b/src/rmarkdown/preview.ts
@@ -86,10 +86,16 @@ class RMarkdownPreview extends vscode.Disposable {
         this.htmlLightContent = $.html();
 
         // make the output chunks a little lighter to stand out
-        const chunkCol = String(config().get('rmarkdown.chunkBackgroundColor'));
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/rmarkdown/preview.ts [hunk 1]
```diff
@@ -86,10 +86,16 @@ class RMarkdownPreview extends vscode.Disposable {
...
-        const outCol = `rgba(${regOut[0] ?? 100}, ${regOut[1] ?? 100}, ${regOut[2] ?? 100}, ${Number(regOut[3]) + 0.05 ?? .5})`;
+        let chunkCol = String(config().get('rmarkdown.chunkBackgroundColor'));
+        let outCol: string;
+        if (chunkCol) {
+            const colReg = /[0-9.]+/g;
+            const regOut = chunkCol.match(colReg);
+            outCol = `rgba(${regOut[0] ?? 100}, ${regOut[1] ?? 100}, ${regOut[2] ?? 100}, ${Number(regOut[3]) + 0.05 ?? .5})`;
+        } else {
+            chunkCol = 'var(--vscode-selection-background)';
+            outCol = 'var(--vscode-selection-background)';
```
renkun-ken, line 97: Should we make the chunk and output appear slightly different?

image

image

renkun-ken:
Looks like we should use something like `rgba(128, 128, 128, 0.1)` and `rgba(128, 128, 128, 0.05)` to get a more consistent coloring under different themes.

src/rmarkdown/preview.ts
```diff
@@ -86,10 +86,16 @@ class RMarkdownPreview extends vscode.Disposable {
         this.htmlLightContent = $.html();
 
         // make the output chunks a little lighter to stand out
-        const chunkCol = String(config().get('rmarkdown.chunkBackgroundColor'));
-        const colReg = /[0-9.]+/g;
-        const regOut = chunkCol.match(colReg);
-        const outCol = `rgba(${regOut[0] ?? 100}, ${regOut[1] ?? 100}, ${regOut[2] ?? 100}, ${Number(regOut[3]) + 0.05 ?? .5})`;
```
renkun-ken, line 92: I suspect that the alpha channel should be something like `${Number(regOut[3]) - 0.05 ?? .05}` to make the output more like the background. In light themes, the default rendering looks like the code chunk is more highlighted than the output from the background.

src/rmarkdown/preview.ts
renkun-ken: Looks like we should use something like `rgba(128, 128, 128, 0.1)` and `rgba(128, 128, 128, 0.05)` to get a more consistent coloring under different themes.

renkun-ken (approved):
LGTM

Issue #796, opened 2021-09-23
Custom `xaringan::inf_mr` does not trigger Web View.


jooyoungseo:
In the latest version 2.3.0, a new feature has been introduced for custom knit management. This is very great! Thanks for @ElianHugh's commit.

I have just found that `xaringan::inf_mr` custom knit does not open a new Web View while either manual or rstudio.api is working as expected.

Could you please address this issue?

# Reprex

````
---
output: html_document
knit: xaringan::inf_mr
---

# Test

Para.
````

ElianHugh:
Thanks for the issue! This is because, at the moment,  the background process doesn't communicate with the main R process and therefore doesn't trigger session-watcher capabilities (in this case, webviews). 

I agree that this is something that should just 'work '

renkun-ken:
I wonder what is the desired way to render this document?

ElianHugh:
If this is a normal html file, you could try enabling the setting `r.rmarkdown.knit.openOutputFile`, which will attempt to open the file when knitting finishes. *Generally* this opens in a browser, but depending on your computer, may open in vscode. 

An alternative would be to open the file, and use the Live Preview extension to handle previewing html files

jooyoungseo:
I tested the following two cases; `r.rmarkdown.knit.openOutputFile` doesn't work as expected.

1. `knit: rmarkdown::render` with `r.rmarkdown.knit.openOutputFile = TRUE` setting: Opens HTML file as a raw source code (not trigger WebView).

2. `knit: xaringan::inf_mr` with `r.rmarkdown.knit.openOutputFile = TRUE` setting: Doesn't even open the raw source code file.

jooyoungseo:
PDF file is open as expected though:

3. `knit: pagedown::chrome_print` with `r.rmarkdown.knit.openOutputFile = TRUE` setting: Opens PDF file.

ElianHugh:
I've been looking at this again, and a more simple solution would be to hardcode the edge cases to not use the background process

ElianHugh:
Scratch that, I've got this *semi*-functioning. Need to look into how this affects how we manage the child processes

![image](https://user-images.githubusercontent.com/60372411/142558316-a6e0ab18-6639-4cb0-b33b-1ee56244ec8a.png)

ElianHugh:
unstale

PR #875, opened 2021-11-22
WIP - custom knit: handle xaringan::inf_mr 

ElianHugh:
# What problem did you solve?
Solves #796. 

Still WIP as need to ensure that the URL regex is appropriate, and needs some tidying of the code.

Knitting the following should work (i.e., progress bar signals completion), and closing the webview should dispose of the process.

``````
---
knit: xaringan::inf_mr
---

# Test

Para.

---

# Next
``````


ElianHugh:
i'm a little iffy on how broad the url regex should be, because i don't want it to trigger on just any url that might pop up in the output stream

PR #880, opened 2021-11-24
adding devtools tasks to command palette

alex-gable:
# What problem did you solve?
while the majority of `devtools::*` package develop/build shortcuts were available, `devtools::check()` was missing.

## (If you have)Screenshot

R: Check available in Command Palette and sends to R console

Build version > `Cmd + Shift + P` > Type `R: Check` > Press Enter

### DIFF
--- a/package.json
+++ b/package.json
@@ -433,22 +433,22 @@
       {
         "title": "Check",
         "category": "R Package",
-        "command": "r.checkTask"
+        "command": "r.check"
       },
       {
         "title": "Document",
         "category": "R Package",
-        "command": "r.documentTask"
+        "command": "r.document"
       },
       {
         "title": "Install",
         "category": "R Package",
-        "command": "r.installTask"
+        "command": "r.install"
       },
       {
         "title": "Test",
         "category": "R Package",
-        "command": "r.testTask"
+        "command": "r.test"
       },
       {
         "title": "Attach Active Terminal",
--- a/package.json
+++ b/package.json
@@ -422,33 +422,33 @@
       },
       {
         "title": "Load All",
-        "category": "R",
+        "category": "R Package",
         "command": "r.loadAll"
       },
       {
-        "title": "Test Package",
-        "category": "R",
-        "command": "r.test"
-      },
-      {
-        "title": "Install Package",
-        "category": "R",
-        "command": "r.install"
-      },
-      {
-        "title": "Build Package",
-        "category": "R",
+        "title": "Build",
+        "category": "R Package",
         "command": "r.build"
       },
       {
-        "title": "Document",
-        "category": "R",
-        "command": "r.document"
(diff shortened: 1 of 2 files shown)


ElianHugh:
I wonder if it would be worth moving away from the runTextInTerm paradigm for these commands (I.e. commands that don't influence/affect the R environment). Is there a benefit to running in the terminal as opposed to a task? We have the devtools::check task definition but iirc no command palette bbinding

renkun-ken:
I think it should be a task instead of sending code to terminal.

alex-gable:
I believe that's how RStudio treats it as well, for whatever that's worth. Should that be in the scope of this PR though?

ElianHugh:
I think for this PR, it would probably be worth pointing the command palette to the task binding. We can do the same for the other command palette bindings in a future PR

alex-gable:
Understood. Just to be transparent, that's likely outside of my skill set

ElianHugh:
No worries, you're 90% there already! I think you would only need to change the command and its arguments. E.g., in my keybindings I have the following:

```
{		
  "key": "ctrl+shift+t",		
  "command": "workbench.action.tasks.runTask",		
  "args": "R: Test"
},
```

alex-gable:
okay, I _think_ this is the solution you alluded to, @ElianHugh, I appreciate the pointer. I did a bit of research and there's definitely more complex (and I think appropriate) ways to accomplish this. see: microsoft/vscode#40758.

any how. I took a couple of liberties in terms of naming and organization. I wanted to be sure the commands that match `Check`, which execute outside of the Terminal, all had tasks. so I added a task for `Build` and then grouped these task commands separately from the other editor independent commands. 

I then categorized these "Package" command contributions separately in the because, as I was testing, it made the command in the palette clearer.

Feel free to revert/change whatever!

package.json
```diff
@@ -422,28 +422,33 @@
...
+        "category": "R Package",
         "command": "r.loadAll"
       },
       {
-        "title": "Test Package",
-        "category": "R",
-        "command": "r.test"
+        "title": "Build",
+        "category": "R Package",
+        "command": "r.build"
```
ElianHugh, line 431: This should probably point to r.buildTask (currently errors out)

ElianHugh:
Thank you for the nice work you've done! The commands work nicely (barring the one issue I've noted above), and I think the naming is appropriate too. I don't have a strong opinion re: naming the commands r.commandTask, but it may cause unexpected  issues for VSCode-R users that have shortcuts pointing to the old command names (e.g. r.test vs r.testTask)

ElianHugh (approved):
LGTM

ElianHugh:
Anything else to add? @renkun-ken

renkun-ken:
LGTM sorry for late reply!

Issue #862, opened 2021-11-09
Ikuyadeu.r formatter is 


Yemoloh:
Hi, I've installed R on VScode, but when I try to format my code, I have this message :
"There is no formatter for 'r' files installed"
This is my settings.json

{
"tabnine.experimentalAutoImports": true,
"workbench.colorTheme": "Cobalt2",
"r.alwaysUseActiveTerminal": true,
"r.rpath.windows": "C:\Program Files\R\R-4.1.2\bin",
"r.rterm.windows": "C:\Program Files\R\R-4.1.2\bin\R.exe",
"files.associations": {
"*.rmd": "markdown"
},
"r.dependencyChecks": true,
"editor.formatOnPaste": true,
"editor.formatOnType": true,
"r.lsp.debug": true,
"[r]": {
"editor.defaultFormatter": "Ikuyadeu.r"
},
}

Does anybody had this issue before ?

Yemoloh

To Reproduce

1. Install languageserver
2. Install r.extension
3. Add : "[r]": {
"editor.defaultFormatter": "Ikuyadeu.r"
}, in settings.json
4. Open an r.file, type unformatted code and hit shift+alt+f

- OS: [Windows 10]
- VSCode Version: [1.62.0]
- R Version: [4.1.2]
- vscode-R version: [2.3.2]

renkun-ken:
Looks like the R language client does not correctly start and thus not registered as a valid R formatter. Could you go to `Output -> R Language Server` and take a look at the log of language server. Does it start correctly?

robokamran:
I have the same problem with pretty much the same configuration.
how should I go to `Output -> R Language Server` ??

renkun-ken:
Go to the Output pane and choose "R Language Server" and view the log:

<img width="1476" alt="image" src="https://user-images.githubusercontent.com/4662568/143675112-42e3730f-0f32-44f9-8a39-c97b05f6f9aa.png">

robokamran:
Thank you for your quit reply.
Here is my screenshot, the code runs correctly, but when I hit `ALT+Shift+F` I get that error.
And the output looks like this:

![Screenshot 2021-11-27 191819](https://user-images.githubusercontent.com/8957516/143677254-f73c01cf-49bf-4d2e-a2d0-1a7833f843fe.jpg)

renkun-ken:
Looks like the language server does not start at all.

@ManuelHentschel Does the value of `r.rpath.windows` looks correct?

[...]

PR #884, opened 2021-11-27
Remove quotes from rpath if necessary

ManuelHentschel:
Re #862

Removes the quotes around `r.rpath.XXX` if they are used in the settings but not required for the specific usecase.


renkun-ken (approved):
LGTM

PR #887, opened 2021-11-28
Clean up of help related files

ManuelHentschel:
Includes some general clean up of help related files.
 - Convert methods to functions if they can be standalone
 - Correctly handle `null`/`undefiend` (set `"strictNullChecks": true` in tsconfig.json to check)

### DIFF
--- a/src/helpViewer/helpProvider.ts
+++ b/src/helpViewer/helpProvider.ts
@@ -34,9 +34,10 @@ export class HelpProvider {
         this.port = this.launchRHelpServer(); // is a promise for now!
     }
 
-    public refresh(): void {
+    public async refresh(): Promise<void> {
         kill(this.cp.pid); // more reliable than cp.kill (?)
         this.port = this.launchRHelpServer();
+        await this.port;
     }
 
     public async launchRHelpServer(): Promise<number>{
--- a/src/helpViewer/helpProvider.ts
+++ b/src/helpViewer/helpProvider.ts
@@ -35,7 +35,9 @@ export class HelpProvider {
     }
 
     public async refresh(): Promise<void> {
-        kill(this.cp.pid); // more reliable than cp.kill (?)
+        if(!this.cp.exitCode){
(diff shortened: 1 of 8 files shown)


src/helpViewer/helpProvider.ts
```diff
@@ -34,9 +34,10 @@ export class HelpProvider {
         this.port = this.launchRHelpServer(); // is a promise for now!
     }
 
-    public refresh(): void {
+    public async refresh(): Promise<void> {
         kill(this.cp.pid); // more reliable than cp.kill (?)
```
renkun-ken, line 38: I'm wondering if it could kill the wrong process?

Say we start R help server (pid=12345) and `this.cp.pid` remembers it. And we manually kill the process. Then another user process starts with pid=12345. Now if I click refresh cache & restart help server, it will kill the wrong process as the pid is used by another process.

When the process is killed externally, the only difference here seems to the value of `cp.signalCode` is `'SIGTERM'`.

ManuelHentschel: I added a check to see if the cp process has exited (for whatever reason), before killing it.

src/helpViewer/helpProvider.ts
```diff
@@ -34,9 +34,12 @@ export class HelpProvider {
         this.port = this.launchRHelpServer(); // is a promise for now!
     }
 
-    public refresh(): void {
-        kill(this.cp.pid); // more reliable than cp.kill (?)
+    public async refresh(): Promise<void> {
+        if(!this.cp.exitCode){
```
renkun-ken, line 38: This check might not work as expected. I kill the help server process and here is what `cp` looks like when I try to restart the help server while debugging:

image

ManuelHentschel: Does the latest push work reliably for you?

renkun-ken: Thanks. The latest push is much more reliable.

renkun-ken (approved):
LGTM

PR #893, opened 2021-11-30
Fix helpserver issue

### LINKED ISSUE #892: Help server restart unreliable when installing packages
When I install a package, the numbers of help servers will increase after the installation. It looks like the way to restart the process is not reliable or at least not working consistently as expected.

image

ManuelHentschel:
Fix #892


renkun-ken (approved):
Works nicely! LGTM

PR #914, opened 2021-12-14
RMD - don't set undefined wd

### LINKED ISSUE #895: R markdown: Knit button and keyboard knit shortcut doesn't work
**Describe the bug**
I use the same rmd file as in #886 in this report. (The preview problem has been fixed.) When the setting "Knit: Use Background Process" is on, the knit button and keyboard shortcut (Ctrl + Shift + k) don't work. The output panel of R markdown showed nothing. But after turning off the "Knit: Use Background Process", the rmd file can be knitted correctly.

**To Reproduce**

1. turn on "Knit: Use Background Process"
2. Click on 'Knit'

## Do you want to fix by self? (We hope your help!)

No, I don't know how vscode works.

```jsonc
// R.exe path for windows
"r.rterm.windows": "D:\Softwares\Anaconda3\Scripts\radian.exe",

// R path for Mac OS X
"r.rterm.mac": "",

// R path for Linux
"r.rterm.linux": "",

// R command line options (i.e: --vanilla)
"r.rterm.option": [--no-save --no-restore],

// An optional encoding to pass to R when executing the file, i.e. 'source(FILE, encoding=ENCODING)'
"r.source.encoding": "UTF-8",

// Keeping focus when running
"r.source.focus": "editor",

// Use active terminal for all commands, rather than creating a new R terminal
"r.alwaysUseActiveTerminal": false,

// Use bracketed paste mode
"r.bracketedPaste": false,

// Enable R session watcher
"r.sessionWatcher": true,

// Delay in milliseconds before sending each line to rterm (only applies if r.bracketedPaste is false)
"r.rtermSendDelay": 8,
```

**Expected behavior**
The rmd file can be knitted correctly when "Knit: Use Background Process" is on.

**Screenshots**
"Knit: Use Background Process" is on:
image

"Knit: Use Background Process" is off:
image

- OS: Windows
- VSCode Version: 1.62.3 (user setup)
- R Version: 4.1.1
- vscode-R version: v2.3.4

ElianHugh:
Closes #895

# What problem did you solve?

- Prevents setting the workspace rmd if the value is undefined
- Fall back to document directory when workspace folder does not exist

*Without opening a workspace folder* attempt to knit & preview an RMD file. Previously this would result in an error.

### DIFF
--- a/R/rmarkdown/knit.R
+++ b/R/rmarkdown/knit.R
@@ -11,7 +11,7 @@ knit_command <- Sys.getenv("VSCR_KNIT_COMMAND")
 
 # set the knitr chunk eval directory
 # mainly affects source calls
-if (!is.null(eval(parse(text = knit_dir)))) {
+if (nzchar(knit_dir)) {
     knitr::opts_knit[["set"]](root.dir = knit_dir)
 }
 
--- a/R/rmarkdown/knit.R
+++ b/R/rmarkdown/knit.R
@@ -11,7 +11,9 @@ knit_command <- Sys.getenv("VSCR_KNIT_COMMAND")
 
 # set the knitr chunk eval directory
 # mainly affects source calls
-knitr::opts_knit[["set"]](root.dir = knit_dir)
+if (!is.null(eval(parse(text = knit_dir)))) {
+    knitr::opts_knit[["set"]](root.dir = knit_dir)
+}
 
 # render and get file output location for use in extension
 cat(
--- a/R/rmarkdown/preview.R
+++ b/R/rmarkdown/preview.R
@@ -29,7 +29,7 @@ set_html <- tryCatch(
 
 # set the knitr chunk eval directory
 # mainly affects source calls
-if (!is.null(eval(parse(text = knit_dir)))) {
+if (nzchar(knit_dir)) {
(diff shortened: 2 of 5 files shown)


R/rmarkdown/knit.R
```diff
@@ -11,7 +11,9 @@ knit_command <- Sys.getenv("VSCR_KNIT_COMMAND")
 
 # set the knitr chunk eval directory
 # mainly affects source calls
-knitr::opts_knit[["set"]](root.dir = knit_dir)
+if (!is.null(eval(parse(text = knit_dir)))) {
```
renkun-ken, line 14: When `knit_dir` is a path like `/home/user/documents`, then `eval(parse(text = knit_dir))` will result in an error

```
Error in parse(text = knit_dir) : <text>:1:1: unexpected '/'
1: /
    ^
Calls: eval -> parse

Backtrace:

1: parse(text = knit_dir)
2: eval(parse(text = knit_dir))

Exiting on error
```

renkun-ken: I think `knitWorkingDirText` should fallback to empty string `""` and checking `nzchar(knit_dir)` should be enough.

PR #915, opened 2021-12-14
readability adjustments for help pages

18kimn:
# What problem did you solve?

For those who use VSCode in a light palette (here, the default Light and the Better Solarized Light themes), some of the help pages were hard to read:

<div>
    
    
</div>

I just updated CSS styling to be different depending on if the user had a light or dark theme. Now: 

<div>
    
    
</div>

You can get to the page shown with `?.GlobalEnv`


renkun-ken:
Is it possible to use vscode colors in all syntax highlighting entries?

18kimn:
Do you mean using the colors from the user's theme? The answer is sort of, but mostly no, I don't think it's possible to programmatically access colors. See the extended discussion here: https://github.com/Microsoft/vscode/issues/32813

I'm not really sure why but the best option seems to be either 1) adapt to whether the user has a dark or light theme (which I did) or 2) try to use the CSS variables specified at the top level of the webview (they begin with `--vscode-editor` I believe). This second option doesn't let you use tokens for highlighting, which is probably why `vscode-R` is using highlight-js for the help pages

Open to suggestions!

ManuelHentschel:
Thanks for the PR! In https://code.visualstudio.com/api/extension-guides/webview#theming-webview-content it is described how to theme webviews using (only) CSS. Have you tried using that approach? This should e.g. have the advantage that you don't need to reload vscode to use a different theme.

18kimn:
Here's an adjustment for that! 

It's still pretty much limited to just using either light or dark, since VSCode doesn't provide CSS variables to style tokens, but it at least seems to immediately update when I change my theme

ManuelHentschel:
Works nicely, thanks!

Should we just use the dark-theme colors for high-contrast themes?

And before merging, please don't forget to fix the indentation (4 spaces in typescript and css files).

ManuelHentschel:
> It's still pretty much limited to just using either light or dark, since VSCode doesn't provide CSS variables to style tokens

I think that's good enough for now. Let's just hope, the vscode guys give us access to the token colors at some point...

renkun-ken (approved):
LGTM

18kimn:
Thanks a bunch @renkun-ken @ManuelHentschel, I added the high-contrast theme and made indentation be 4 spaces (I Prettier'ed the `helpViewer/index.ts` file since there seemed to be quite a few indentation mismatches there)

I'm a bit unsure about the `vscode-high-contrast` theming; I tried out a few different puportedly "high-contrast" themes (Sonokai, VLight/VDark, Nebula Oni, see more [here](https://open-vsx.org/?search=high%20contrast)) to test this and the actual class name stayed as `vscode-dark` or `vscode-light`. [Github's](https://github.blog/changelog/2021-09-21-github-dark-high-contrast-theme-for-vs-code/) high-contrast theme was the only one I saw recognized as high-contrast.  Also, high-contrast themes should be able to come in both light and dark flavors, which it would be great to adjust for in the CSS, but this is impossible to tell via classname currently and both Github and VSCode's high contrast themes only come in dark. 

In any case, this might need more adjustment down the road, but maybe it's good enough for now? I'm not sure if I have merge permissions, would one of you be able to merge?

renkun-ken:
I think it is good enough for now. Thanks for working on this!

PR #931, opened 2022-01-04
Fixed `"punctuation.section.parens.end.r"` under `"function-parameters"`

RedCMD:
# What problem did you solve?
Fixed end bracket not being tokenized `"punctuation.section.parens.end.r"` under `"function-parameters"` because `"name"` was not used

Happened to see it while looking through random vscode extensions
(in red at the bottom)
image

### DIFF
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -560,14 +560,6 @@
                     "name": "punctuation.separator.parameters.r"
                 },
                 {
-                    "end": "\\)",
-                    "endCaptures": {
-                        "0": {
-                            "name": "punctuation.section.parens.end.r"
-                        }
-                    }
-                },
-                {
                     "include": "source.r"
                 }
             ]
--- a/syntax/r.json
+++ b/syntax/r.json
@@ -562,7 +562,9 @@
                 {
                     "end": "\\)",
                     "endCaptures": {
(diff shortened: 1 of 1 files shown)


renkun-ken:
Thanks for the fix!

Would you like to provide an example code that demonstrates the problem?

RedCMD:
On closer inspection, it seems that, that part of the highlighter code does nothing
and should just be removed entirely
'Remove absolute `"end"` pattern under `"function-parameters"`'

renkun-ken:
Would you like to show the incorrect syntax highlighting before this fix?

RedCMD:
A R code example?
there would be none because the code never tries to match anything
its just dead code
most likely left over from some testing
because the `"end"` rule has no `"begin"` rule, it is never evaluated during runtime and thus does nothing
(Sorry I didn't notice this at first)
The code in question was added here on line 333/467 https://github.com/REditorSupport/vscode-R/commit/743737a1fdbf482ff28cbe10e00fc13ce4d27823#diff-f6f05bbc187255f5cf06a39e3a4f7befa7288489c5fd44bb14586143076ffd1d

renkun-ken (approved):
LGTM

PR #949, opened 2022-01-21
update the .gitignore file for R

eitsupi:
# What problem did you solve?

Reflects the update of the original .gitignore file for R.
It also reflects the comments that existed in the original file.

https://github.com/github/gitignore

Check the .gitignore file that is generated by executing the "R: Create gitignore" command from the command palette.


renkun-ken:
Would it be easier to maintain if we just put the file at `R/R.gitignore` and just copy the file?

eitsupi:
> Would it be easier to maintain if we just put the file at `R/R.gitignore` and just copy the file?

That would be great.
In that case, I think GitHub Actions can detect the upstream updates and update that file semi-automatically by using https://github.com/peter-evans/create-pull-request.

eitsupi:
I noticed that this command now silently overwrites the .gitignore file, so I changed it to ask the user if they want to overwrite it on the last commit.

src/rGitignore.ts
```diff
@@ -1,32 +1,77 @@
...
+    ''].join('\n');
 
-export function createGitignore(): void {
+export async function createGitignore(): Promise<void> {
     if (workspace.workspaceFolders[0].uri.path === undefined) {
         void window.showWarningMessage('Please open workspace to create .gitignore');
 
         return;
     }
     const ignorePath = join(workspace.workspaceFolders[0].uri.path, '.gitignore');
```
renkun-ken, line 65: Just notice that the usage of `workspace.workspaceFolders[0].uri.path` here does not work with no workspace case, nor does it work with multi-root workspace. Do you want to fix it here?

eitsupi: Sorry, I didn't know the proper way to fix this.

eitsupi:
I've isolated the R.gitignore file, but I'm not sure if this is the right way to do it. It does not seem to work on the extension development host.

src/rGitignore.ts
renkun-ken: Never mind, I'll fix this later in a separate PR.

src/rGitignore.ts
```diff
@@ -1,33 +1,31 @@
...
-    writeFile(ignorePath, ignoreFiles, (err) => {
+    if (existsSync(ignorePath)) {
+        const override = await window.showWarningMessage(
+            '".gitignore" file is already exist. Do you want to override?',
```
renkun-ken, line 21: May it should be "overwrite" instead of "override"?

renkun-ken (approved):
LGTM

PR #954, opened 2022-01-27
correcting typo on command argument (slient instead of silent)

achey2016:
# What problem did you solve?

A typo where silent had been misspelled slient

it is just a typo.


renkun-ken (approved):
LGTM thanks!

PR #958, opened 2022-01-30
Fix rmd comment

ManuelHentschel:
Modifies the rmd commenting. Before this would add a heading (`#`), now a HTML-style comment is added. Comments inside code blocks are unaffected.


renkun-ken (approved):
LGTM. Works nicely.

PR #961, opened 2022-02-03
set the LANG env when rendering rmarkdown

### LINKED ISSUE #933: Rmarkdown inside the VSCode-R cannot render Chinese character in figure/table caption correctly
**Describe the bug**

The rmarkdown within the VSCode cannot render Chinese character correctly.
(after clicking the `knit rmd` button, the corresponding pdf renders UTF8 codes instead of correct Chinese character).

However, the same document can be rendered correctly if I open an terminal and type

```bash
R
rmarkdown::render("file.rmd")
```
So I think there's nothing to do with the encoding of my source file. The `render()` function takes "UTF8" encoding as default option.

**To Reproduce**

Suppose we have the following rmd file `file.rmd`:

~~~text
---
title: "minimal example"
author: "Kobe Bryant"
output: html_document    
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = F, warning = F, message = F, eval = T)
```


Chinese character in the main article can be rendered normally.
(正文区的中文可以正确的渲染)

Figure legends cannot be rendered normally.

```{r, fig.cap = "科比"}

plot(1:5, 1:5)

```
~~~

```jsonc


// R path for Mac OS X
"r.rterm.mac": "/usr/local/bin/R",

// R command line options (i.e: --vanilla)
"r.rterm.option": ["--no-save", "--no-restore"],

// An optional encoding to pass to R when executing the file, i.e. 'source(FILE, encoding=ENCODING)'
"r.source.encoding": "UTF-8",

// Keeping focus when running
"r.source.focus": "editor",

// Use active terminal for all commands, rather than creating a new R terminal
"r.alwaysUseActiveTerminal": false,

// Use bracketed paste mode
"r.bracketedPaste": true,

// Enable R session watcher
"r.sessionWatcher": true,

// Delay in milliseconds before sending each line to rterm (only applies if r.bracketedPaste is false)
"r.rtermSendDelay": 8,
```

**Expected behavior**

The Chinese character within a figure caption can be rendered correctly.

**Screenshots**
The **correct** html document (via calling rmarkdown in terminal) should be:

Screen Shot 2022-01-05 at 11 45 51 PM

What rendered by VSCode-R extension  `knit rmd` button (**wrong**) is:

Screen Shot 2022-01-05 at 11 47 05 PM

- OS: [macOS 12.0.1, Apple M1 MacBook Air]
- VSCode Version: [1.63.2(universal)]
- R Version: [4.1.2(aarch64)]
- vscode-R version: [2.3.5]

shrektan:
# What problem did you solve?

Closes #933 

The `process.env` doesn't contain the LANG environment variable. Thus, when the child process calls R, R will set the locale to "C". This causes an issue that UTF-8 strings can't be corrected handled.

## (If you have)Screenshot

image

Open the below Rmd file in VSCode. Click the knit button to render the file. View the generated HTML file you will see:

(I only verified this on macOS)

1. Before: The locale is "C" and there's no string or garbaged string in the left-bottom corner.
2. After: The locale is "xxx_UTF-8" and the left-bottom displays "科比", as expected.

``````rmd
---
title: "minimal example"
author: "Kobe Bryant"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = F, warning = F, message = F, eval = T)
```

Chinese character in the main article can be rendered normally.
(正文区的中文可以正确的渲染)

Figure legends cannot be rendered normally.

```{r, fig.cap = "科比"}
print(Sys.getlocale())
plot(1:5, 1:5)
```
``````

### DIFF
--- a/src/rmarkdown/manager.ts
+++ b/src/rmarkdown/manager.ts
@@ -74,9 +74,16 @@ export abstract class RMarkdownManager {
 					'-f',
 					scriptPath
 				];
+				// When there's no LANG variable, we should try to set to a UTF-8 compatible one, as R relies
+				// on locale setting (based on LANG) to render certain characters.
+				// See https://github.com/REditorSupport/vscode-R/issues/933
+				let env = process.env;
+				if (env.LANG === undefined) {
+					env.LANG = 'en_US.UTF-8';
+				}
 				const processOptions: cp.SpawnOptions = {
 					env: {
-						...process.env,
+						...env,
 						...scriptArgs
 					},
 					cwd: args.workingDirectory,
--- a/src/rmarkdown/manager.ts
+++ b/src/rmarkdown/manager.ts
@@ -77,7 +77,7 @@ export abstract class RMarkdownManager {
 				// When there's no LANG variable, we should try to set to a UTF-8 compatible one, as R relies
 				// on locale setting (based on LANG) to render certain characters.
 				// See https://github.com/REditorSupport/vscode-R/issues/933
-				let env = process.env;
+				const env = process.env;
 				if (env.LANG === undefined) {
 					env.LANG = 'en_US.UTF-8';
 				}


renkun-ken (approved):
Works as expected. LGTM. Thanks!

PR #964, opened 2022-02-03
should use Strong quote for shell commands

### LINKED ISSUE #932: Document not work now
image

shrektan:
# What problem did you solve?

Fixes #932 

The cause of this issue is that different shell uses different quote style. The reliable way of doing that in VSCode is to use the [`vscode.ShellQuotedString`](https://code.visualstudio.com/api/references/vscode-api#ShellQuotingOptions).

1. Open any R package project.
2. Cmd + Shift + P and choose any "R Package - xxx" command, e.g., "R Package - Document".
3. You'll find the command executes as expected.
4. Probably, you can switch to another Shell and test it again. For example, on Windows you can choose CMD or PowerShell, as the two use different quote for strings `"` for CMD and `'` for PowerShell. Now the command generated will quote the string uses correct style.

## (If you have)Screenshot

(Note that the commands in the two shells use different quote identifiers)

### Windows, Powershell
截屏2022-02-03 21 48 19

### Windows, CMD

image

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -206,17 +206,19 @@ export async function activate(context: vscode.ExtensionContext): Promise<apiImp
     const type = 'R';
     vscode.tasks.registerTaskProvider(type, {
         provideTasks() {
+            // `vscode.ShellQuoting.Strong` will treat the "value" as pure string and quote them based on the shell used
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #965, opened 2022-02-03
prompt to install languageserver is not available

### LINKED ISSUE #754: Prompt and command to install `languageserver`
vscode-R now requires `languageserver` to be installed.

If the language server fails to start due to missing `languageserver` package, it would make sense to prompt to user and ask whether to install the package for the user.

shrektan:
# What problem did you solve?

Fixes #754 

Newbies may find it confusing when the R package `languageserver` is not installed. This PR enables a prompt to instruct the user to install `languageserver` if missing.

(We can use the added function to check other required R packages as well)
(I've also verified that it works on Windows)

1. (R) remove the package via `remove.packages("languageserver")`
2. Open VSCode with any R project
3. A prompt will show up. Click "Not sure" the message vanishes. Click "Click to install" will launch a task to install the package with a notice that the user may need to reload to take effect.

## Screenshot

image

image

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -457,15 +457,21 @@ export function exec(command: string, args?: ReadonlyArray<string>, options?: cp
  * @returns a boolean Promise
  */
 export async function isRPkgIntalled(name: string, msg?: string): Promise<boolean> {
-    const cmd = `cat(suppressPackageStartupMessages(require(${name}, quietly=TRUE)))`;
-    const isInstalled = await executeRCommand(cmd) === 'TRUE';
+     // users may write stdout in .Rprofile at the begining or the end of the session. so we need some strings to surround the output
+    const cmd = `cat(sprintf('vscodeRCheckPkgAvailability={%s}', require(${name}, quietly=TRUE)))`;
+    const rOut = /vscodeRCheckPkgAvailability=\{(.{4,5})\}/.exec(await executeRCommand(cmd));
+    if (rOut.length !== 2) {
+        // somehow the code failed to perform, so we don't throw (probably) false alert
+        return true;
+    }
+    const isInstalled = rOut[1] === 'TRUE';
     if (!isInstalled && msg !== undefined) {
         const repo = await getCranUrl();
         const rPath = await getRpath(false);
         void vscode.window.showErrorMessage(msg, 'Click to install', 'Not sure')
         .then(function(select) {
             if (select === 'Click to install') {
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -449,3 +449,28 @@ export function exec(command: string, args?: ReadonlyArray<string>, options?: cp
...
 }
+
+/**
+ * Check if an R package is available or not
+ *
+ * @param name the R package name that need to be checked
+ * @returns a boolean Promise
+ */
+export async function isRPkgIntalled(name: string, msg?: string): Promise<boolean> {
+    const cmd = `cat(suppressPackageStartupMessages(require(${name}, quietly=TRUE)))`;
```
renkun-ken, line 460: Looks like the method will break if user write anything to stdout in `.Rprofile`.

shrektan: You are right. Now the code is more robust. You can test this by adding the following lines to .Rprofile

```r
print("starting message.")
.Last <- function() {
  print("session end message.")
}
```

src/util.ts
```diff
@@ -449,3 +449,34 @@ export function exec(command: string, args?: ReadonlyArray<string>, options?: cp
...
+        return true;
+    }
+    const isInstalled = rOut[1] === 'TRUE';
+    if (!isInstalled && msg !== undefined) {
+        const repo = await getCranUrl();
+        const rPath = await getRpath(false);
+        void vscode.window.showErrorMessage(msg, 'Click to install', 'Not sure')
+        .then(function(select) {
+            if (select === 'Click to install') {
+                const args = [`--silent`, '--slave', `-e`, `install.packages('${name}', repos='${repo}')`];
```
eitsupi, line 474: It seems that only around `--slave` are single quotes, not back quotes.

shrektan: Fixed, despite the code still works, but they should be the single quotes.

renkun-ken:
@shrektan I rewrite some code in this PR:

* Checking `languageserver` installation at `createClient` to respect the working directory to be used (because different workspace folder might have different `.Rprofile`, e.g. renv might be used). Also, after installing languageserver, reopening an R file could activate the language service again.
* `executeRCommand` uses `spawnSync` instead of `execSync` to better handle the quoting of the arguments.
* `executeRCommand` now handles the case where user writes anything to stdout before and after the supplied expression is evaluated.

shrektan:
@renkun-ken Looks good to me. Much cleaner. Thanks!

PR #974, opened 2022-02-06
[data frame viewer] Add type of column to headerTooltip

eitsupi:
# What problem did you solve?

I was wondering if it was possible to display the column types in the data viewer.
I think it would be easier to implement and not spoil the look of the viewer if the type is displayed in the tooltip that appears when you mouse over the header.
https://www.ag-grid.com/javascript-data-grid/column-headers/#header-tooltips

## Screenshot

headertooltip

It may be a little unresponsive in some cases.

### DIFF
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -321,6 +321,7 @@ show_view <- !identical(getOption("vsc.view", "Two"), FALSE)
 if (show_view) {
   get_column_def <- function(name, field, value) {
     filter <- TRUE
+    tooltip <- typeof(value)
     if (is.numeric(value)) {
       type <- "numericColumn"
       if (is.null(attr(value, "class"))) {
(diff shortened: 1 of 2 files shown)


R/session/vsc.R
```diff
@@ -321,6 +321,7 @@ show_view <- !identical(getOption("vsc.view", "Two"), FALSE)
 if (show_view) {
   get_column_def <- function(name, field, value) {
     filter <- TRUE
+    tooltip <- typeof(value)
```
renkun-ken, line 324: What about 

```r
tooltip <- sprintf(
  "class: [%s], type: %s",
  toString(class(value)),
  typeof(value)
)
```

so that the tooltip contains the value of both `class()` and `typeof()`.

renkun-ken:
>It may be a little unresponsive in some cases.

Using `tooltipShowDelay: 100` seems to look much more responsive.

renkun-ken (approved):
LGTM thanks!

eitsupi:
Thank you for the quick review and improvement!

PR #979, opened 2022-02-07
Activate extension on subfolder

### LINKED ISSUE #976: Not able to trigger Web Help Page when there's no R file opened
## Reproducible Steps

1. Open any project in VSCode
2. Close all the R files and Web viewers
3. Reload the Window
4. Open a regular terminal, type `radian` or `R` to launch the R terminal
5. Type `?print`, you will find no response. Only after an R script is opened (probably only so it can trigger the R extension), then `?print` can open the help webpage. 

I know it's a very trivial issue but it confuses users when it occurs, as there's no webpage help or the terminal help shown. 

Is it possible to fix this?

Thanks.

renkun-ken:
# What problem did you solve?

Closes #976 

Now, if a subfolder of the workspace folder contains any `*.{rproj,Rproj,r,R,rd,Rd,rmd,Rmd}` file, then the extension will be activated.

1. Create a folder A and a subfolder B, and put a `file.R` in the subfolder B.
2. Open the folder A in vscode
3. The extension should be activated.


package.json
```diff
@@ -34,6 +34,7 @@
     "onLanguage:rmd",
     "onLanguage:debian-control.r",
     "workspaceContains:*.{rproj,Rproj,r,R,rd,Rd,rmd,Rmd}",
+    "workspaceContains:*/*.{rproj,Rproj,r,R,rd,Rd,rmd,Rmd}",
```
shrektan, line 37:
```suggestion
    "workspaceContains:**/*.{rproj,Rproj,r,R,rd,Rd,rmd,Rmd}",
```

renkun-ken: `**/*.{}` means recursive matching. For large and deep workspace folder, it might slow down the startup.

shrektan: OK. I think it's enough to match one subfolder layer only.

PR #984, opened 2022-02-09
R markdown templates

### LINKED ISSUE #839: [Feature request] R Markdown: Choose templates
It would be so great if we had an interactive command showing all the available Rmd skeleton templates that users can choose from.

Currently, RStudio IDE has this feature. As you know, all the Rmd templates path has the following structure:

`[pkg_name]/rmarkdown/templates/`

renkun-ken:
# What problem did you solve?

Closes #839 

This PR implements choosing template and create draft rmd from template via `rmarkdown::draft()`.

The templates are obtained via scanning the R package folders so that the information of all templates are written to a temp JSON file. When user picks a template, saveFileDialog will show up and user needs to find a location to save the file.

Some templates have multiple files. If user specify e.g. `draft.Rmd`, multi-file templates will create a new folder `draft` and then put files in it.

## (If you have)Screenshot

https://user-images.githubusercontent.com/4662568/153199351-3838bdfb-a2a5-44c5-be7d-2dbf752e07b6.mp4

1. "R Markdown: New Draft" will let user choose template and choose a location to save file.
2. Then a rmd file or a folder will be created.
3. The rmd file will be open in editor.

### DIFF
--- a/src/rmarkdown/draft.ts
+++ b/src/rmarkdown/draft.ts
@@ -6,14 +6,15 @@
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 
-import { QuickPickItem, QuickPickOptions, window, workspace } from 'vscode';
+import { QuickPickItem, QuickPickOptions, Uri, window, workspace } from 'vscode';
 import { extensionContext } from '../extension';
-import { getCurrentWorkspaceFolder, getRpath } from '../util';
+import { executeRCommand, getCurrentWorkspaceFolder, getRpath } from '../util';
 import * as cp from 'child_process';
 import * as path from 'path';
 import * as fs from 'fs';
 import * as os from 'os';
 import { readJSON } from 'fs-extra';
+import { join } from 'path';
 
 interface TemplateItem extends QuickPickItem {
   id: string;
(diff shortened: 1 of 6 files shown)


ManuelHentschel:
I'm not very familiar with rmd templates, but the ones that were found in my installed packages (rmarkdown, reprex) were only single-file documents. In these instances it might feel more natural to just create an unsaved editor with the content of the file, instead of having to save it somewhere.

Would it be feasible to do this for the packages that have `create_dir = false`?

renkun-ken:
Yes. In my initial implementation, I create untitled documents in all cases, but switched to save before creation when I test with `create_dir = true`. Using untitled documents for `create_dir = false` certainly makes good sense, and it is consistent with the behavior of RStudio. I'll implement this soon.

src/rmarkdown/draft.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export async function newDraft(): Promise<void> {
+    const cwd = getCurrentWorkspaceFolder()?.uri.fsPath ?? os.homedir();
+    const template = await launchTemplatePicker(cwd);
+    if (!template) {
+        return;
+    }
+
+    if (template.info.create_dir) {
+        const uri = await window.showSaveDialog({
```
ManuelHentschel, line 107: The `.Rmd` extension might be a bit confusing here. I did not find a really clean solution to configure the SaveDialog to specify a folder, but the following could be used:
``` ts 
        let defaultPath = path.join(cwd, 'draft');
        let i = 1;
        while(fs.existsSync(defaultPath)){
            defaultPath = path.join(cwd, `draft_${++i}`);
        }

        const uri = await window.showSaveDialog({
            defaultUri: Uri.file(defaultPath),
            filters: {
                'Folder': ['']
            },
            saveLabel: 'Create Folder',
            title: 'R Markdown: New Draft'
        });
```

The open dialogue seems to have an option to select only folders, but does not allow to select non-existing folders

src/rmarkdown/draft.ts
```diff
@@ -0,0 +1,134 @@
...
+    const options: QuickPickOptions = {
+        matchOnDescription: true,
+        matchOnDetail: true,
+        canPickMany: false,
+        ignoreFocusOut: false,
+        placeHolder: '',
+        onDidSelectItem: undefined
+    };
+
+    const items = await getTemplateItems(cwd);
```
ManuelHentschel, line 84: Here we could indicate that a folder needs to be created by e.g.:
``` ts
    for(const item of items){
        if(item.info.create_dir){
            item.label = (item.label || '') + ' $(list-tree) ' ;
        }
    }
```

src/rmarkdown/draft.ts
```diff
@@ -0,0 +1,134 @@
...
+    const args = [
+        '--silent',
+        '--slave',
+        '--no-save',
+        '--no-restore',
+        '-f',
+        rScriptFile
+    ];
+
+    let items: TemplateItem[] = [];
```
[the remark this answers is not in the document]
renkun-ken: Agree

src/rmarkdown/draft.ts
renkun-ken: Didn't know that icons could appear in the item label. `create_dir` is known in `getTemplateItems` when `label` is first specified. `$(new-folder)` might be more direct here.

Thanks for the tip!

src/rmarkdown/draft.ts
renkun-ken: Yes, it is something I'm also considering. I tried `showOpenDialog` with folder-only [options](https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions), but it didn't work smoothly for saving purpose.

renkun-ken: Also, `rmarkdown::draft()` will throw error if the target file or folder exists, which I have not handled yet.

renkun-ken: `rmarkdown::draft()` will create a folder with last file extension removed. For example, `draft.Rmd` will become `draft`; `draft.test.Rmd` will become `draft.test`.

I update the code to check if this folder already exists and let user confirm whether to remove it before creating the draft folder.

renkun-ken:
@ManuelHentschel Any more suggestions?

ManuelHentschel (approved):
Works nicely, thanks!

PR #985, opened 2022-02-09
Use spawn

renkun-ken:
# What problem did you solve?

This PR makes use of `spawn` and `spawnSync` across the entire project. In particular, `getAliases` is updated to use `spawnSync` to be consistent with others.

### DIFF
--- a/src/helpViewer/helpProvider.ts
+++ b/src/helpViewer/helpProvider.ts
@@ -270,9 +270,8 @@ export class AliasProvider {
     // call R script `getAliases.R` and parse the output
     private async getAliasesFromR(): Promise<undefined | AllPackageAliases> {
         const lim = '---vsc---';
-        const options: cp.ExecSyncOptionsWithStringEncoding = {
+        const options: cp.CommonOptions = {
             cwd: this.cwd,
-            encoding: 'utf-8',
             env: {
                 ...process.env,
                 VSCR_LIM: lim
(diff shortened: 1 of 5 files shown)


ManuelHentschel:
Thanks for the work. If I remember correctly, we only use a file to receive the results of `getAliases.R`, since there were some buffer overflow issues with `exec`. These should not be an issue with `spawn` according to e.g. https://stackoverflow.com/a/56688608, so I it might make sense to use stdout again. If you agree, I could implement that later today.

renkun-ken:
Personally, I prefer receiving small data directly from stdout whereas large data from a temp file. Here we use `spawn` and `spawnSync` for consistency across the project in terms of command argument quoting, shell wrapping, etc.

Does that make sense?

ManuelHentschel:
If there are no issues with buffer overflows, are there advantages to using a file over stdout? We load the entire file into memory at some point, anyways, and using temp files always has some potential to cause problems with cleanup, access restrictions, or even performance (on very limited systems).

src/helpViewer/helpProvider.ts
```diff
@@ -270,33 +268,54 @@ export class AliasProvider {
...
+        const args = [
+            '--silent',
+            '--slave',
+            '--no-save',
+            '--no-restore',
+            '-f',
+            this.rScriptFile
+        ];
+
+        return new Promise((resolve) => {
```
ManuelHentschel, line 291: Is there a reason that we show some errors to the user and log the errors in the childprocess/non-matching regex only to the console?

src/helpViewer/helpProvider.ts
```diff
@@ -270,33 +268,54 @@ export class AliasProvider {
...
+                    str += chunk.toString();
+                });
+                childProcess.on('exit', (code, signal) => {
+                    let result: AllPackageAliases | undefined = undefined;
+                    if (code === 0) {
+                        const re = new RegExp(`${lim}(.*)${lim}`, 'ms');
+                        const match = re.exec(str);
+                        if (match.length === 2) {
+                            const json = match[1];
+                            result = <{ [key: string]: PackageAliases }>JSON.parse(json) || {};
```
ManuelHentschel, line 305: The type could be expressed as `AllPackageAliases` as above

src/helpViewer/helpProvider.ts
```diff
@@ -270,33 +268,54 @@ export class AliasProvider {
...
-            // read and parse aliases
-            const txt = fs.readFileSync(tempFile, 'utf-8');
-            const json = txt.replace(re, '$1');
-            if(json){
-                allPackageAliases = <{[key: string]: PackageAliases}> JSON.parse(json) || {};
+    private async getAliasesFromR(): Promise<undefined | AllPackageAliases> {
+        const lim = '---vsc---';
+        const options: cp.ExecSyncOptionsWithStringEncoding = {
+            cwd: this.cwd,
+            encoding: 'utf-8',
```
ManuelHentschel, line 275: I think this encoding is ignored and the type could be changed to `cp.CommonOptions`

src/helpViewer/helpProvider.ts
renkun-ken: Now an error message is shown to user.

I'm wondering if it makes sense to use a global output channel to show details of such information? But we already have multiple output channels: R Language Server, R Markdown, and VSCode-R-Debugger also has R Debugger.

src/helpViewer/helpProvider.ts
renkun-ken: Yes thanks!

src/helpViewer/helpProvider.ts
renkun-ken: Done thanks!

src/helpViewer/helpProvider.ts
```diff
@@ -1,14 +1,12 @@
+/* eslint-disable @typescript-eslint/no-unsafe-return */
```
ManuelHentschel, line 1: I think this eslint-disable can be removed

src/helpViewer/helpProvider.ts
```diff
@@ -270,33 +268,42 @@ export class AliasProvider {
...
+            '--no-restore',
+            '-f',
+            this.rScriptFile
+        ];
+        
+        try {
+            const result = await spawnAsync(this.rPath, args, options);
+            if (result.status === 0) {
+                const re = new RegExp(`${lim}(.*)${lim}`, 'ms');
+                const match = re.exec(result.stdout);
```
ManuelHentschel, line 294: Just a stylistic suggestion to reduce nesting/improve readability:
``` ts
            const result = await spawnAsync(this.rPath, args, options);
            if (result.status !== 0) {
                throw result.error || new Error(result.stderr);
            }
            const re = new RegExp(`${lim}(.*)${lim}`, 'ms');
            const match = re.exec(result.stdout);
            if(match.length !== 2){
                throw new Error('Could not parse R output.');
            }
            const json = match[1];
            return <AllPackageAliases>JSON.parse(json) || {};
```

ManuelHentschel (approved):

src/helpViewer/helpProvider.ts
renkun-ken: Good suggestion. Looks better now.

Issue #987, opened 2022-02-12
Mixed quotes stop commented pipes from being ignored


gowerc:
Whilst looking through the code I noticed that there's a minor bug that commented out pipes aren't properly ignored if there are mixed quotes in command to be executed i.e.

```
x <- print("  'wadw'  ") # comment %>%
y
```
Results in y getting executed as well

Will submit PR shortly to fix

PR #988, opened 2022-02-12
Fix commented pipe bug

gowerc:
# What problem did you solve?

Resolved the mixed quote issue bug here: #987 
Renamed unit test file to better represent what they are testing.
Added new unit tests for cleanlines 

## (If you have)Screenshot

Before:
image

After:
image


renkun-ken (approved):
Works as expected.

LGTM

PR #989, opened 2022-02-12
WIP: Add problemMatching to task

gowerc:
# What problem did you solve?

Wanted to try and incorporate problemMatching into the default `R: test` task to make it easier to find test failures when testing a R package.  The change so far seems to be working (see screenshot below). That being said I'm not sure how to test this. I was thinking about setting up a dummy repo and running the task inside of it but I can't quite work out how to do that... I had tried

```
    test('workbench actions', async () => {
        const action = await vscode.commands.executeCommand('workbench.action.tasks.runTask', 'R: Test');
        .....
    });
```
But that doesn't seem to work. My next idea was to try and extract the regex patterns and see if they work aginst pre-defined error messages but I couldn't figure out the api for getting access to them. Any advice here would be appreciated !

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1670,6 +1670,27 @@
       {
         "type": "R"
       }
+    ],
+    "problemMatchers": [
+      {
+        "name": "testthat",
+        "owner": "R",
+        "severity": "error",
+        "fileLocation": ["relative", "${workspaceFolder}/tests/testthat"],
+        "pattern": [
+          {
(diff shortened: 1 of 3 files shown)


renkun-ken:
I test the implementation and it works as expected. When some test cases fail, the errors with proper message and location will show up in the diagnostics. When the problem is fixed and run tests again, the errors will disappear.

As for the tests of these tasks, I suggest that we could discuss it in a separate issue about how we could add more useful tests of vscode-R extension in general.

PR #992, opened 2022-02-13
Fix incorrect syntax highlighting for variables starting with "function"

### LINKED ISSUE #982: wrong symbol recognition when variable name starts with function
image

result should not be considered as a function

gowerc:
# What problem did you solve?

Closes #982  -  incorrect syntax highlighting for variables starting with the word `function`

- Resolved by adding a negative lookahead to the `function-declarations` file to make sure there are no additional word characters following the `function` keyword
- Added unit tests to show my code is working in the most basic settings
    - Unit tests work by reading in the regex pattern from the syntax file and then running them again pre-made strings.
    - Note that apparently vscode parses regex as POSIX but javascript doesn't so had to add a translation function.

Please let me know if there's a better way to test this or if you need me to refactor the code (am still learning node/ts/js/vscode).

## (If you have)Screenshot

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -43,7 +43,10 @@
       "outFiles": [
         "${workspaceFolder}/out/test/**/*.js"
       ],
-      "preLaunchTask": "npm: pretest"
+      "preLaunchTask": "npm: pretest",
+      "env": {
+        "EXTENSION_ROOT": "${workspaceRoot}"
+      }
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


gowerc:
Looks  like the unit tests have failed with:
> TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined

I'm assuming this is because the environment variable `process.env.EXTENSION_ROOT` isn't getting populated. I'm then guessing that this means the CICD job isn't using the same launch configuration as is being used by testing in vscode ?

gowerc:
@renkun-ken , fixed the bug with pathing for getting the config file, this is ready for review now. Please let me know if there is a better way of handling this.

renkun-ken (approved):
LGTM thanks!

PR #1022, opened 2022-03-03
Enhance workspace viewer

### LINKED ISSUE #1018: Show Environment Package Stack
I know there is `View(globalenv())` or `R: Preview Environment` from the command pallette.
Is there any way to show the loaded packages in the environment; I am referring to this fly-out from R Studio:

renkun-ken:
# What problem did you solve?

Closes #1018

This PR renames the global environment watcher to workspace watcher and includes `search()` and `loadedNamespaces()` in the writing of `workspace.json` instead of the previous `globalenv.json`. The following are improved:

* The workspace viewer contains more information of the R workspace: attached namespaces, loaded namespaces, and global environment symbols.
* The global environment symbols now get proper icons consistent with their completion item icons.
* Clicking the `package:foo` items in Attached Namespaces or items in Loaded Namespaces will show quick pick of help topics of that package.

## (If you have)Screenshot

https://user-images.githubusercontent.com/4662568/156767723-aad57c1a-b02b-4c21-90bd-a4c662111f81.mp4

### DIFF
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -208,7 +208,7 @@ inspect_env <- function(env, cache) {
         info$slots <- slotNames(obj)
       }
 
-      if (is.list(obj) && !is.null(dim(obj))) {
+      if (!is.null(dim(obj))) {
         info$dim <- dim(obj)
       }
     }
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -221,25 +221,26 @@ inspect_env <- function(env, cache) {
 dir_session <- file.path(tempdir, "vscode-R")
 dir.create(dir_session, showWarnings = FALSE, recursive = TRUE)
 
-removeTaskCallback("vsc.globalenv")
+removeTaskCallback("vsc.workspace")
 show_globalenv <- isTRUE(getOption("vsc.globalenv", TRUE))
-if (show_globalenv) {
-  globalenv_file <- file.path(dir_session, "globalenv.json")
-  globalenv_lock_file <- file.path(dir_session, "globalenv.lock")
-  file.create(globalenv_lock_file, showWarnings = FALSE)
+workspace_file <- file.path(dir_session, "workspace.json")
+workspace_lock_file <- file.path(dir_session, "workspace.lock")
+file.create(workspace_lock_file, showWarnings = FALSE)
 
-  update_globalenv <- function(...) {
-    tryCatch({
-      objs <- inspect_env(.GlobalEnv, globalenv_cache)
-      jsonlite::write_json(objs, globalenv_file, force = TRUE, pretty = FALSE)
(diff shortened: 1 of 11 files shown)


renkun-ken:
I'm trying to give those attached packages and loaded namespaces commands so that clicking them will trigger a quick pick of topics for that package. But it does not seem to be easily achievable with the existing help manager. @ManuelHentschel  Any thoughts?

ElianHugh:
Thanks for the PR, this is super cool :) I will have a look at this in a few hours - the main thing I'd like to check is the liveshare functionality

renkun-ken:
> I'm trying to give those attached packages and loaded namespaces commands so that clicking them will trigger a quick pick of topics for that package. But it does not seem to be easily achievable with the existing help manager. @ManuelHentschel Any thoughts?

Resolved.

ElianHugh:
I tried with LiveShare and everything seems to work nicely :) 

Small nitpick - would it make sense to have the 'Global Environment' node expanded instead of closed by default? At present, it requires a click upon startup. This would be closer to the RStudio implementation of showing the global env by default

renkun-ken:
>Small nitpick - would it make sense to have the 'Global Environment' node expanded instead of closed by default? At present, it requires a click upon startup. This would be closer to the RStudio implementation of showing the global env by default

It is expanded by default. See https://github.com/REditorSupport/vscode-R/pull/1022/files#diff-dc026a6295ecf05710e5e88e39d5a1a77d2d67c0a8e21b3ec2e23a6b97b985a1R37.

renkun-ken:
Now attached packages in "Loaded Namespaces" will display "attached" in the description:

image

renkun-ken:
Now if `r.session.watchGlobalEnvironment` is disabled, the "Global Environment" node will not show in the workspace viewer.

ElianHugh:
The tooltip for the namespaces can be a little superfluous, e.g.
image

Would it be viable to have the result of `packageDescription(package)[["Description"]]` as the tooltip?

renkun-ken:
>Would it be viable to have the result of packageDescription(package)[["Description"]] as the tooltip?

I thought about this. To avoid overhead, we don't want to write the package description to `workspace.json` repeatedly. The package information could be obtained from `globalRHelp` but it looks like the package nodes are only populated when the nodes need to be displayed. Therefore, in the workspace viewer, I delayed forcefully populating those populated nodes to when it is clicked.

If we put the package description in the tooltip, then we need to forcefully populate `globalRHelp`'s package index more early, when the workspace viewer needs to populate tree items. Let me see if it works nicely.

renkun-ken:
> Would it be viable to have the result of packageDescription(package)[["Description"]] as the tooltip?

Done via ac44caace4a6a470fbfc774d531bc0ae9034d3bb.

PR #1060, opened 2022-04-04
Hide smart knit env values to prevent accidental deletion

ElianHugh:
Fix for  https://github.com/REditorSupport/vscode-R/issues/860.

Prevent the accidental deletion of env values necessary for smart knitting. 

# What problem did you solve?

Smart knitting will fail if the user attempts to delete the env values used for knitting. This PR makes it harder for users to accidentally delete these values.

Try to knit the following:

``````
```{r}
rm(list = ls(all.names = TRUE))
```
``````


renkun-ken (approved):
Works nicely! LGTM.

PR #1068, opened 2022-04-09
Add `r.session.data.pageSize`

### LINKED ISSUE #1067: Add a setting for number of rows per page in the data viewer
As suggested in #1066, it should be useful for some user to change the number of rows per page (default 100) in the data viewer.

It is allowed via <https://www.ag-grid.com/angular-data-grid/row-pagination/#example-customising-pagination> in the ag-grid API.

We could add a vscode setting (e.g. `r.session.data.pageSize`) for this.

renkun-ken:
# What problem did you solve?

Closes #1067 

Adjust `r.session.data.pageSize` to 100, 500 and 0 and `View(data.frame(id = 1:1000))` and see if the setting applies properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1498,7 +1498,7 @@
         },
         "r.session.data.pageSize": {
           "type": "integer",
-          "default": 100,
+          "default": 500,
           "markdownDescription": "The maximum number of rows per page in the data viewer. `0` means to disable pagination."
         },
         "r.session.viewers.viewColumn": {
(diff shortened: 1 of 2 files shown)


eitsupi:
Just a question: what good does it do to reduce the page size? (display performance?)
My understanding is that all data is read regardless of page size, so the initial value for page size seems like it should be larger.

renkun-ken:
I think it only affects the rendering performance, as larger page size means more objects in the web page. Also, if the page size is too large, the scrolling will be harder to locate certain rows.

Here, the default value 100 comes from ag-grid. Maybe we could use 500.

PR #1084, opened 2022-04-22
Take out http from `Insert image` Rmd snippet

jooyoungseo:
# What problem did you solve?

Currently, `Insert image` Rmd snippet assumes that the image link should be a remote URL. However, this is overgeneralization because local image can b also used. Thus, I've taken out the http from the snippet.

Try using `Insert image` snippet in any Rmd file.


renkun-ken (approved):
LGTM

PR #1085, opened 2022-04-26
Take out http from link snippet

jooyoungseo:
# What problem did you solve?

similar to #1084: link could be a relative path or even local path. Also `https` could be used. Thus, do not fix it as `http://`.

In Rmd file, try `Insert link` snippet.


renkun-ken (approved):
LGTM

PR #1108, opened 2022-05-24
Rewriting nested ifs

LucasFA:
It seems to me there's some unnecessary "if" nesting happening, but I'm a beginner, feel free to shut me down or explain.

IIRC similar patterns occur through the codebase. If this is a reasonable change, I'd have no problem tracking those down.


LucasFA:
I'll reopen after fixing in my branch

PR #1112, opened 2022-05-25
Add `Create .linr` command

### LINKED ISSUE #712: Add "Create .lintr" command
How about adding a command to add a `.lintr` file to the workspace, similar to the command that already exists to add `.gitignore`?
(Perhaps a function in the {lintr} or {languageserver} package would be more appropriate ?)

When I used vscode-R and R LSP for the first time, I struggled a lot because I did not know how to write a `.lintr` file.

eitsupi:
# What problem did you solve?

Close #712

Add a simple command that simply calls [the `lintr::use_lintr` function](https://lintr.r-lib.org/reference/use_lintr.html) that will be added in lintr 3.0.0.

Select "Create .linr to the workspace" from the command palette, with the development version of `lintr` package installed.

### DIFF
--- a/src/lintrConfig.ts
+++ b/src/lintrConfig.ts
@@ -1,0 +1,27 @@
+'use strict';
+
+import { existsSync, unlinkSync } from 'fs';
+import { join } from 'path';
+import { window } from 'vscode';
+import { runTextInTerm } from './rTerminal';
+import { getCurrentWorkspaceFolder } from './util';
+
+export async function createLintrConfig(): Promise<void> {
+    const currentWorkspaceFolder = getCurrentWorkspaceFolder()?.uri.fsPath;
+    if (currentWorkspaceFolder === undefined) {
+        void window.showWarningMessage('Please open a workspace folder to create .lintr');
+        return;
+    }
+    const lintrFilePath = join(currentWorkspaceFolder, '.lintr');
+    if (existsSync(lintrFilePath)) {
+        const overwrite = await window.showWarningMessage(
+            '".lintr" file already exists. Do you want to overwrite?',
+            'Yes', 'No'
+        );
+        if (overwrite === 'No') {
+            return;
+        }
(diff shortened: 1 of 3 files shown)


ElianHugh:
I think now that we are moving toward more file creation functions, after this is merged we should look at relocating the files to a dedicated folder

src/lintrConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+        const overwrite = await window.showWarningMessage(
+            '".lintr" file already exists. Do you want to overwrite?',
+            'Yes', 'No'
+        );
+        if (overwrite === 'No') {
+            return;
+        }
+        void unlinkSync(lintrFilePath);
+    }
+    void runTextInTerm('lintr::use_lintr()');
```
renkun-ken, line 26: Should we create lintr config file using a background process rather than sending command to terminal?

renkun-ken:
Currently, a `.lintr` is created silently and nothing happens when it is created.

Does it make sense to open the lintr file once it is created? 

For example, we could capture the path of the created `.lintr`

```r
normalizePath(lintr::use_lintr())
```

and open it in vscode when the background process is finished.

eitsupi:
> Does it make sense to open the lintr file once it is created?

I think it maybe a bit noisy.
The `remote-containers.createDevContainerFile` command that adds the devcontainer definition files to the project is the vscode command that I often use to add files, but this also does not open the file after it is added.

renkun-ken (approved):
LGTM Thanks!

renkun-ken:
> > Does it make sense to open the lintr file once it is created?
> 
> I think it maybe a bit noisy. The `remote-containers.createDevContainerFile` command that adds the devcontainer definition files to the project is the vscode command that I often use to add files, but this also does not open the file after it is added.

OK sounds reasonable. Let's keep it silent for now.

PR #1121, opened 2022-06-04
Update data viewer column resizing

### LINKED ISSUE #1120: cannot display calculated value in data viewer
**Describe the bug**
Hey everyone, I'm new to vscode-R and I'm so appreciate for making such excellent plugin.
The problem I found out is that the data viewer only display original value after read a dataset.
For example, 
say I have a 75-columns dataset "aaa", I used "read_csv" to load it then calculated a new "mean" column.
When I view it only the original 75 columns displayed, and the "mean"column were somehow "hidden".
If I want to see the calculated value I need to drag it.
Is this an issue?

**To Reproduce**

[aaa.csv](https://github.com/REditorSupport/vscode-R/files/8829197/aaa.csv)
aaa <- read_csv("aaa.csv", na = c("999", "NA"))
aaa$mean <- mean(aaa$o8)
スクリーンショット 2022-06-03 11 54 19
スクリーンショット 2022-06-03 11 56 22

Thanks in advance!

renkun-ken:
# What problem did you solve?

Closes #1120

`suppressColumnVirtualisation: true` does not look like a good idea mainly for performance reasons. Setting `minWidth` and `width` seems to be enough for most purposes.

```r
n <- 1000
df <- data.frame(id = 1:n)
for (i in 1:100) {
  df[[paste0("abcdef", i)]] <- rnorm(n)
}

View(df)
```

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -507,8 +507,10 @@ export async function getTableHtml(webview: Webview, file: string): Promise<stri
             sortable: true,
             resizable: true,
             filter: true,
+            width: 100,
+            minWidth: 50,
             filterParams: {
-            buttons: ['reset', 'apply']
+                buttons: ['reset', 'apply']
             }
         },
         columnDefs: data.columns,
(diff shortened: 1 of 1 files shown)


ElianHugh:
LGTM, thanks! Works as expected on Arch

Issue #1137, opened 2022-06-27
Code/Text in Help Window can't be copied


iago-pssjd:
**Describe the bug**
Text/Code in Help Window can't be copied and pasted.

**To Reproduce**
1. Search the help of any function, e.g., `?t.test`
2. On the help for the function, try to copy something, e.g., `t.test(1:10, y = c(7 : 20))`
3. Try to paste it in the terminal, in the editor or in any other place.

## Do you want to fix by self? (We hope your help!)

No

**Expected behavior**
We can Copy text/code From R Help View Window and Paste in Source Code Document

**Screenshots**
![scs](https://user-images.githubusercontent.com/40892925/175891600-a3d15cbf-99d5-4077-aed4-6523aaa4ebe7.png)

**Environment (please complete the following information):**

- OS:  Windows 10
- VSCode Version: 1.68.1
- R Version: 4.2.0
- vscode-R version: v2.5.0

**Additional context**
Related issues: 
- Very similar to https://github.com/REditorSupport/vscode-R/issues/1062
- Actually, I would prefer to apply feature requested in https://github.com/REditorSupport/vscode-R/issues/459, but don't having it, I have to copy and paste.

ManuelHentschel:
I can't reproduce this bug on windows 10, vscode 1.68.1, vscode-R 2.5.0. Are you aware of any other settings etc. that might influence this?

iago-pssjd:
I am not aware of that. I only modified the next settings to use `radian`:
```
 "r.bracketedPaste": true,
  "r.rterm.windows": ...
```

PR #1138, opened 2022-06-27
Click code in help views

### LINKED ISSUE #459: Help view: send code to console?
Since the new help view (v1.6.0) (which is great @ManuelHentschel) includes hyperlinks and highlighting, I was wondering if it would be technically possible to send code from the examples to the active R session?

ManuelHentschel:
Closes #459 #1137

Code sections in the help viewer are highlighted on hover, and can be clicked to be copied to the clipboard, or run in the terminal.

The behavior can be modified with the setting `"r.helpPanel.clickCodeExamples"`. Changes to this setting might require reloading the window to take effect.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1464,6 +1464,16 @@
           "default": true,
           "markdownDescription": "Show links to matching help pages in hover"
         },
+        "r.helpPanel.clickCodeExamples": {
+          "type": "string",
+          "default": "Copy",
+          "markdownDescription": "What happens when clicking code examples on help pages.",
+          "enum": [
+            "Disabled",
+            "Copy",
+            "Run"
+          ]
+        },
         "r.source.encoding": {
           "type": "string",
           "default": "UTF-8",
--- a/package.json
+++ b/package.json
@@ -1825,6 +1825,7 @@
     "vscode:prepublish": "tsc -p . && tsc -p ./html/help && tsc -p ./html/httpgd",
     "compile": "tsc -p . && tsc -p ./html/help && tsc -p ./html/httpgd",
     "watch": "tsc -p . --watch",
+    "watchHelp": "tsc -p ./html/help --watch",
     "pretest": "tsc -p ./",
     "test": "node ./out/test/runTest.js",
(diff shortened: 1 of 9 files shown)


src/helpViewer/panel.ts
```diff
@@ -264,6 +265,18 @@ export class HelpPanel {
...
 			console.log(`Message (text): ${String(msg.text)}`);
+		} else if(msg.message === 'codeClicked') {
+			const code = String(msg.code || '');
+			console.log(`Code clicked: ${code}`);
+			const codeClickAction = config().get<CodeClickAction>('helpPanel.clickCodeExamples');
+			if(codeClickAction === 'Disabled' || !code){
+				// pass
+			} else if(codeClickAction === 'Copy'){
+				void vscode.env.clipboard.writeText(code);
+				void vscode.window.showInformationMessage(`Copied to clipboard: \`${code}\`.`);
```
renkun-ken, line 276: The message seems to look too messy in many cases:

image

package.json
```diff
@@ -1464,6 +1464,21 @@
...
           "markdownDescription": "Show links to matching help pages in hover"
         },
+        "r.helpPanel.clickCodeExamples": {
+          "type": "string",
+          "default": "Copy",
+          "markdownDescription": "What happens when clicking code examples on help pages.",
+          "enum": [
+            "Disabled",
+            "Copy",
+            "Run"
```
renkun-ken, line 1474: It is nice to have the capability to either copying or running the example code. I wonder if it is possible to have both at the same time. For example, click to copy, and cmd+click to run, or something like this?

ManuelHentschel: Good idea. I wasn't sure how to handle the possible [modifier keys](https://github.com/reditorsupport/vscode-r/blob/23aa7ca821956e6dd86330e3b44789c800b28431/html/help/script.ts#L100-L103) best, especially w.r.t. mac (Should we just make them all configurable, or are there sensible "groupings"?).

renkun-ken (approved):
Works nicely! Thanks!

PR #1142, opened 2022-07-03
Whitespace in typescript files

ManuelHentschel:
Converts indentation in all ts files (where necessary) to 4 spaces and adds an eslint rule to enforce this indentation. Approximately 90% of ts files followed this convention anyways, so I assumed this change would be rather uncontroversial.

This PR should not change any actual code behavior.

Enforcing linter rules for e.g. R and json files might also make sense, but should be discussed in a separate issue/PR.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -34,3 +34,4 @@ rules:
   indent:
     - 'error'
     - 4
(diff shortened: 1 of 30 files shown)


renkun-ken (approved):
LGTM

Issue #1130, opened 2022-06-14
How to automatically context switch bettween different conda environment R and radians depending on currently opened file?


hermidalc:
I'm new to VSCode (just moved from Atom because it's being retired), and I'm trying to understand how to set up the R extension to work seamlessly with conda environments. Each of the project folders in my workspace is associated with its own conda environment, each with it's own R and radian builds/versions and libraries.

How can I set up the R extension to use the appropriate R, radian, and libraries for the project file I have currently open in the editor? VSCode can do this with python and I've set that up, so every time I open a code file from a specific project it just context switches to the appropriate conda environment, python, and libraries.

I want to do the same with R. But when I open an R file in a project folder, in the status bar it shows `R: (not attached)`. Cannot figure out how to fix this

elsherbini:
This is going to be the blind leading the blind, so if someone else can chime in that'd be great.

But what happens when you open the command prompt (F1) and run the command R: Create R terminal ?

What settings do you have for the vscode-R extension for Rterm path?

One way forward would be to make a wrapper script and point the Rterm path to that, something like:

~/activate_local_conda_and_start_R.sh
```
#!/bin/bash

cd $(dirname $0)

# First we need to init conda bc ~/.bashrc is not executed when starting non-interactive shells. 
# Assuming PATH was already set by parent shell, and reusing script in profile.d/ if possible...
__conda_init_file=$(dirname $(which conda))/../etc/profile.d/conda.sh
if [ -f "$__conda_init_file" ]; then
    source "$__conda_init_file"
else
    eval "$(conda shell.bash hook 2> /dev/null)"
fi    

# Now we can activate
conda deactivate && conda activate ./

# And start R, passing through all command-line arguments
R --no-save --no-restore
```

hermidalc:
> But what happens when you open the command prompt (F1) and run the command R: Create R terminal ?

```
Cannot find R client at radian. Please check r.rterm setting.
```

My `r.rterm` is the default `radian` without a path, as it should be. `vscode-R` should be smart like the python VSCode design and I should be able to select an interpreter per project folder from an auto populated list of conda environments and python binary paths that VSCode figured out. This is how code dev should be because each project has it's own dependencies/versions/builds. The same for R code files and projects with R code.

Why didn't the `vscode-R` devs just use the exact same design as was done for the python side? That design just makes sense and works beautifully, and I can have different files in different projects open simultaneously and it will automatically switch to the appropriate binaries and env based on the file in focus.

[...]

Issue #747, opened 2021-08-09
Can't start with the R in conda environment


yangmean:
**Describe the bug**
I firt start the R with click 'R: Create R terminal', and it always start with R in conda base environment. Howerver, I want to start in another conda environment not base. R version in base was R 3.6.3, but in r4-base environment was R 4.0. So I don't know how to change the setting to activate the R 4.0. My remote system was Centos7.

(If related)setting.json
{
    "git.ignoreLegacyWarning": true,
    "jupyter.askForKernelRestart": false,
    "workbench.colorCustomizations": {
    "editorError.foreground": "#cf2121",
    "editorWarning.foreground": "#00000000",
    "editorInfo.foreground": "#00000000"
    },
    "r.sessionWatcher": true,
    "r.rterm.linux": "/Data1/Users/xxx/Software/anoconda2/envs/r4-base/bin/radian",
    "r.bracketedPaste": true,
    "terminal.integrated.profiles.linux": {
        "bash": {
            "path": "bash",
            "icon": "terminal-bash"
        },
        "zsh": {
            "path": "zsh"
        },
        "fish": {
            "path": "fish"
        },
        "tmux": {
            "path": "tmux",
            "icon": "terminal-tmux"
        },
        "pwsh": {
            "path": "pwsh",
            "icon": "terminal-powershell"
        },
        "sh": {
            "path": "/bin/sh"
        }
    },
    "terminal.integrated.defaultProfile.linux": "sh",
    "r.alwaysUseActiveTerminal": true
}
- OS: [e.g. Windows, macOS, Linux]
- VSCode Version: [e.g. 1.59.0]
- vscode-R version: [e.g. v2.1.0]

**Additional context**

andycraig:
Hi @yangmean, here's how to handle this case:

1. Add this to your `settings.json`:
    ```json
    "r.alwaysUseActiveTerminal": true
    ```
2. When you want to use R, create a new terminal (_not_ a new R terminal), then activate your conda environment and start R from the command line like this:
    ```bash
    conda activate YOUR_CONDA_ENVIRONMENT_NAME # probably: conda activate r4-base
    R
    ```

hermidalc:
Sorry I'm new to VSCode (just moved from Atom), and I'm trying to understand how to set up the R extension to work with mamba/conda environments. Each of my project folders is associated with its own mamba/conda environment, each with it's own R and radian. How can I set up the extension to use the appropriate R and radian for the project file I have open?

[...]

Issue #696, opened 2021-07-05
Create or provide choice for multiple R versions and x64 or i386 architectures for R terminal


gwd666:
**Is your feature request related to a problem? Please describe.**
No problem only a feature request ... 
When hitting CTRL+P the `R: Create R terminal` command opens up the VSC Terminal and starts a R session based on the Rpath (in my case WINdows) setting, but only allows for one distinct R architecture (either x64 or i386) and R version 

Because of (mainly database client) driver architecture issues on Windows (ODBC etc) this means a lot of manual hassle with extension `Settings` in case one would like/need to switch to R i386 architecture (in my case b/c x64 is usually my default) or vice versa. 

**Describe the solution you'd like**
Choice between 
`R: Create R-x64 terminal ` 
`R: Create R-i386 terminal ` 

Maybe even side by side? 

**Describe alternatives you've considered**
None so far - I am open for workarounds (eg keyboard shortcuts or similar). 

**Additional context**
NA

PS: this extension of yours ROCKS!

renkun-ken:
Have you tried typing in the terminal to start R-x64 and R-i386 terminal and do both work?

gwd666:
Hi 
Sorry I do not fully understand your proposal or hint. 
If I execute a fully qualified path to R_HOME\bin\i386\R.exe or ...x64\R.exe versions in the Terminal the respective R version will start up, yes. 
I was just hoping for a small enhancement to VSC R to be available when hitting the CTRL+P keyboard shortcut. 
![image](https://user-images.githubusercontent.com/57595487/124469974-4c9da000-dd9b-11eb-8b1d-b2142ef8edf4.png)

gwd666:
OTHERWISE ... I forgot to mention that for R DEV work I can or usually always do "tweak" the  Workspace R extension Setting accordingly! It is mainly when I use VSC interactively ie when I use that package (with its 32 bit driver dependency) that I tend to forget or need to switch the R-ARCH in the "User" Settings.

ManuelHentschel:
I can 100% understand where you're coming from, but I'm not quite sure what the best approach would be here. E.g. when a new R version is released which I want to try, but not switch to permanently (yet), I'm in a similar situation as you, but not between architectures but R versions.

Also, e.g. in https://github.com/REditorSupport/vscode-R/issues/597 we're looking to simplify the path settings, to make using the extension easier for new users, and adding support for different architectures/versions is somewhat contradictory to this.

[...]

PR #1143, opened 2022-07-04
R executable selection

ElianHugh:
# WIP R executable selection

Major contribution: extended R executable selection. Related issues #696 #1130 #1148 #747 

This PR may need to be split into two parts: (1) R executable selection and (2) improved multi-root support

## Screenshots

image

image

image

### R Executable

- [x] Language status item
- [x] Executable quick pick
- [x] Unix support for standard R installations
- [ ] Unix support for conda, mamba
- [x] Windows support for standard R installations
- [ ] Windows support for conda, mamba
- [ ] Look into better supporting renv

### Multi-root

Multi-root executable support potentially requires modification of many vscode-R features

- [x] R terminal
- [ ] Language service
- [ ] Help provider
- [x] Tasks

Other features should probably be deferred to another PR


src/executables/locator/shared.ts
```diff
@@ -0,0 +1,36 @@
...
+export function getArchitectureFromPath(path: string): string {
+    if (process.platform === 'win32') {
+        // \\bin\\i386 = 32bit
+        // \\bin\\x64 = 64bit
+        return '';
+    } else {
+        return '64-bit';
+    }
+}
+
```
ManuelHentschel, line 30: Would it make sense to use `R.version` (inside R) for this?

ElianHugh: That would definitely be more straightforward. The main reason I avoided that was because we would then need to spawn an R process for every single executable found

ElianHugh: I wonder what the load is like for Rscript? Alternatively, it looks like the R version is also exposed for windows by the `Rversion.h` header file in /R/include/

src/executables/ui/status.ts
```diff
@@ -0,0 +1,59 @@
...
+import { RExecutableService } from '../service';
+
+enum BinText {
+    name = 'R Language Indicator',
+    missing = '$(warning) Select executable'
+}
+
+export class ExecutableStatusItem implements vscode.Disposable  {
+    private readonly service: RExecutableService;
+    private languageStatusItem: vscode.LanguageStatusItem;
```
ManuelHentschel, line 13: vscode 1.65.0 seems to be the lowest version that supports this api

src/executables/service/locator/shared.ts
```diff
@@ -0,0 +1,72 @@
+import { execSync } from 'child_process';
+import * as fs from 'fs-extra';
+import * as vscode from 'vscode';
+
+export function getRDetailsFromPath(rPath: string): {version: string, arch: string} {
+    try {
+        const child = execSync(`${rPath} --version`).toString();
```
ManuelHentschel, line 7: This gives the following error on windows:
`'C:\Program' is not recognized as an internal or external command, operable program or batch file.`

src/executables/service/locator/windows.ts
```diff
@@ -0,0 +1,102 @@
...
+                    }
+
+                    if (fs.existsSync(x64)) {
+                        registryBins.push(x64);
+                    }
+                }
+            });
+        }
+
+        return registryBins;
```
ManuelHentschel, line 60: If I'm not mistaken, this returns an empty array that gets populated in an async manner later on.

src/executables/ui/status.ts
ElianHugh: Right, I believe it's definitely a newer addition. I'll look into creating an adapter class of some sort for older versions

src/executables/service/locator/shared.ts
ElianHugh: Good catch, thank you. I haven't had a chance to develop on windows so I think the windows side is definitely lagging behind the unix side of things

src/executables/ui/status.ts
ManuelHentschel: We can just change the minimum version of `@types/vscode` and `"engines": {"vscode": ...}` to 1.65, I'd suggest.

ElianHugh:
I'll have access to windows later today, so I should be able to work on the windows side to get to feature parity

ElianHugh:
An issue I've found that on windows side, if an R executable isn't openable it causes an annoying popup when loading the extension. It doesn't look like fs.access or fs.stat checks for this on windows

ElianHugh:
This is currently how the picker is looking:

image

Recommendation is due to the renv lock file denoting a matching R version

ElianHugh:
Coming back to this PR, I think it's probably okay if support for "all" virtual environments is scaled back to just conda and renv?

eitsupi:
> if support for "all" virtual environments is scaled back to just conda and renv?

I think it's almost the same as conda, but also mamba?

ElianHugh:
I'm reasonably happy with the executable UI & the executable locator services. main issue for this PR right now are conda activation (seems a little fragile in terms of how I'm handling it) and also feature parity on windows devices

floriandeboissieu:
I'm waiting eagerly for the R executable selection with conda/mamba support. Any chance to have this PR finalized & merged in the near future?

ElianHugh:
Hi @floriandeboissieu, thank you for your interest in the PR. Unfortunately this year has been pretty chock-a-block and I haven't had much time to do open source projects unless they are directly related to work. Having said that, I will have plenty of free time over the Christmas break, so I'm hoping to jump back into things then.

ElianHugh:
Thank you all for the feedback. I am closing this PR in favour of work on a new branch (#1473).

Issue #999, opened 2022-02-15
Differentiate methods/functions or fields in autocompletion for R6 class.


Fred-Wu:
[mlr3](https://github.com/mlr-org/mlr3) is implemented with R6 class. However, after creating a task, all autocompletes appear to be fields. Some of them actually should be methods, such as `data` in the screenshot below. 

<img width="495" alt="image" src="https://user-images.githubusercontent.com/4111978/154001387-fb1e7434-8a49-43b6-8b19-1aa1bc720f36.png">

Can these be differentiated within VSCode?

Fred-Wu:
Just would like to check whether it is something easy to change?

renkun-ken:
Currently, we write out the `names()` of each object in the global environment on each top-level user input. To achieve this, we'll need to not only access `names()` but check if each child is a function, which could be a significant overhead given the way we gather such information at the moment.

The main difference here is only accessing `names()` or calling `typeof()` with each `name`.  Here is some benchmarking:

```r

get_names <- function(obj) {
  names(obj)
}

get_names_and_types <- function(obj) {
  names <- names(obj)
  vapply(names, function(name) {
    typeof(obj[[name]])
  }, character(1L))
}
```

For an object with many elements (e.g. `baseenv()`), the performance difference is like

```r
env <- baseenv()
get_names(env)
get_names_and_types(env)

microbenchmark::microbenchmark(
  get_names(env),
  get_names_and_types(env)
)
```

```
Unit: microseconds
                     expr      min        lq      mean  median        uq      max neval
           get_names(env)  598.524  655.3905  738.7634  709.28  758.0825 2769.102   100
 get_names_and_types(env) 1433.744 1519.0875 1819.3571 1733.98 1804.9030 6086.458   100
```

But for small objects like a typical R6 classes, the performance difference is quite significant:

```r

tsk <- mlr3::tsk("iris")
microbenchmark::microbenchmark(
  get_names(tsk),
  get_names_and_types(tsk)
)
```

```
Unit: microseconds
                     expr      min        lq       mean    median       uq      max neval
           get_names(tsk)    1.646    2.1385    4.17186    2.7675    5.080   32.734   100
 get_names_and_types(tsk) 1662.417 1993.9890 2212.53392 2040.4430 2310.729 4910.143   100
```

If our environment contains many objects like this, there will be obviously a notable delay on each user input given the current session watcher mechanism.

ElianHugh:
[...]

Issue #837, opened 2021-10-24
Dynamic dataframe loading when using View() 


danielbasso:
**Is your feature request related to a problem? Please describe.**
Using `View()` on large dataframes causes the system to crash or become slow since it has to load all data into a html template.

**Describe the solution you'd like**
Use pagination to load only rows that are on screen. Filtering and sorting should be handled "server-side". 

**Describe alternatives you've considered**
I use things like `View(my_dataframe[1:100,])`, but it compromises filtering and sorting.

**Additional context**

Example: 

```r
dataframe_size <- 5000000 # 5mi rows

my_dataframe <- data.frame(a = rnorm(dataframe_size),
                 b = rnorm(dataframe_size),
                 c = rnorm(dataframe_size),
                 d = rep("test", dataframe_size))

View(my_dataframe)
```
Screenshot:

![image](https://user-images.githubusercontent.com/20959177/138598565-504f1d2e-9068-45f5-bee1-4d397a5771cb.png)

I'd like to help/implement, if possible. I already took a look at` vsc.R` and `session.ts` since the Ag Grid update (#708). Tips/advice are welcomed.

litn2018:
any update about this? it should be implemented as lazy loading, like RStudio implementation.

eitsupi:
As a workaround, it is now possible to set the maximum number of rows to be loaded into the data viewer (#945).
In my environment, I think setting it to about 100,000 rows is comfortable to use.

![image](https://user-images.githubusercontent.com/50911393/152667251-d471d373-0404-477c-8d9a-3e92ce04bca9.png)

danielbasso:
@eitsupi Had a huge problem with 2FA and almost lost my account. But I'm back.

Thank you for letting me know about the workaround. I've talked to @ElianHugh a while ago about this, and he gave me the idea of making the solution completly optional, to not disturb the current functionality.

I'm trying to find some time to delve into this issue.

d-golzato:
Hi, thanks a lot for your huge work on this extension. Vscode complete noob here!

I'm having troubles opening and working interactively with big dataframes (> 1,000,000 rows). The data viewer is quite slow at opening these dataframes and when it does, nothing is shown in the respective preview editor. For small
dataframes there are no particular problems.

From my enabled _R:Session watcher_ and set the option _Row limit_ to 50.

Do you have any update about this feature/fix?

Thanks!

danielbasso:
Hi @d-golzato, how are you? 

[...]

Issue #440, opened 2020-11-06
R session background connections


andycraig:
In #394, `notebook.R` runs a socket connection that the notebook communicates with. Would it be worth extracting part of this out so that a socket connection is available for each R session started by the user? This would make the following possible:

- Sending VS Code's R commands to the session invisibly, rather than having to send them as text to the console
- Querying the session, for example to get the names of data frames
- For `Run Selection/Line`, potentially using R instead of TypeScript to choose which lines to send (allowing use of R's parse functions)
- Inline evaluation (sending user code to the session invisibly, and displaying the results in the editor) - this one's not really necessary but some other languages/IDE's have it and it's kind of cool

@renkun-ken  You wrote `notebook.R` and the session watcher, which has some overlap with this proposal, so you're definitely the best person to comment. Does this sound practical? Let me know if I need to clarify anything.

ManuelHentschel:
I've tried adding similar functionality to the debugger in https://github.com/ManuelHentschel/VSCode-R-Debugger/issues/94.

To use it, start an R session (any terminal works), create some variables etc., and then run `vscDebugger::.vsc.listen()`. To connect vscode to that session, launch the R debugger with the following launch config:
``` jsonc
    {
      "type": "R-Debugger",
      "request": "attach",
      "name": "Attach to R process",
      "splitOverwrittenOutput": true
    }
```

This will allow you to view the current workspace, the callstack (if you called .vsc.listen while inside a function), and evaluate arbitrary code in the debug console (autocomplete should work here, but seems to be a bit unstable). If the R session was launched inside a VSCode terminal, it will also allow settings breakpoints and using the flow control buttons to a limited extent.

One of the biggest challenges with this approach is that the R process will be idle while showing the input prompt. During that time it won't be able to handle any commands from vscode. AFAIK the session watcher uses `addTaskCallback()` to update the session info after each command, this could of course also be used to listen/write to a socket, but we would still have the problem that R only listens to either the user or vscode at any time.

I'm not sure if it would be possible to spawn a child process that continuously listens on a socket and still has access to the workspace of the parent process, but this does sound rather difficult if not impossible.

---

My opinion regarding the features you suggested:

> * Sending VS Code's R commands to the session invisibly, rather than having to send them as text to the console
> * Inline evaluation (sending user code to the session invisibly, and displaying the results in the editor) - this one's not really necessary but some other languages/IDE's have it and it's kind of cool

[...]

PR #1151, opened 2022-07-17
Websocket communication

renkun-ken:
This is a very rough demo of using httpuv server in user R session and websockets on vscode side for communication.

If `r.session.useWebSocket` is enabled, a httpuv server is started in the background of an interactive R session, and vscode-R session watcher will create a WebSocket to connect to the server on attach.

The hover and completion of `$` and `@` are roughly implemented.

And it works on both local and remote development.

This makes it possible to address #440, #594, #837, #999.

https://user-images.githubusercontent.com/4662568/228568579-e6eedbed-f27d-413a-b302-c8681569842d.mp4


ElianHugh:
pretty neat! I tried this with an R6 object and it worked nicely. this could also have implications for liveshare requests

Fred-Wu:
Just would like to check whether there is any update and plan to release this implementation. Thanks.

renkun-ken:
Completion and hover are improved to support recursive `$` and `@`.

ElianHugh:
Works as expected for

```r
a <- list(
    b = list(
        c = list(
            d = 5L
        )
    )
)
```

and 

```r
x <- setClass("student", slots = list(name = "character", age = "numeric", GPA = "numeric"))
```

---

Would it be worth allowing the socket ip address to be set by the user? Currently it's set to 127.0.0.1. 
Also, is it possible to use unix domain sockets or windows pipes via `httpuv::startPipeServer`? I'm not sure if it returns the same messages as the standard socket server function

renkun-ken:
>Would it be worth allowing the socket ip address to be set by the user? Currently it's set to 127.0.0.1.

Is there a use case for setting an ip other than `127.0.0.1`? Since the session watcher itself still relies on a local filesystem to work properly, I'm not sure if it is useful to allow a websocket to attach to an IP other than `127.0.0.1`.

> Also, is it possible to use unix domain sockets or windows pipes via httpuv::startPipeServer? I'm not sure if it returns the same messages as the standard socket server function

The main point of using a httpuv websocket server in user R session and websocket in vscode-R extension here is that it does not block the user session and provides a simple event loop mechanism. I'm not sure if the bare bone `socketServer` or `socketConnection` could do the same?

Also, I'm thinking of using the [Infinite Row Model](https://www.ag-grid.com/javascript-data-grid/infinite-scrolling/) to allow for dynamic loading of data frame in the viewer via WebSocket or http socket directly talking to the R session that creates that data viewer.

renkun-ken:
>In 99% of situations, I'd say no, unless for whatever reason only IPv6 is allowed.

Good catch. Didn't think of/meet such scenario before. I'm curious how other R packages that start a web server (e.g. shiny) handle this?

>I've used socketConnection in the past, iland haven't run into major issues. Not sure of the major difference between it and httpuv admittedly. I might fiddle with the different socket implementations to see the differences

If there is a good way to handle non-blocking communication between client and server via sockets, I'd be happy to see how we could adjust our implementation.

>Sounds like a good idea! I'll have a closer look at the PR tonight.

Thanks, just take your time!

The dynamic loading of data frames will require another PR for certain, as I'm just playing around with the ag-grid api at <https://github.com/REditorSupport/vscode-R/compare/master...renkun-ken:vscode-R:websocket-data-viewer> but looks like there is more work such as server-side handling of filtering and ordering.

ElianHugh:
So, I tried using httpuv::startPipeServer, and it appeared to work as expected.

I couldn't find much on IPv6 in R, so it might just be not super important ATM. I tried "localhost" but it looks like httpuv doesn't resolve names so no luck there.

renkun-ken:
I switch to using just `httpuv::startServer()` to start a non-blocking http server and use `fetch()` on extension side. This could make it easier to implement dynamic data loading as we could directly access the R session's http server via resource uri inside the data viewer webview.

renkun-ken:
Let's merge this as an experimental feature first and see if it works properly under different scenarios.

ElianHugh:
Sorry for the delay! I think that's a fair idea, we can get more feedback that way. Similar to how we handled rstudioapi and the session watcher

Issue #1166, opened 2022-07-31
[Cannot use quarto::quarto_render in custom knit]: unused argument (encoding = "UTF-8")


jooyoungseo:
CC @ElianHugh

When users attempt to use `quarto::quarto_render` as a custom knit, it fails because `encoding="UTF-8"` argument is automatically passed.

### test.Rmd

````
---
title: test
output: html_document
knit: quarto::quarto_render
---

```{r}
print("Hello world!")
```


````

ElianHugh:
thank you for the report, should have a fix by tonight

PR #1167, opened 2022-08-01
Remove encoding param from knitting

ElianHugh:
# What problem did you solve?

Solves #1166. We currently pass an encoding param to the knitting render process, which can cause issues when a given renderer does not have an encoding parameter. The [encoding parameter of `rmarkdown::render` is also ignored by rmarkdown](https://pkgs.rstudio.com/rmarkdown/reference/render.html) anyway, meaning that there is no point to passing the parameter.

Knitting the following will fail on the live version of vscode-r

``````
---
title: Test
knit: quarto::quarto_render
---

Test.

``````


renkun-ken (approved):
LGTM

PR #1188, opened 2022-09-03
Reload help pages on help refresh

ManuelHentschel:
This PR adds a `refresh()` method to the `HelpPanel` class that reloads the currently shown help page (and marks cached history entries as stale). This method is now called as part of the `Clear Cache & Restart Help Server` command meaning that is no longer necessary to manually close and re-open help pages after installed packages change.

How to test:
Open a help page for package X, install a different version of X, click `Clear Cache & Restart Help Server` (this might not even be necessary if the code in [helpServer.R](https://github.com/REditorSupport/vscode-R/blob/master/R/help/helpServer.R) notices the change), check that the shown help page updates.

### DIFF
--- a/html/help/script.js
+++ b/html/help/script.js
@@ -13,21 +13,12 @@ window.onmousedown = (ev) => {
         scrollY: window.scrollY
     });
 };
-// handle back/forward requests from vscode ui
-// simulates a mousclick on key 3 or 4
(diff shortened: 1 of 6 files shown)


renkun-ken (approved):
Works nicely! LGTM

PR #1192, opened 2022-09-10
Avoid .R file lock on windows

ManuelHentschel:
This PR avoids a file lock issue (observed on windows, probably not an issue on linux): Sourcing .R file using `R -f xxx.R` locks the file and as a consequence it's impossible to reinstall this extension while it is running. This is particularly annoying when working on the extension itself, since it contains R code and is therefore activated automatically.

This problem is solved by sourcing .R files using `R -e source('xxx.r')`, which does not appear to lock the file.

### DIFF
--- a/src/helpViewer/helpProvider.ts
+++ b/src/helpViewer/helpProvider.ts
@@ -47,7 +47,10 @@ export class HelpProvider {
 
         // starts the background help server and waits forever to keep the R process running
         const scriptPath = extensionContext.asAbsolutePath('R/help/helpServer.R');
-        const scriptExpression = `"base::source('${scriptPath.replaceAll('\\', '\\\\')}')"`;
+        const scriptExpression = (process.platform === 'win32' ? 
+            `"base::source('${scriptPath.replaceAll('\\', '\\\\')}')"` :
+            `base::source('${scriptPath}')`
+        );
         const args = [
             '--silent',
             '--slave',
--- a/src/helpViewer/helpProvider.ts
+++ b/src/helpViewer/helpProvider.ts
@@ -47,17 +47,15 @@ export class HelpProvider {
 
         // starts the background help server and waits forever to keep the R process running
         const scriptPath = extensionContext.asAbsolutePath('R/help/helpServer.R');
-        const scriptExpression = (process.platform === 'win32' ? 
-            `"base::source('${scriptPath.replaceAll('\\', '\\\\')}')"` :
-            `base::source('${scriptPath}')`
-        );
         const args = [
             '--silent',
             '--slave',
             '--no-save',
             '--no-restore',
             '-e',
-            scriptExpression
+            'base::source(base::commandArgs(TRUE))',
+            '--args',
+            scriptPath
         ];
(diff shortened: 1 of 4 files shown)


src/helpViewer/helpProvider.ts
```diff
@@ -47,14 +47,14 @@ export class HelpProvider {
 
         // starts the background help server and waits forever to keep the R process running
         const scriptPath = extensionContext.asAbsolutePath('R/help/helpServer.R');
-        // const cmd = `${this.rPath} --silent --slave --no-save --no-restore -f "${scriptPath}"`;
+        const scriptExpression = `"base::source('${scriptPath.replaceAll('\\', '\\\\')}')"`;
```
renkun-ken, line 50: The quoting here does not work on Linux and macOS. The following works, but `replaceAll` might not be useful in the first place.

```typescript
const scriptExpression = `base::source('${scriptPath.replaceAll('\\', '\\\\')}')`;
```

renkun-ken (commented):
Not sure if it is a good idea to use `source()` instead of direct supplying a file. Chaning the content of an R file while it is running by Rscript might cause the R session to fail sometimes.

ManuelHentschel:
> Not sure if it is a good idea to use `source()` instead of direct supplying a file. Changing the content of an R file while it is running by Rscript might cause the R session to fail sometimes.

This problem would be solved by using `source()`, since `source` first parses the entire file and then starts executing. Therefore, background processes like the languageserver and helpserver do not depend on the original file at all, once they start running. IMO this is the ideal behavior. In contrast, on windows `Rscript` locks the file, making it impossible to quickly reinstall the extension.

On Linux/macos, I'm not sure how exactly this is handled by Rscript. Apparently it is possible to write a script that deletes itself (and does something afterwards) and run it with Rscript, but I'm not sure what the limits to this are (i.e. whether Rscript always reads the entire file first, or I just got lucky and my test script was short enough to be read at once).

src/helpViewer/helpProvider.ts
```diff
@@ -47,14 +47,17 @@ export class HelpProvider {
 
         // starts the background help server and waits forever to keep the R process running
         const scriptPath = extensionContext.asAbsolutePath('R/help/helpServer.R');
-        // const cmd = `${this.rPath} --silent --slave --no-save --no-restore -f "${scriptPath}"`;
+        const scriptExpression = (process.platform === 'win32' ? 
+            `"base::source('${scriptPath.replaceAll('\\', '\\\\')}')"` :
```
renkun-ken, line 51: What if we use something like

```typescript
const args = [
            '--silent',
            '--slave',
            '--no-save',
            '--no-restore',
            '-e',
            'source(commandArgs(TRUE))',
            '--args',
            scriptPath
        ];
```

where the argument quotation is handled internally, not sure if it works on Windows?

ManuelHentschel: Very elegant and works nicely, thanks!

renkun-ken (approved):
LGTM

PR #1194, opened 2022-09-10
Fix empty code examples

ManuelHentschel:
This PR fixes a small issue with help pages where an empty line at the end of code would be highlighted as if it was a clickable example.


renkun-ken (approved):
LGTM

PR #1200, opened 2022-09-16
devTasks

ManuelHentschel:
This PR removes two compiled .js files (keeping only their source .ts files) and updates the vscode tasks so that they generate those files for debugging/compilation.


renkun-ken (approved):
LGTM

renkun-ken:
The following files are generates as untracked:

```
html/help/script.js
html/httpgd/index.js
```

@ManuelHentschel Maybe we should ignore these files in a gitignore?

ManuelHentschel:
> The following files are generates as untracked each time I debug the extension:
> 
> ```
> html/help/script.js
> html/httpgd/index.js
> ```
> 
> @ManuelHentschel Maybe we should ignore these files in a gitignore?

Done in https://github.com/REditorSupport/vscode-R/commit/7ddd0d55129ac5007de348923b217877320fca44

renkun-ken:
Seems `html/session/webview/observer.js` is ignored too. Not sure if it should be?

ManuelHentschel:
> Seems `html/session/webview/observer.js` is ignored too. Not sure if it should be?

Files that are already tracked (like `observer.js`) shouldn't be affected by new .gitignore entries. You can check that e.g. by modifying the file and running `git status` (which correctly shows `observer.js` for me).

renkun-ken:
OK. Thanks for reminding!

PR #1221, opened 2022-10-12
Add language support for NAMESPACE & .Rbuildignore

nx10:
Edit: not sure if "language support" is the right terminology. Maybe "syntax highlighting grammars and configurations" would be correct.

This PR adds language support for `NAMESPACE` and `.Rbuildignore` R package files according to the specification: 

- Namespaces: https://cran.r-project.org/doc/manuals/R-exts.html#Package-namespaces 
  - Written from scratch
  - Visually differentiates imports and exports and highlights im-/exported symbols
- R ignore: https://cran.r-project.org/doc/manuals/R-exts.html#Building-package-tarballs
  - JavaScript Regex grammar is dynamically imported
  - Much easier to visually parse lines and not wrongly detected as XML anymore

## Screenshots

image

image

## Note

(I am currently participating in the [hacktoberfest](https://hacktoberfest.com/). If you like this PR consider to tag it with `hacktoberfest-accepted` .)


renkun-ken:
Thanks for working on this?

Looks like `NAMESPACE` file is more like R syntax, just like the following:

* https://github.com/r-lib/lintr/blob/main/NAMESPACE
* https://github.com/Rdatatable/data.table/blob/master/NAMESPACE

where both do some R version checking.

Is there a reason we don't just use R syntax and extend a bit if helpful?

nx10:
> Thanks for working on this?
> 
> Looks like `NAMESPACE` file is more like R syntax, just like the following:
> 
> * https://github.com/r-lib/lintr/blob/main/NAMESPACE
> * https://github.com/Rdatatable/data.table/blob/master/NAMESPACE
> 
> where both do some R version checking.
> 
> Is there a reason we don't just use R syntax and extend a bit if helpful?

Interesting, the documentation does mention version checks, I wonder if general R code is allowed in these files. 
I will see if I can link with the R grammar for everything that is not an import/export directive.

As for the use-case: I find myself looking at NAMESPACE files solely to figure out what symbols a package imports and exports, which I tried to simplify with this. I realize that this is subjective so it's perfectly fine if you don't want to merge this feature.

The problem with extending the R grammar is that the NAMESPACE directives are legal symbols in R:

```R
> S3method
Error: object 'S3method' not found
> S3method <- "asd"
> S3method 
[1] "asd"
> import
Error: object 'import' not found
> import <- 12
> import
[1] 12
```

so it would be wrong to highlight them differently in R code files.

renkun-ken:
It seems that functions in `baseenv()` are allowed. We might need to add highlighting for `if else`, string literal and simple function calls.

I tried something like the following and it works, but anything like `utils::capture.output` won't work.

```r
export(fun1)
export(fun2)
export(fun3)

if (findInterval(1, c(0, 1)) >= 1) {
  export(fun4)
}
```

[`parseNamespaceFile()`](
https://github.com/wch/r-source/blob/2008918f89a741cc8e214d014d41d242b0cc31e7/src/library/base/R/namespace.R#L1248) is called on package installation to parse `NAMESPACE` file.

renkun-ken (approved):
Let's merge this first and revisit if we need further improvements.

nx10:
Thanks! It's not easy to to fall back to the R grammer for everything but specific functions as the grammar is "hungry" and once a scope starts (i.e. with `{`, `(`) it will completely take over. I am pretty sure there is a way to do it but I could not find it in the short time I tried.

Issue #1239, opened 2022-10-27
More devtools commands for R package


jooyoungseo:
The following `devtools` commands are frequently used for R package develpment. It would be nice if they could be added to R package commands.

- [ ] `devtools::release()`
- [ ] `devtools::spell_check()`
- [ ] `devtools::check_rhub()`
- [ ] `devtools::check_win_devel()`
- [ ] `devtools::dev_mode()`

PR #1242, opened 2022-10-27
Close 1239: More devtools and usethis commands for R package development

jooyoungseo:
# What problem did you solve?

I have added the following commands which are frequently used for R package developpment:

- [x] `devtools::dev_mode()`

- [x] `devtools::spell_check()`

- [x] `devtools::check_rhub()`

- [x] `devtools::check_win_devel()`

- [x] `devtools::release()`

- [x] `usethis::use_version()`

- [x] `usethis::use_cran_comments()`

- [x] `usethis::use_news_md()`

- [x] `usethis::use_git()`

- [x] `usethis::use_github()`

- [x] `pkgdown::build_site()`

I did not use task because these commands require interaction.

- [ ] 

## (If you have)Screenshot

N/A

In the command palette, use the following commands:

* r.devMode
* r.spellCheck
* r.checkRhub
* r.checkWinDevel
* r.release
* r.useVersion
* r.useCranComments
* r.useNewsMd
* r.useGit
* r.useGitHub
* r.pkgdownBuildSite

### DIFF
--- a/package.json
+++ b/package.json
@@ -447,6 +447,11 @@
         "command": "r.loadAll"
       },
       {
+        "title": "Release",
+        "category": "R Package",
+        "command": "r.release"
+      },
+      {
         "title": "Build",
         "category": "R Package",
         "command": "r.build"
@@ -1767,7 +1772,7 @@
           "default": false,
           "description": "Default boolean value for automatically sharing R browser ports with guests."
         },
-        "r.plot.devArgs" :{
+        "r.plot.devArgs": {
           "type": "object",
           "markdownDescription": "The arguments for the png device to replay user graphics to show in VSCode. Requires `#r.plot.useHttpgd#` to be set to `false`. \n\nChanges the option `vsc.dev.args` in R.",
           "default": {
--- a/package.json
+++ b/package.json
@@ -447,6 +447,11 @@
         "command": "r.loadAll"
       },
       {
+        "title": "Toggle Development Mode",
+        "category": "R Package",
+        "command": "r.devMode"
+      },
+      {
         "title": "Spell check",
         "category": "R Package",
         "command": "r.spellCheck"
--- a/package.json
+++ b/package.json
@@ -467,6 +467,11 @@
         "command": "r.useCranComments"
       },
       {
+        "title": "Build {pkgdown} Web Site",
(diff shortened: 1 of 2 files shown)


ManuelHentschel:
I agree that these are useful functions to have available, but I'm not sure this is the best way to implement them. If I'm not mistaken, most (or all) of these commands are independent of the current R session and should therefore better be run in a separate process, not the user's R terminal. Furthermore, users might want to customize the specific commands being run by passing arguments to the functions called.

A more intuitive way to implement this behaviour would be the vscode tasks API, e.g. as follows:
``` jsonc
{
  "type": "R",
  "code": [
    "devtools::install(quick=TRUE, upgrade=FALSE, keepSource=TRUE)"
  ],
  "group": "build",
  "problemMatcher": [],
  "label": "R: Install"
}
```

jooyoungseo:
Thanks for the suggestions. These commands that I am adding here require interactions. For example, `devtools::release()` asks a series of questions and users need to interactively and manually respond. Does task job allow these kinds of interactions?

jooyoungseo:
#1239

renkun-ken:
Is there a criterion that determines which one goes into the list of commands while others don't?

ManuelHentschel:
> Thanks for the suggestions. These commands that I am adding here require interactions. For example, `devtools::release()` asks a series of questions and users need to interactively and manually respond. Does task job allow these kinds of interactions?

In general, tasks can be interactive. To check this, you can try e.g. the following task definition (requires the github cli installed, but that can be replaced by any other interactive command).
``` json
{
  "type": "shell",
  "label": "test",
  "command": "gh repo create"
}
```

Implementing this for R commands could be a bit more difficult, since R usually goes into non-interactive mode if you call it using e.g. `R -e ...`, but should be possible.

ManuelHentschel:
@jooyoungseo  I implemented a proof of concept for interactive R tasks in #1267. Please let me know if this works to run the tasks you have in mind.

jooyoungseo:
Thanks for the task implementation, @ManuelHentschel ! It looks good to me. However, we may want to keep `devtools::dev_mode()` in current R session following its design principle so that users can toggle on and off the dev_mode.

PR #1257, opened 2022-11-11
fixed broken tasks problemMatcher regex

gowerc:
# What problem did you solve?

Problem matcher for testthat wasn't working as it doesn't always return a column. I updated the regex syntax to make capturing the column optional

## (If you have)Screenshot

Example of testthat not returning a column number:
image

Updated Regex now capturing this error properly again:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1911,7 +1911,7 @@
         ],
         "pattern": [
           {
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #644, opened 2021-05-25
WIP: Preview help pages

ManuelHentschel:
Resolve #637 

**What problem did you solve?**
Adds support to preview the .Rd files in the ./man directory of the current project.

**How to check this pull request?**
This is WIP, currently the regular help viewer is replaced by this preview functionality.
To check, open a folder containing an R package and click "Open Help Topic using `?`" in the R: Help Pages view.

Note: this PR is based on https://github.com/Ikuyadeu/vscode-R/pull/641, hence the large diff.

Issue #637, opened 2021-05-08
Preview .Rd


ElianHugh:
**Is your feature request related to a problem? Please describe.**

During package development, its sometimes nice to preview what the documentation of your package looks like. RStudio has a neat thing where you can open a .Rd file and preview what the help file would look like. 

**Describe the solution you'd like**

We could have a preview feature similar to RStudio. Just as Rmd files have a knit button, we could have a preview button for Rd files. 

**Describe alternatives you've considered**
The help viewer is the alternative, but it can't (I think?) access local files due to permission issues.

**Additional context**
The commands listed [here](https://stat.ethz.ch/R-manual/R-patched/library/tools/html/Rd2HTML.html) seem useful for this, particularly
- parse_rd
- Rd2HTML

ElianHugh:
Thanks for the reply @ManuelHentschel ! I'd be happy to look at this, but like you I wouldn't be able to get to it right away :)

Once I get around to it, I might shoot you a question or two if I get stuck!

PR #1259, opened 2022-11-15
Implement help preview for local package(s)

ManuelHentschel:
Addresses #637, replaces #644.

This PR implements a preview feature for the help pages based on .Rd files in the current directory:
- If a requested help-topic-path matches one of the present .Rd files, this file is converted to html and shown instead of the original help page
- The R side bar contains a new item which lists all of the help topics from local .Rd files
- If local files change, help previews are updated accordingly
- The setting "r.helpPanel.previewLocalPackages" controls which (sub)directories are checked for local packages. By default this is just the main workspace folder (`["."]`), but subdirectories can be specified. The feature can be disabled by setting this to `[]`.

image

image

Some possible future improvements are:
- ~Preview meta files (DESCRIPTION, Index)~
- Toggle between installed help/preview per page etc. (see https://github.com/REditorSupport/vscode-R/pull/1259#issuecomment-1324890174)
- Fix encoding issue on non-utf8 locales
- Performance improvements (?)

### DIFF
--- a/src/helpViewer/index.ts
+++ b/src/helpViewer/index.ts
@@ -14,12 +14,15 @@ import {
     DummyMemento,
     getRPathConfigEntry,
     escapeHtml,
+    makeWebviewCommandUriString,
+    uniqueEntries,
 } from '../util';
 import {HelpPanel} from './panel';
 import {HelpProvider, AliasProvider} from './helpProvider';
 import {HelpTreeWrapper} from './treeView';
 import {PackageManager} from './packages';
 import {isGuestSession, rGuestService} from '../liveShare';
+import { makePreviewerList, RHelpPreviewerOptions, RLocalHelpPreviewer } from './helpPreviewer';
 
 export type CodeClickAction = 'Ignore' | 'Copy' | 'Run';
 export interface CodeClickConfig {
@@ -79,6 +82,7 @@ export async function initializeHelp(
     try {
         rHelp = new RHelp(rHelpOptions);
     } catch (e) {
+        console.log('Error while launching R Help:', e);
         void vscode.window.showErrorMessage(`Help Panel not available`);
     }
 
@@ -145,13 +149,21 @@ export interface HelpFile {
     scrollY?: number
     // used to open the file in an external browser
     url?: string
+    // indicates that this is a preview generated from a .Rd file
+    isPreview?: boolean;
+    // the .Rd file that this is based on (if it is a preview)
+    rdPath?: string;
+    // if available, the .R file from which the documentation is generated
+    rPath?: string;
+    // if available, the directory of the previewed R package
+    packageDir?: string;
 }
 
 // Internal representation of an "Alias"
 export interface Alias {
-    // name of a help topic as presented to the user
+    // main name of a help topic 
     name: string
-    // name of a help topic as used by the help server
+    // one of possibly many aliases of the same help topic
     alias: string
     // name of the package the alias is from
     package: string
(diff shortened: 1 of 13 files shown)


src/helpViewer/index.ts
```diff
@@ -620,6 +703,28 @@ function pimpMyHelp(helpFile: HelpFile): HelpFile {
...
+            const rdUri = vscode.Uri.file(helpFile.rdPath);
+            const cmdUri = makeWebviewCommandUriString('vscode.open', rdUri);
+            rdInfo = `<a href="${cmdUri}" title="Open File">${localRdPath}</a>`;
+        } else{
+            rdInfo = `an .Rd file`;
+        }
+        if(helpFile.rPath){
+            const localRPath = vscode.workspace.asRelativePath(helpFile.rPath);
+            const rUri = vscode.Uri.file(helpFile.rPath);
+            const cmdUri = makeWebviewCommandUriString('vscode.open', rUri);
```
renkun-ken, line 720: The command uri here does not work for me.

When I click the hyperlink to the file, vscode cannot find the file:

image

ManuelHentschel: I added a check, that makes sure the linked file actually exists. Do you still get failing links with the latest commit?

renkun-ken: I tried with the same package (RMariaDB), and it works on macOS and Linux. But on Windows, it never works for me. The file does exists, but it seems vscode does not accept the uri as a valid one.

image

image

ManuelHentschel: The file path vscode is trying to open in the lower pane seems to be an absolute, linux-style path (`home/ken/Workspaces/RMariaDB/R/default.R`), which surprises me, since it should be a relative path, as in the top pane (`./R/default.R`).

Does the top of the corresponding Rd file say `% Please edit documentation in R/default.R` or is there a different path in there?

Do the path variables [here](https://github.com/REditorSupport/vscode-R/blob/06b13fb23c52cdacc4139b521b7a6823f18e7667/src/helpViewer/index.ts#L719-L722) point to the expected places (`helpFile.rPath`: absolute path of R file, `localRPath`: same file, but local w.r.t. workspace root, `rUri`: vscode.Uri of the same file)?

(I checked with a fresh clone of RMariaDB on windows and all links are working for me)

renkun-ken: Forgot to tell, It is in Remote WSL.

renkun-ken: I tried in Windows and everything works nicely. The problem only occurs in Remote WSL.

ManuelHentschel: The issue was apparently caused by the fact that the vscode-R extension runs on the remote machine (WSL) and webviews run on the UI machine (windows) and therefore the URI embedded in the help page was interpreted as a windows one. I added a custom command instead of the default open command, which should fix this.

renkun-ken: It works nicely on WSL now 👍

renkun-ken:
When I'm trying with RMariaDB, I wonder why `R/MariaDBResult.R`, `R/dbIsValid_MariaDBResult.R` do not have links to the files in the workspace in the following help page 

image

ManuelHentschel:
> When I'm trying with RMariaDB, I wonder why `R/MariaDBResult.R`, `R/dbIsValid_MariaDBResult.R` do not have links to the files in the workspace in the following help page
> 

I wasn't aware that there can be multiple R file references in a single .Rd file. Should be fixed with the last commit.

renkun-ken:
Thanks for the quick fix! Another one is following where the workspace folder appears for some reason?

image

ManuelHentschel:
> Thanks for the quick fix! Another one is following where the workspace folder appears for some reason?

Turns out there can even be line breaks in the list of related R files. Should work with the latest commit.

renkun-ken:
The original entries become the following

image

renkun-ken:
Also, "Open Help Topic using `?`" will have duplicate entries regarding the package under development:

image

ManuelHentschel:
I noticed that there is an issue with special characters if R's default encoding on the machine is not utf-8. If you know a simple way to e.g. set R's stdout to utf-8, please let me know.

renkun-ken (approved):
LGTM Thanks!

PR #1273, opened 2022-12-07
Make help preview async

ManuelHentschel:
Makes the generation of help page previews from .Rd files asynchronous to avoid blocking vscode, if the generation takes longer.

I don't know a reproducible example to illustrate this, but on slower machines this PR should eliminate one (possible) source of vscode delays.


renkun-ken (approved):
LGTM

PR #1278, opened 2022-12-17
Add the column name to the tooltip on data viewer

eitsupi:
# What problem did you solve?

Related to #1121
When a data frame with many columns is displayed in the data viewer, the column names may not be readable due to column width limitations.
Display the column name in the tooltip so that we can see the column name.

## (If you have)Screenshot

image

```r
n <- 1000
df <- data.frame(id = 1:n)
for (i in 1:100) {
  df[[paste0("loooooooooooooooooooooooong", i)]] <- rnorm(n)
}

View(df)
```

### DIFF
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -326,7 +326,8 @@ if (show_view) {
   get_column_def <- function(name, field, value) {
     filter <- TRUE
     tooltip <- sprintf(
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM

eitsupi:
Thank you for merging!

PR #1286, opened 2023-01-03
Ability to set echo=TRUE in Run source by default

jakub-jedrusiak:
# What problem did you solve?

The only option to run source with `echo=TRUE` was to use `ctrl+shift+enter`. I've added an option to set `echo=TRUE` by default which is the default behaviour in RStudio and might be convenient for some users.

## Screenshot

image


renkun-ken (approved):
LGTM Thanks!

PR #1293, opened 2023-01-26
Add new command for load all with reset = FALSE

kummerer94:
# What problem did you solve?

When working with a large codebase in R, it is handy to use `devtools::load_all(reset = FALSE)`. It will be much faster and usually do just as good a job as the full `devtools::load_all()`.

I would like to have a similar way to run this as when I press `Ctrl + Shift + L`. However, when I want to add a key binding for this, I need a command.

So I tried to replicate the code that was used for adding the `devtools::load_all()` command (`r.loadAll`) and added a `r.loadAllCached` command.

I am happy to make any additional changes but I have to say that I know absolutely nothing about VS code extensions.

I am unsure if I added this command in the right way. I tried replicating what was done to add `r.loadAll` and hope I found the right places to add this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -479,6 +479,11 @@
         "command": "r.loadAll"
       },
       {
+        "title": "Load All (reset = FALSE)",
+        "category": "R Package",
+        "command": "r.loadAllCached"
+      },
+      {
         "title": "Build",
         "category": "R Package",
         "command": "r.build"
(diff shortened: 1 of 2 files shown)


renkun-ken:
Thanks for submitting a PR!

It might be better done via adding a keyboard shortcuts on your side. Please take a look at the [Creating keybindings for R commands](https://github.com/REditorSupport/vscode-R/wiki/Keyboard-shortcuts#creating-keybindings-for-r-commands).

kummerer94:
Thank you, @renkun-ken, for this suggestion! I did not know that was an option.

What I want can thus be achieved like this:

```json
[
  {
    "key": "ctrl+l",
    "command": "r.runCommand",
    "args": "devtools::load_all(reset=FALSE)",
    "when": "editorTextFocus && editorLangId == 'r'"
  }
]
```

PR #1302, opened 2023-02-15
Fix broken tests

ElianHugh:
Setting `knitDir` moved to the constructor of the knit manager, preventing early access of the `config` function.


renkun-ken (approved):
Great fix!

Issue #990, opened 2022-02-13
Unit and integration tests


renkun-ken:
Currently, we only have very limited testing of the features of vscode-R.

We might need to put a demo R package and R script folder in the repo and run tests with opening the folder as a workspace folder and test at least the following features:

* Language server startup
* Session watcher
* R markdown commands

gowerc:
Just want to say that I think given the size of this project now that it would be valuable to add more tests mostly for the obvious reasons

- Tests act as proof that someones contribution is working as expected
- Tests prevent regressions and alert us to breakages in the package

That being said testing in vscode does not appear to be straight forward. One issue in particular seems to be that you [can't switch between workspaces](https://stackoverflow.com/a/69021044/4651564) (which I figured would be critical for us testing different setups/configurations).

Having said that it seems to be easy enough to launch the tests on a single specific directory though by editing the `launch.json` file to specify the target directory:
```
{
      "name": "Extension Tests",
      "type": "extensionHost",
      "request": "launch",
      "runtimeExecutable": "${execPath}",
      "args": [
        "${workspaceFolder}/test/sample-project",   // <--------------------------
        "--extensionDevelopmentPath=${workspaceFolder}",
        "--extensionTestsPath=${workspaceFolder}/out/test/suite/index"
      ],
      "outFiles": [
        "${workspaceFolder}/out/test/**/*.js"
      ],
      "preLaunchTask": "npm: pretest"
    }
```

Potentially one idea could be to have multiple test scripts each loading a separate directory and then placing conditions on test suites to only fire if a certain directory is active ?

Likewise I'm happy to help writing some of these tests as it will be a good opportunity to learn, that being said I will need some help / pointers as I've really been struggling to get my head around what is even possible with the vscode api sorry :(

ElianHugh:
unstale

renkun-ken:
Currently, the testing no longer works in all recent PRs (#1290, #1296) where test fails with [TypeError: util.config is not a function](https://github.com/REditorSupport/vscode-R/actions/runs/4151024838/jobs/7182063277#step:5:80). No clue how to resolve this. Tried for a while but nothing works.

ElianHugh:
just spitballing - is it possible that there's a namespace collision between `util.ts` and the node `util` module?

renkun-ken:
[...]

PR #1304, opened 2023-02-15
Basic unit test for RMD knitting

ElianHugh:
- Introduces unit testing for the background knitting of documents 
(Also this is an example of how we can use mocking/stubbing for testing)

Related to #990 

More tests can be added, I just thought I would get the ball rolling :)


ElianHugh:
not sure why this is failing in actions but not when I run locally

ElianHugh:
will revisit

PR #1305, opened 2023-02-17
Basic unit test for workspace viewer

ElianHugh:
- Introduces (very basic) unit testing for the workspace viewer

Also proposes unit testing through mocking the vscode API via the `sinon` library. E.g., we can "set" `vscode.window.activeTextEditor` via the following function:

```ts
export function mockActiveTextEditor(document: vscode.TextDocument, sandbox: sinon.SinonSandbox) {
    return sandbox.stub(vscode.window, 'activeTextEditor').value({
        document
    });
}
```


renkun-ken (approved):
Really nice work!

PR #1317, opened 2023-03-02
Explicit workspace behaviour

ElianHugh:
- Always show workspace viewer
- Fallback message when session watcher is disabled
- Don't error when calling view without an active workspace viewer

# What problem did you solve?
It can be confusing for users when the workspace viewer doesn't show up, especially when there is no message explicitly stating why it isn't appearing. This PR ensures that users are pointed in the right direction.

## Screenshot

image

Disable session watcher & reload --> should be greeted by welcome text

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,8 +94,7 @@
           "id": "workspaceViewer",
           "name": "Workspace",
           "icon": "./images/Rlogo.svg",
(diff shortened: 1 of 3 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #1318, opened 2023-03-02
Ensure workspace is cleared

ElianHugh:
# What problem did you solve?

- Workspace viewer will no longer show "ghost" data from previous sessions. Global env, search, and loaded namespaces will all be cleared when an attached term is closed.
- Status bar item will be reset to 'R: (not attached)'

Workspaces will be cleared if: 
- R is quit safely
- An attached *integrated* terminal is closed

## (If you have)Screenshot

After closing the terminal:


renkun-ken:
Thanks for working on this!

Ideally, if the R session exits, the workspace viewer should go empty, and the status bar item should be reset to "R: (not attached)".

In my case, I mostly use self-managed R terminals (e.g. starting radian in tmux windows manually). Not sure if there is a way to handle this so that whenever the attached R session exits, no workspace viewer and status bar item will be reset.

ElianHugh:
Hmm, that's a good point, thank you. I wonder if using a finalizer on the R side of things could be used... will have a play around with it

renkun-ken:
I tried `reg.finalizer` to the global environment before, it seems to only work if the R process is exited by `q()` but not killed externally or crashed.

A simple way is to add the following in `vsc.R`:

```r
reg.finalizer(globalenv(), function(e) .vsc$request("detach"), onexit = TRUE)
```

and handle `detach` request on extension side by resetting workspace viewer and status bar item.

ElianHugh:
I'm not sure there's a way to handle unsafe exits -- couldn't find any way to handle them. Even temp folders don't seem to be cleared when R crashes or is killed

I think having a finalizer would still be worth it in the long run? I'd like a solution for unsafe exits though

renkun-ken:
Not sure if there is a way to get notified by the exit signal of an external process in nodejs since we know the pid of the attached R session.

ElianHugh:
Good idea! Looks like it is possible to check if an arbitrary process is alive via `process.kill(pid, 0)`. See: https://stackoverflow.com/questions/14390930/how-to-check-if-an-arbitrary-pid-is-running-using-node-js

Not exactly a watcher though

renkun-ken:
The `process.kill(pid, 0)` approach seems good enough for session watcher to check every 1 or more seconds once attached so that we don't have to rely on the R session to report its exiting?

ElianHugh:
Hi @renkun-ken, would you like to try the current implementation to see if it works with your tmux setup?

renkun-ken:
I tried with the following use cases and all work properly!

* "Create R terminal" and remove terminal
* Manually start an R session, attach, and `q()`.
* Manually start an R session, `Sys.getpid()`, and `kill <pid>` externally.

renkun-ken (approved):
LGTM Thanks!

PR #1319, opened 2023-03-04
Added r.view command: View(variable)

yeyun1999:
# What problem did you solve?
Added a vscode command for the View function, which is convenient for adding shortcut keys


renkun-ken (approved):
LGTM Thanks!

PR #1321, opened 2023-03-05
Always check pid before clearing workspace

ElianHugh:
Ensure that we aren't clearing the workspace too eagerly.
- Only clean the workspace if the current pid === stored pid

# What problem did you solve?

If you attach two terminals, closing the first one will cause vscode-R to detach the second one

### DIFF
--- a/R/session/vsc.R
+++ b/R/session/vsc.R
@@ -792,4 +792,4 @@ print.hsearch <- function(x, ...) {
   invisible(NULL)
 }
 
-reg.finalizer(globalenv(), function(e) .vsc$request("detach"), onexit = TRUE)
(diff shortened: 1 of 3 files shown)


R/session/vsc.R
```diff
@@ -792,4 +792,4 @@ print.hsearch <- function(x, ...) {
   invisible(NULL)
 }
 
-reg.finalizer(globalenv(), function(e) .vsc$request("detach"), onexit = TRUE)
+reg.finalizer(globalenv(), function(e) .vsc$request("detach", pid = pid), onexit = TRUE)
```
renkun-ken, line 795: Looks like this change is not necessary since `request()` already includes `pid` in the output.

PR #1323, opened 2023-03-06
Workspace viewer command visibility

ElianHugh:
- save, clear, and refresh are disabled when there is no attached terminal
- Load can be used at any time to attach a terminal + load data easily

image

- View, remove, and quickpick are no longer shown in the command palette
   - these were not meant to be user-facing, and didn't do anything when called

image


renkun-ken (approved):
LGTM

Issue #1332, opened 2023-03-16
Is there a way to control the font size of the `rmarkdown` preview?


mihaiconstantin:
When opening the preview window to the side for `rmarkdown` files--is there a way to specify the font size? The `VS Code` setting `markdown.preview.fontSize` doesn't seem to do anything. The current *workaround* I use now is to just increase the zoom level via `window.zoomLevel`. Just curious if there is a better way to do this.

PR #1333, opened 2023-03-18
RMD Preview: font-size setting

ElianHugh:
# What problem did you solve?
Solves #1332 by injecting font-size in the preview HTML's css.

Render the following, modifying the `r.rmarkdown.preview.fontSize` setting or the `markdown.preview.fontSize` setting to change the size of text in the preview.

``````
# H1

Foo.

## H2

Bar.

### H3

Baz.

#### H4

Foo.

##### H5

Bar.

```{r}
x <- 1
x
```

``````

### DIFF
--- a/package.json
+++ b/package.json
@@ -1544,10 +1544,10 @@
           "default": true,
           "markdownDescription": "Enable automatic refresh of R Markdown preview on file update."
         },
-        "r.rmarkdown.preview.fontSize": {
-          "type": "string",
-          "default": "",
-          "markdownDescription": "Controls the font size in pixels used in the R Markdown preview. When unspecified, `#markdown.preview.fontSize#` is used."
+        "r.rmarkdown.preview.zoom": {
+          "type": "number",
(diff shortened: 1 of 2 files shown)


renkun-ken:
Does it overwrite a themed document which has its own CSS font-size of those h1-h5 headers?

ElianHugh:
It would depend on the html theme's css -- if it has higher priority styling then no. typically yes this will override that.
admittedly, i'm not sure how to handle that, should themeing be respected in previews?

renkun-ken:
Not sure if we should hard-code some font-size adjustment or just use some zoom parameter?

renkun-ken:
I tried with some scenarios, looks like adjusting font size may result in many size inconsistencies between different HTML elements such as normal text, headers, images, code blocks, etc.

I find using `zoom` seems good enough.

ElianHugh:
That sounds fair, it will work in more situations compared to just font size

PR #1334, opened 2023-03-21
Handle errors in `getAliases.R`

csaybar:
# What problem did you solve?

When an `aliases.rds` file in one of your installed packages is corrupt or improperly formatted, it can cause the documentation to fail every time it is requested, resulting in an error message that can be quite frustrating.

(x) Failed to get list of R functions. Make sure that `jsonlite` is installed and r.rpath.linux points to a valid R executable.
(x) Error in readRDS(filename) : error reading from connection Calls: lapply -> FUN -> as.list -> readRDS Execution halted

This PR adds a more informative message to the users.

### DIFF
--- a/R/help/getAliases.R
+++ b/R/help/getAliases.R
@@ -22,18 +22,15 @@ ret <- lapply(rownames(ip), function(row) {
     aliases = NULL
   )
   if (file.exists(filename)) {
-    info[["aliases"]] <- tryCatch(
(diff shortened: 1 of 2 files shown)


csaybar:
I am not sure why I cannot upload images on GH, but I am able to upload PDF files.
[bug01.pdf](https://github.com/REditorSupport/vscode-R/files/11024087/bug01.pdf)

PR #1348, opened 2023-04-24
Use zoom instead of fontSize

renkun-ken:
Use `zoom` instead of `fontSize` for better consistency between HTML elements.


ElianHugh:
LGTM

PR #1394, opened 2023-07-17
wip: start working on connection to the requests by a tcp server.

### LINKED ISSUE #1359: Allow the R `.vsc.attach()` function to connect by TCP to VSCode - supporting remote R sessions
Currently, the only way to run R session that will be connected to VSCode by this extension, is to run a remote VSCode server on this machine, and then the R terminal/session is "local" in the sense it's running on the same machine running the VSCode server. This is due the following more general issue:

Currently, this extension connects to the R session via `.vsc.attach()`, and then VSCode and the R session communicates via temporary files.
While this solution makes the usage of the R plugin simple, it would be problematic in two scenarios:
1. Connecting different R sessions on different VSCode backends - currently the attachment mechanism may conflict.
2. Connecting to a remote R session - right now the only solution is to execute VSCode remote backend on the same remote machine.

A simple solution that wouldn't ruin the current mechanism is to support communicating with the VSCode instance in two channels - one by the file-system(currently), and the second is by listening to a TCP connection.
This way the R function will become `.vsc.attach(port=NULL)` and when the `port` isn't null, it will try to attach by the TCP port.
For the UI, the VSCode extension status-bar display could be changed to "R - (not attached, TCP:**_PORT_NUMBER_**)".

Tal500:
Fixes #1359

Roadmap:
- [X] TCP R connection
- [ ] Visual&logical attaching/detaching from the TCP server
- [X] Namespaces and global environment view
- [X] View data
- [ ] Plots
- [ ] ?


R/session/vsc.R
```diff
@@ -428,8 +465,24 @@ if (use_httpgd && "httpgd" %in% .packages(all.available = TRUE)) {
           dev_args <- getOption("vsc.dev.args")
           do.call(png, c(list(filename = plot_file), dev_args))
           on.exit({
+            cur_dev <- dev.cur()
             dev.off()
+            created_devices <<- created_devices[created_devices != cur_dev]
             cat(get_timestamp(), file = plot_lock_file)
+            if (!is.na(request_tcp_connection)) {
+              tryCatch({
+                  plot_file_content <- readr::read_file_raw(plot_file)
```
renkun-ken, line 474: Does `readr::read_file_raw()` behave differently from `readBin(file, file.size(file))`?

renkun-ken: Maybe `readBin(file, file.size(file))` is good enough so that we don't have to introduce `readr` as a new dependency here.

src/liveShare/shareSession.ts
```diff
@@ -43,7 +46,17 @@ export interface IRequest {
         command: string,
         start_time: string
     };
-}
+} & (
+    {
+        coomad: 'dataview';
```
renkun-ken, line 51: Seems like a typo?

Tal500: It was my attempt to use [typescript type discramination](https://www.typescriptlang.org/play?#code/PTAEEFQEwSwZwMYCcYFsYDsCGAXAplKDgJ4AOeoArhjAPYajygDuAFnkhcbZVXBQlpQKAMwA2tZgCgQoLNjHE4THLVCcolBBSyh+OULRGhStfBhwwsYwwCMAVngQ440SQ1WGMeGWDjkEGBEYBDtHZwA6X19QABVWJlJcfCQGZlokAGtXTmtFFjwxGxEMvTwXQ2NldDEsJBjaBycK5hgcVjloIJEOPAs9HBQMAHNDJFAMSlRbDlBBDDgceRwYkvG8AA8sVFIxPAAuTrF4AyMJ7YJQPAA3PpcAGjHQW6RlegIY-VczxvCXKKkJHIcTQmGGAFFbv0ALygADe51QB1AACJFnUcCiANx8DgAZSWSHwUEOtlotD28lAAF9QAAfeGI5EohASfhQbHQShIXB0DCHSbTWbUrFSGIAdXYDBudzmtCRrkwnnaTBE1GcfMebTllDEhBmcgwxAaxnaFBw6RMZjuVhsQLwcAB80WoFY8ige0hsthAAoZRZDrFQSMvRYAJSgaEAPnhUlAoFkACFiHwwZ04K0cAgOlhhlhMC7-TgItgkXEyHg8cgYKQcAByVyCYRxhNgcSSQ3WJRMBBU4QpdDeIhugxU37NOZjjCKFuyA2cUicfgWS62FPvIgVxgedhUGj0AHxjNtbOgP1Q4ulvARuEt+O9-io9FElH7O-x1txBKuJHyVzcXhewYOAsB6fIsAQbQ4FcSh+CQAkMQ+D8PznJwsFgig2gbYcKHofJ7W3ZRhBBQcIQvd941kNheiZRhXDRQlMQouUFgMTA2isYlkwAVTgyMrgvCIMPgxiCFFZD41sXJMlFd8HwoFk2QIV9mOdAxLAVJIGFhIsIk0HlLHocSJKkvAsBkltqSkEUxVkeJEmSDgGBUXcQLLTMOkFGZXhYBJT2YAQqQwmIsFcM0ukQFBB1wDIAVkABJHcmE2bZdjwR4AtdLBbk6WBIuDXAKGoPlDUIeQQqgWBDIUK4kCQUp0XwIg1DdDAPTwAECPAAAFeKACUHVMBYHX4hEXjeflQAABhxVA4GGQ5FiGUZaQZMaOAmw4AEZZodEDhmRJawRxRrYIFKZvJpelGQ4eqkEWwY0xsqQ1Kytq9gG-x6EfX0lyG-hDh6-rBu+h0I2jWNKLAAAJd09hw2q7snfhHndHCGE4HBuQwFsgjPFFboyFFt3UEHhpvOTvopDrCaQH0-tBiJabDYz40x7HjKs2cwFiCsqxQWsJg7TIMEkMKRwgXrPv+wKMFFgwZm5hHac3cgIlAeLjG1ALOGVuqMkecK1QwDV6CV9JdUIN0cvZ1ICHVgBNHhJwYF4ghTFVXDXJXWFoN2RkMF5Sa+4bQBmCRmEPRhjHpsn+AicaSuhZPpop5DnWpiIJGGWOQ-jubhhZyyrjER88dzmWE42pOU62tOPwzvYs9oHOGeGiJTrgR42-zvbc2vTnrNk2QncoOtdZmHAUkqYxYLTXRjyzDpGrwJF+gLfAsEIIwYjxle14qGZeww0AAJd0BUHMnRoO5eQoOHXA5CKGIkiJb5TV3Yr6DkXXsycTJ7ZfhokwXQwxySECSFPJyqxSie1TAHcK3gg4EUwAjVqUAyS0EyG+GIrAp6kDgPsEAzASERHtPlWstQRgRAyMMYAUBaAIDgMAdBmDMjAC3tcO+BAAC05CIh4NQGIAAxHlas0ViQ8K-gsKQQA).
However, I have many type design issues for now, so I'm rolling back from this attempt. You can use this typing strategy later if you want to.

R/session/vsc.R
Tal500: > Maybe `readBin(file, file.size(file))` is good enough so that we don't have to introduce `readr` as a new dependency.

You need to specify the correct encoding parameters, and I'm unable to make this work with jsonlite::base54_enc function with `readBin`.

eitsupi:
Outdated.
Closing in favor of #1763

PR #1398, opened 2023-08-05
allow to invoke R terminal also in relative paths

Tal500:
Allow to invoke R terminal also in relative paths, and specify the right working directory for the workspace.

Now the user is allowed to enter a relative path in the option 'rterm', and this VSCode extension will invoke this relative path relative to the workspace directory. This is useful for example for people(like me) who have a bash script creating a remove R session.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -10,7 +10,7 @@ import * as util from './util';
 import * as selection from './selection';
 import { getSelection } from './selection';
 import { cleanupSession } from './session';
-import { config, delay, getRterm, getWorkspaceFolder } from './util';
+import { config, delay, getRterm, getCurrentWorkspaceFolder } from './util';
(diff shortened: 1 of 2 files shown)


renkun-ken:
Thanks for working on this!

Does it work properly in the following scenarios:

* Single workspace but only no or untitled document is open/active.
* Multi-root workspace
* No open workspace folder

renkun-ken (approved):
LGTM Thanks!

Tal500:
> Thanks for working on this!
> 
> Does it work properly in the following scenarios:
> 
>     * Single workspace but only no or untitled document is open/active.
> 
>     * Multi-root workspace
> 
>     * No open workspace folder

I think it should work if I remember correctly, since it used the already implemented function `getCurrentWorkspaceFolder`

PR #1415, opened 2023-09-05
Add task to run `testthat::test_file()` on current file

gowerc:
# What problem did you solve?

Added a VScode task to run `testthat::test_file()` on the currently open file. This is particularly useful when snapshot tests which don't really work in batch or if you have a large test script that you want to run but you don't want to run your entire test suite.

Likewise it looks like the failure string is slightly different for this function so I attempted to modify the regex to ensure the problem matching still works. 

## (If you have)Screenshot

In this GIF I show that the feature works and correctly errors for a failed test whilst it works correctly if I open a file that has a successful test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2081,7 +2081,7 @@
         ],
         "pattern": [
           {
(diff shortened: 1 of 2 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #1423, opened 2023-09-18
Add Option to Sync renv Cache

nateybear:
# What problem did you solve?

I frequently run into issues where the language or help server will start up and then crash because it cannot find packages (usually `jsonlite`). There is the option to add `.libPaths` in the extension currently, but the `renv` package cache is unstable and liable to change with different architectures and versions of `renv`.

Users can get "lucky" when the extension starts an R process in a directory where an .Rprofile bootstraps the `renv` environment correctly, but for situations where this does not happen (which I find happens often for me), the user is out of luck.

This pull request adds a boolean option to the extension where users can opt into automatically adding the `renv` package cache to `.libPaths` when these R processes start up. The mechanism is to write an environment variable on the typescript side that is then parsed as a boolean on the R side, following patterns elsewhere in the code. We then ensure that `renv` is in fact installed—issuing a warning but not crashing if it isn't—and then read the package cache path from `renv::paths$cache()` and add this to our `.libPaths`.

## Screenshot

When the language server starts up and `renv` is bootstrapped, the `.libPaths` look like this:

Screenshot 2023-09-18 at 10 25 59 AM

When the language server starts up somewhere that does not bootstrap `renv`, `.libPaths` looks like this:

Screenshot 2023-09-18 at 10 28 16 AM

## Is it a problem that we see packages that the user doesn't?

This change does *not* automatically bootstrap `renv` when an R terminal is created. Hence, the language and help servers may have more packages visible to them than the user's terminal does. Users familiar with `renv` should know how to rectify this, and the user who "accidentally" activates this option without knowing anything about `renv` can just install the missing packages. The major problem that this change overcomes is that `jsonlite` and `httpgd` often get installed into the `renv` package cache by the extension, and this is a problem that can't be overcome as easily as the accidental user error mentioned above.

### DIFF
--- a/R/help/getAliases.R
+++ b/R/help/getAliases.R
@@ -1,8 +1,23 @@
+.paths <- .libPaths()
+
 add_lib_paths <- Sys.getenv("VSCR_LIB_PATHS")
 if (nzchar(add_lib_paths)) {
   add_lib_paths <- strsplit(add_lib_paths, "\n", fixed = TRUE)[[1L]]
-  .libPaths(c(.libPaths(), add_lib_paths))
+  .paths <- c(.paths, add_lib_paths)
 }
+
+use_renv_lib_path <- Sys.getenv("VSCR_USE_RENV_LIB_PATH")
+use_renv_lib_path <- if (nzchar(use_renv_lib_path)) as.logical(use_renv_lib_path) else FALSE
+if (use_renv_lib_path) {
+  if (requireNamespace("renv", quietly = TRUE)) {
+    .paths <- c(.paths, renv::paths$cache())
+  } else {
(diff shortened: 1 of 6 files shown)


nateybear:
I should add that even when `renv` is bootstrapped, you have to manually maintain a dependencies.R file and make sure to add `jsonlite`, `httpgd` et al. to it, since `.libPaths` will point to the *sandbox* not the *cache*. This is another pain point for me that this pull request fixes.

renkun-ken (approved):
LGTM
Many thanks for working on this!

nateybear:
> LGTM
> 
> Many thanks for working on this!

Very welcome! I enjoyed working on it.

PR #1454, opened 2023-11-24
Improve code chunk handling in base `.R` files

### LINKED ISSUE #1453: Code chunks not being split correctly
I am facing bugs with how the extension marks R code chunks. This code does not get split correctly into three chunks:
```r
# %% Chunk 1
print("Chunk 1")

# %% Chunk 2
print("Chunk 2")

# %% Chunk 3
print("Chunk 3")
```

Instead, the `# %% Chunk 2` gets included in chunk 1:

CleanShot 2023-11-24 at 13 18 37@2x

When running "R: Select Current Chunk", we confirm the chunk is being mislabeled
CleanShot 2023-11-24 at 13 18 51@2x

**Environment:**

VSCode version: `Version: 1.84.2 (Universal)`
```
R version 4.3.0 (2023-04-21) -- "Already Tomorrow"
Platform: aarch64-apple-darwin20 (64-bit)
```

kylebutts:
This PR fixes #1453 and fixes some bugs in handling of code chunks in `.R` and `.Rmd` files. 

The first change is with `getChunks()`. Previously, for `.R` files, the scan would find the first `# %%` and mark it as the start line. Then, it would go until it found the next `# %%` and *incorrectly* mark this as the end line. The line just before the next `# %%` needs to be the end line (see #1453). This PR fixes this:
CleanShot 2023-11-24 at 15 40 35@2x

The second change is with `getCurrentChunk()`. I have simplified this code greatly and do much less work. Basically, I check for three edge cases (no chunks, line < first chunk, line < last chunk) and return early. Otherwise, I select the first chunk with `.endLine > line`. This works both for cursor within chunk and cursor between chunks. 

There is one change worth noting. Before, `getCurrentChunk()` would return undefined when `line` was after the last chunk. I instead return the last chunk and fix bugs with any commands this impacted.

## Testing

To test this, I created these two files (`temp.R` and `temp.Rmd`) and tried out the commands for running chunks, selecting current chunk, and going to next and previous chunk.

```r


# %% Chunk 1
print("Chunk 1")

# %% Chunk 2

print("Chunk 2")

# %% Chunk 3
print("Chunk 3")



```

`````r
---
---

```{r}

print("Chunk 1")

```


```{r}
print("Chunk 2")
```

```{r}
print("Chunk 3")
```

`````

### DIFF
--- a/src/rmarkdown/index.ts
+++ b/src/rmarkdown/index.ts
@@ -25,7 +25,8 @@ function isChunkStartLine(text: string, isRDoc: boolean) {
 
 function isChunkEndLine(text: string, isRDoc: boolean) {
     if (isRDoc) {
-        return (isRChunkLine(text));
+        const isSectionHeader = text.match(/^#+\s*.*[-#+=*]{4,}/g);
+        return (isRChunkLine(text) || isSectionHeader);
     } else {
         return (!!text.match(/^\s*```+\s*$/g));
     }
--- a/src/rmarkdown/index.ts
+++ b/src/rmarkdown/index.ts
@@ -31,12 +31,19 @@ function isChunkEndLine(text: string, isRDoc: boolean) {
     }
 }
 
-function getChunkLanguage(text: string) {
+function getChunkLanguage(text: string, isRDoc: boolean = false) {
+    if (isRDoc) {
+        return 'r';
+    }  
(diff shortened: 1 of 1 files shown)


renkun-ken (approved):
LGTM Thanks for working on this!

PR #1455, opened 2023-11-26
Improvements to `.R` file chunks:

kylebutts:
Made a few changes to better support `.R` file chunks. Mainly, this has to do with chunk highlighting. It doesn't really make sense to highlight the background of code cells for `.R` scripts, since it just highlights the whole thing. I copied python in highlighting the active cell (see below). I also added a "Run Current Chunk and Move to Next Chunk" command.

The code in the main file was getting quite long, so I moved to a separate folder if that's okay? Otherwise, I can put it back how it was.

CleanShot 2023-11-26 at 16 52 36@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "onCommand:r.runSelectionInActiveTerm",
     "onCommand:r.selectCurrentChunk",
     "onCommand:r.runCurrentChunk",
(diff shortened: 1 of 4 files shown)


renkun-ken (approved):
LGTM Thanks!

PR #1511, opened 2024-04-11
handling terminals created by vscode-Python

tomasnobrega:
# What problem did you solve?
If Python extension and R extension are open at the same time on VSCode a conflict emerges because Python extension creates a new terminal.

image

vscode-R currently checks for only one terminal available at (vscode.window.terminals.length === 1). I propose to fix that by actively excluding the "Deactive" terminal.

This issue has been reported before:
https://github.com/REditorSupport/vscode-R/issues/1459#issuecomment-2045382766
Note all 3 screenshots the second terminal is "Python ... Deactive" indicating the same root cause.

To recreate the bug you should install Python VSCode extension and run some Python on VSCode (anything, just to activate the extension). Then close all R terminals and attempt to run a new R line of R code by using "Ctl + Enter" or "Cmd + Enter", should run into the error identifying terminal.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -174,10 +174,18 @@ export function deleteTerminal(term: vscode.Terminal): void {
 }
 
 export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
-    if (config().get('alwaysUseActiveTerminal')) {
-        if (vscode.window.terminals.length < 1) {
-            void vscode.window.showInformationMessage('There are no open terminals.');
+    // VSCode Python's extension creates hidden terminal with string 'Deactivate'
+    // For now ignore terminals with this string
+    const ignoreTermIdentifier = 'Deactivate';
 
+    // Filter out terminals to be ignored
+    const visibleTerminals = vscode.window.terminals.filter(terminal => {
+        return !terminal.name.toLowerCase().includes(ignoreTermIdentifier);
+    });
+
+    if (config().get('alwaysUseActiveTerminal')) {
+        if (visibleTerminals.length < 1) {
+            void vscode.window.showInformationMessage('There are no open terminals.');
             return undefined;
         }
 
@@ -192,15 +200,8 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
 
     const rTermNameOptions = ['R', 'R Interactive'];
 
-    // General identifier for terminals to ignore
-    const ignoreTermIdentifier = 'Deactivate';
-
-    const validRTerminals = vscode.window.terminals.filter(terminal => {
-        // Check if the terminal name matches the R terminal options
-        const isRTerminal = rTermNameOptions.includes(terminal.name);
-        // Check if the terminal should be ignored based on a general identifier
-        const shouldIgnore = terminal.name.toLowerCase().includes(ignoreTermIdentifier);
-        return isRTerminal && !shouldIgnore;
+    const validRTerminals = visibleTerminals.filter(terminal => {
+        return rTermNameOptions.includes(terminal.name);
     });
 
     if (validRTerminals.length > 0) {
@@ -208,8 +209,8 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
         if (vscode.window.activeTerminal && rTermNameOptions.includes(vscode.window.activeTerminal.name)) {
             return vscode.window.activeTerminal;
         }
-        // Otherwise, find the first valid R terminal
-        const rTerminal = validRTerminals[0];
+        // Otherwise, use last valid R terminal
+        const rTerminal = validRTerminals[validRTerminals.length - 1];
         rTerminal.show(true);
         return rTerminal;
     } else {
@@ -220,6 +221,7 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
         return rTerm;
     }
 }
+
 
 export async function runSelectionInTerm(moveCursor: boolean, useRepl = true): Promise<void> {
     const selection = getSelection();
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -189,51 +189,36 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
     for (let i = 0; i < vscode.window.terminals.length; i++){
         msg += `Terminal ${i}: ${vscode.window.terminals[i].name} `;
     }
-    if (vscode.window.terminals.length > 0) {
-        const rTermNameOptions = ['R', 'R Interactive'];
-        if (vscode.window.activeTerminal !== undefined) {
-            const activeTerminalName = vscode.window.activeTerminal.name;
-            if (rTermNameOptions.includes(activeTerminalName)) {
-                return vscode.window.activeTerminal;
-            }
-            for (let i = vscode.window.terminals.length - 1; i >= 0; i--){
-                const terminal = vscode.window.terminals[i];
-                const terminalName = terminal.name;
-                if (rTermNameOptions.includes(terminalName)) {
-                    terminal.show(true);
-                    return terminal;
-                }
-            }
-        } else {
-            msg += `B. There are ${vscode.window.terminals.length} terminals: `;
-            for (let i = 0; i < vscode.window.terminals.length; i++){
-                msg += `Terminal ${i}: ${vscode.window.terminals[i].name} `;
-            }
-            // Creating a terminal when there aren't any already does not seem to set activeTerminal
-            if (vscode.window.terminals.length === 1) {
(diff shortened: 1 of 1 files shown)


andycraig:
Hi @tomasnobrega, thank you for submitting this PR and very nice work tracking down the cause of the error!

Looking at the screenshot, it appears that the Python extension may have recently started to use the `Terminal` property `hideFromUser` to create hidden terminals. I had a look at the [`hideFromUser` docs](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions), but it was a bit unclear to me what happens if one of those terminals becomes visible - maybe `hideFromUser` changes to `false`, or maybe it stays as `true` since it seems to be part of `creationOptions`. Since people only seem to be running into this problem while using the Python extension, I'm happy to keep the fix specific to the Python extension's hidden terminals at the moment. We can try to generalise it to all hidden terminals in the future if people still run into problems.

I'm going to leave a couple of comments on the code. After those are resolved, I'll do some manual testing. (At that point I might also need to reach out to one of the more active maintainers for help on resolving the failed CI checks.)

src/rTerminal.ts
```diff
@@ -189,51 +189,36 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
...
-                void vscode.window.showErrorMessage('Error identifying terminal! Please run command "Developer: Toggle Developer Tools", find the message starting with "[chooseTerminal]", and copy the message to https://github.com/REditorSupport/vscode-R/issues');
 
-                return undefined;
-            }
-        }
-    }
+    const rTermNameOptions = ['R', 'R Interactive'];
+
+    // General identifier for terminals to ignore
+    const ignoreTermIdentifier = 'Deactivate';
```
andycraig, line 196: At line 177, this function checks whether there are any terminals for the case when the user has set `config().get('alwaysUseActiveTerminal'` to `true`. Can you apply `ignoreTermIdentifier` there too?

src/rTerminal.ts
```diff
@@ -189,51 +189,36 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
...
-                console.info(msg);
-                void vscode.window.showErrorMessage('Error identifying terminal! Please run command "Developer: Toggle Developer Tools", find the message starting with "[chooseTerminal]", and copy the message to https://github.com/REditorSupport/vscode-R/issues');
 
-                return undefined;
-            }
-        }
-    }
+    const rTermNameOptions = ['R', 'R Interactive'];
+
+    // General identifier for terminals to ignore
```
andycraig, line 195: Can you add a comment saying why the string is `'Deactivate`'? (Like, one line mentioning that the Python extension creates hidden terminals including that string.)

src/rTerminal.ts
```diff
@@ -189,51 +189,36 @@ export async function chooseTerminal(): Promise<vscode.Terminal | undefined> {
...
+        return isRTerminal && !shouldIgnore;
+    });
 
-    if (rTerm === undefined) {
+    if (validRTerminals.length > 0) {
+        // If there is an active terminal that is an R terminal, use it
+        if (vscode.window.activeTerminal && rTermNameOptions.includes(vscode.window.activeTerminal.name)) {
+            return vscode.window.activeTerminal;
+        }
+        // Otherwise, find the first valid R terminal
```
andycraig, line 211: Current behaviour in this case is to use the _last_ valid R terminal. It would be good to keep that behaviour consistent.

andycraig (changes requested):

tomasnobrega:
Hi Andy thank you for the helpful comments, I will make the changes over the weekend

tomasnobrega:
@andycraig I just commited with requested changes. In my VS Code its working. Let me know if you need anything else

andycraig:
@tomasnobrega Thank you very much for making these changes. I think it's fine to merge now now. I have written some notes below about some manual testing that I did, which turned up an edge case, but I think we can ignore it at present.

<details><summary>Details of manual tests I ran</summary>

With a file `temp.R` containing the code `print("hello from R")`.
With the Python extension installed and a file `temp.py` containing the code `print("hello from Python")`

Setting `r.alwaysUseActiveTerminal` is `false` (the default):

Situation: There are 3 R terminals. The current terminal is the 2nd R terminal.
1. Create 3 terminals. Start R in each one. Select the 2nd terminal
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. Successful outcome: Code is run in the CURRENT (2nd) R terminal

✅ PASSED

Situation: There are multiple R terminals. The current terminal is not an R terminal.
1. Create 2 terminals. Start R in the 1st and 2nd terminals. Select the last (non-R) terminal
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. Successful outcome: Code is run in the LAST (2nd) R terminal.

✅ PASSED

Situation: Works when there is only one terminal, it is an R terminal, and it has just been created. (This did not work at one point due to a bug in VS Code.)
1. Kill the terminal that appears by default, using the trash icon
1. Create a terminal using `R: Create R terminal`
1. With the cursor in `temp.R`, run `R: Run Selection/Line`. (Do not do anything at all between steps 1. and 2., as that may affect what is considered the active terminal)
1. Successful outcome: Code is run in the R terminal

✅ PASSED

Situation: There WAS a Python terminal, but it has been closed. There is an R terminal. (This did not work at one point, probably due to a bug in VS Code or the Python extension.)
1. Create an R terminal with `R: Create R terminal`
1. With the cursor in `temp.py`, run `Python: Run Selection/Line in Python terminal` to create a Python terminal
1. Kill the Python terminal using the trash icon
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. Successful outcome: Code is run in the R terminal

✅ PASSED

Setting `r.alwaysUseActiveTerminal` is `true`:

Situation: There are no terminals.
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. OUTCOME: VS Code shows message 'There are no open terminals.'

✅ PASSED

Situation: There is an open terminal. Code is run in that terminal.
1. Create a non-R terminal with `Terminal: Create New Terminal`
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. Successful outcome: Code is run in the terminal. (It will produce a bash error since R is not open in the terminal. That is fine)

✅ PASSED

Situation: There WAS a Python terminal, but it has been closed. There are no open terminals. (This did not work at one point, probably due to a bug in VS Code or the Python extension.)
1. With the cursor in `temp.py`, run `Python: Run Selection/Line in Python terminal` to create a Python terminal
1. Kill the Python terminal using the trash icon
1. With the cursor in `temp.R`, run `R: Run Selection/Line`
1. Successful outcome: VS Code shows message 'There are no open terminals.'

❌ FAILED (no VS Code message was shown)
</details>

The tests all passed except the last one, for the situation where `r.alwaysUseActiveTerminal` is `true`, a Python terminal has been created and closed, there are now no open terminals, and the user tries to send code to the (non-existent) terminal. In this case, VS Code should display a message to the user saying that there are no open terminals, but there is no message. I had a look, and in this case there were two hidden terminals: the Python terminal (which is detected and successfully ignored by the check for 'Deactivate' in the terminal name), and also a second hidden terminal just called 'bash'. So the 'bash' terminal is being counted as a terminal. Then `chooseTerminal()` runs `return vscode.window.activeTerminal`, which is `undefined`. In all locations where `chooseTerminal()` is called it checks whether the result is `undefined` and if so does nothing. In summary: (i) In this test case, the expected message is not displayed to the user, but it should be fairly clear to the user anyway that there is no (visible) terminal to send code to; (ii) Code is _not_ run in a hidden terminal, which is good; and (iii) This is the current behaviour anyway, as the existing code also fails to handle the hidden terminals.

In future it would be nice to handle that edge case too, but this PR solves a problem that many users are experiencing and I am happy for it to be merged in its current form.

andycraig (approved):

andycraig:
@renkun-ken @ElianHugh @ManuelHentschel 
Hi all, one of the CI checks is failing (macos-latest) and another was cancelled (windows-latest) but I don't know what needs to be done about them, if anything. Could you advise? If they are fine as-is, please go ahead and merge this PR as I am happy with the code itself.

ElianHugh:
@andycraig I'm fairly certain the stops are just issues with the runner, it started happening recently and tends to resolve (at least for me) if the tests are reran until they succeed. Probably worth investigating, but I don't believe it has anything to do with this PR

eitsupi:
Perhaps a node.js version issue?
It seems that a very old node.js 16 is currently specified and obviously needs to be updated.

ElianHugh:
LGTM, thank you!

@eitsupi I'll create a new issue to track this, I'm not 100% familiar with the github actions used in this repo

PR #1525, opened 2024-05-06
Update init.R

keeo8:
There is a small error in the comment on line 19 that says "check requried package" instead of "check required package"

# What problem did you solve?
I understand this isn't much of a problem, but this is very simply a typo fix in the comment on line 19

You can check this pull request by seeing the short fix I made in the comment on line 19


renkun-ken (approved):
LGTM Thanks!

Issue #1347, opened 2023-04-21
Search for r.rterm.linux/mac  on system PATH


gdevenyi:
**Is your feature request related to a problem? Please describe.**
I should not have to provide the full path to radian in order to use it as the console

**Describe the solution you'd like**
I have radian available on PATH, vscode should just be able to use it without pointing directly to it, like it currently does transparently, for `R` if I don't specify `r.rterm.linux`

The command provided in `r.rterm.linux` should try to execute the command using the system PATH if an absolute path is not specified.

gdevenyi:
This issue is not stale.

BerriJ:
+1 for this issue. This is especially relevant for users of the nix package manager and NixOS system. There, packages are not stored in static places like `/usr/bin/radian` but the path contains a derived hash.
I.e. in my working environment:

```sh
R_environment-env ❯ which R
/nix/store/6kr6ps54j136aiw52xbbq97qx5jrla6r-R-4.4.3/bin/R
R_environment-env ❯ which radian
/nix/store/iybrnvvaqlh6j8w69zgj7w2k1vnq44r6-python3.12-radian-0.6.13/bin/radian
```
The _location_ of the executables will change, with updates to R and radian, but both can be found on my $PATH:

```
R_environment-env ❯ echo $PATH | grep radian
[...] /nix/store/iybrnvvaqlh6j8w69zgj7w2k1vnq44r6-python3.12-radian-0.6.13/bin [...]
```

The workaround I currently use is creating a symlink to radian to `.bin/radian` in my projects and setting `r.rterm.linux` to that. However, it pollutes the projects. And I'm not alone, [here](https://github.com/dwinkler1/rix-flakes/blob/99e445fb369cf015a380cf96ef6fc02796d245b8/templates/radian_vscode/flake.nix#L66-L74) is a similar approach.

To fix this:
- A new option should be introduced `Use Radian`, a bool that defines if the user wants to use radian or the base-R console. 
- vscode-R should look for R / radian on the `PATH` if `r.term.linux` is not set.

I'm willing to make a PR for this, however, I have little experience in JS, but if you direct me to the right files / folder I'd try my best.

Thanks so much, this extension is amazing!

PR #1545, opened 2024-07-17
Fix R terminal attachment when the `r.rterm` setting contains only the executable file name

astrochemx:
# What problem did you solve?

This PR fixes R terminal attachment when the `r.rterm` setting contains only the executable file name.
It takes into account if the executable file name already contains the `.exe` extension in case of the `win32` platform.

In addition, this PR:
- fixes `launch.json` and changes `outDir` to `dist` in all places to fix the extension debugging in VSCode;
- switches to the Node.js built-in `crypto` module from the deprecated external one;
- adds a proper `problemMatcher` to the `npm watch` task (the required extension `amodio.tsl-problem-matcher` is added to the `extensions.json`);
- updates `settings.json` file and fixes formatting in some files;
- fixes an error message when the extension cannot find the R path.

I would love to hear your thoughts and suggestions on this!

Looking forward to your review and feedback. Thanks!

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,8 +1,9 @@
 {
-        "recommendations": [
-                "dbaeumer.vscode-eslint",
-                "GrapeCity.gc-excelviewer",
-                "ikuyadeu.devreplay",
-		"DavidAnson.vscode-markdownlint"
-	]
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "GrapeCity.gc-excelviewer",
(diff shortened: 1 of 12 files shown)


grantmcdermott:
@astrochemx Sorry that no-one got around to reviewing your PR. But if I'm not mistaken, your requested functionality is basically subsumed by the recently merged #1347. (Albeit via the new `rterm.preferredConsoles` setting.) So I'm going to close.

(Feel free to push back if you disagree, although your PR branch is quite stale now and would need to be updated.)

PR #1560, opened 2024-09-08
Rsyntax

randy3k:
- remove the variable.other.r scope to make the syntax highlighting rule more consistent with other languages such as python
- httpgd 0.1.7 doesn't pass npm pretest
- fix a bunch of npm audit


randy3k:
@renkun-ken 
Do you mind adding me as another owner in the openvsx registry? I have decided to put more time in vscode, so it may make sense to me to be able to publish to openvsx. Thanks.

renkun-ken (approved):
LGTM Thanks!

renkun-ken:
> @renkun-ken Do you mind adding me as another owner in the openvsx registry? I have decided to put more time in vscode, so it may make sense to me to be able to publish to openvsx. Thanks.

Of course. Please tell me your login name?

randy3k:
Can you elaborate? The change supposes bring more consistency with other languages such as python.

WallBPG:
The assignment operator in R can be `=`, but is `<-` by standard practice. The variables are not highlighted correctly as a result of this change when using `<-` as an assignment operator. There could be a fix, but I don't have a lot of experience with VScode extensions.
Screenshot 2025-04-14 134754

randy3k:
Hi, the coloring for var1 indeed was the correct and intended. It is consistent with how python and other languages handle variables.

However, it does seem that the `var2 = 2` was incorrectly highlighted, it is supposed to be only highlighted within a function call, e.g. `foo(var2 = 123)`.

WallBPG:
Python and other languages handle variables differently. The light blue represents variables in all my python scripts, which is why I thought there was an unintended issue here. Can there be an option to customize syntax highlighting as users wish?

randy3k:
I am pushing some more updates in #1605. 

Below are my highlights for python and R. 

python
image
r
image

Can you put your cursor in your python `var2` equivalance and report "Inspect Editor Tokens and Scopes" like this?
image

Issue #1498, opened 2024-03-05
Convert stale bot to be manual rather than automatic


gowerc:
Admittedly I think I originally proposed adding the stale bot to automatically close out issues that users had become non-respondent to, but this was during a time when the ratio of users not replying to issues going unaddressed was a lot higher. Given that the pace of development of the extension has slowed down (that is not meant as a slight at all I am extremely grateful for all the work that is done here!) I am wondering if the bot is doing more harm that good now with actual issues getting closed due to a lack of attention rather than user non-respondence. 

I wondering if a pragmatic middle ground would be to make the stale bot manual instead of automatic. E.g. if a developer never gets a response to an issue they can just manually label it as stale to give the user 30 days to respond before it then gets automatically closed?

ElianHugh:
I think that's a fair call to make -- I was thinking something along the same lines too.

PR #1561, opened 2024-09-08
Do not mark issues as stale

ManuelHentschel:
Follow-up to #1498

Given the little activity in this repo recently, I would suggest removing the automatic stale label entirely.


randy3k:
Thanks. It makes sense.

PR #1566, opened 2024-09-16
Add R function to read a S4 object

zzxahahah:
Hi,

It is hard to read S4 objects in vscode-r compared with Rstudio

For example, in bioinformatic research, single-cell RNA-seq Seurat object.

In Rstudio it can be previewed like this 

and this is an big Seurat object which is hard to open it in the vscode

So I change the R code to read the S4 object, and build a .json file like R studio to show the structure of the object


dd793978671:
Perfect solution to my problem, good job!

PR #1579, opened 2024-12-10
Show sidebar icon only when extension is active

### LINKED ISSUE #1572: Conditional activity bar icon
Currently, the R extension icon always appears in the Activity Bar, regardless of whether there are any R-related files (like .R or .Rmd) in the workspace. It shows up even no file/fold is open. This can clutter the sidebar, especially when working on projects that don’t involve R. It also introduces unnecessary visual noise when the extension isn’t actively needed.

For example, the Java extension in VS Code only activates when it detects Java files in the workspace, at which point the “Testing” icon appears in the Activity Bar. Similarly, the Python extension activates when a Python file is opened, triggering the “Testing” icon to appear.

cbrnr:
Fixes #1572. Note that I am completely new to VSC extensions, so I've probably missed something, but I think this change is what other extensions like Julia or Python are doing.

I also tried to add `"onLanguage:r"` to the `"activationEvents"` array, but this does not seem to be necessary (it shows me a quick fix message "This activation event can be removed as VS Code generates these automatically from your package.json contribution declarations.").

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
         {
           "id": "workspaceViewer",
           "title": "R",
-          "icon": "./images/Rlogo.svg"
+          "icon": "./images/Rlogo.svg",
+          "when": "r.isActive"
         }
       ]
     },


Fred-Wu:
This doesn't seem to be working when defined under the key "activitybar" as "when" is not an option value to it.

cbrnr:
I copied that setting from the Julia extension: https://github.com/julia-vscode/julia-vscode/blob/main/package.json#L1315

I'm not saying that you are incorrect, but their activation mechanism is working for me. However, I do think that the `r.isActive` context is not defined anywhere. For reference, the Julia extension sets a similar context [here](https://github.com/julia-vscode/julia-vscode/blob/f204b194c86f1ce793799c8bbc8a308ede5dc16e/src/extension.ts#L53).

PR #1605, opened 2025-04-21
Syntax update and bump to 2.8.6

randy3k:
improve R syntax highlight in function calls.


randy3k:
> what does this changed?
> 
> 
> 
> I found the function call color cannot work:
> 
> 

it still works for me. Maybe a screenshot?

Skolmeister:
Function call highlighting doesn't work anymore for me, too:
image
Should I open an issue?

PR #1610, opened 2025-05-02
Update on AG Grid data viewer, and use infinite row model for dynamic data loading

Fred-Wu:
# What problem did you solve?

1. Fixed the issue that multiple data viewer would open by calling `View()` for the same data. The fix introduced `dataview_uuid`, and check whether the same window has been opened or not. It will create a window if it has been closed, but refresh the window if it has been opened already. 

2. Changed the AG Grid data viewer to `infinite` row model, which allows dynamic data loading, even for very large data. The changes include
      - implementation of server side data fetching, sorting and filtering by sending requests to an R session. The change was made to `showDataView` under `/src/session.ts`;
      - update of `dataview_table()` in `/R/session/vsc.R` to manage the request for sorting and filtering based on AG Grid options and data types, such as `number (numeric)`, `text (character)`, `date (Date)`, `boolean (logical)`;
      - setting cache block size at 100 when scrolling down the data viewer;
      - making R's `data.table` a required package to be installed.;
      - removal of pagination setting in AG Grid, but not yet from vscode-R setting;
      - data viewer UI updates
            - row Id column `(row)` now shows row number from 1 to N even after sorting and filtering. The header name `(row)` was replaced by `(# filtered rows) / (# total rows)`
            - Boolean column initially showed as "true/false" has been changed to "TRUE/FALSE". This is treated as a numeric column, and only `Equals` option is enabled, with hint to use 1 for TRUE, and 0 for FALSE when apply a filter on this column. 
            - filtering pane is now below header name so that one could see what have been filtered with what values. 

Would love to hear your feedbacks on this and whether this change would be something you consider to incorporate.

## (If you have)Screenshot

1. 

https://github.com/user-attachments/assets/9dd1e7db-1128-4428-bca9-86206ad1702f

2. with `n = 1e7`

https://github.com/user-attachments/assets/b8dd5bc7-4953-45b6-a2f9-5957ecf89a46

3. with `n = 1e8`

https://github.com/user-attachments/assets/da52288a-3754-4a2d-9293-8d0eca1d9020


Fred-Wu:
1. Data loading has improved with a cached version for data fetching, sorting and filtering in the data viewer. 
2. A data viewer would be rebuilt if `View()` is called again. Otherwise, it may not reflect in-memory changes if data is updated using `data.table` package. 
3. Both data name and an expression in `View()` is allowed, with maximum 500 width cut off.

Fred-Wu:
- [ ] To fix errors in checks
- [ ] To allow POSIXct date/time filed as date in the data viewer
- [ ] To remove pagination and data limit from extension options

Fred-Wu:
I think I have made all the changes I want. Would like someone interested in this to take a look, and provide feedback on errors I missed.

PR #1631, opened 2025-07-28
Allow bracketedPaste on win32 platform

heroine-x:
latest radian seems to correctly support bracketed paste, and `bracketedPaste` defaults to disabled, so leaving this switch to windows users to control might be reasonable.

### DIFF
--- a/src/rTerminal.ts
+++ b/src/rTerminal.ts
@@ -271,10 +271,8 @@ export async function runTextInTerm(text: string, execute: boolean = true): Prom
             return;
         }
         if (config().get<boolean>('bracketedPaste')) {
-            if (process.platform !== 'win32') {
-                // Surround with ANSI control characters for bracketed paste mode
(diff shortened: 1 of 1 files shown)


misaka4a92 (approved):

eitsupi:
This issue seems to be preventing the new arf console from working properly on Windows. (eitsupi/arf#19)

ElianHugh:
I'm not sure if I have the permissions for tagging a new release, I think I had problems with that in the past. I'll take a look and ping someone if not

Robinlovelace:
I'm still seeing an issue that seems to be related to this in Devcontainers FYI: https://github.com/eitsupi/arf/issues/212#issuecomment-4622412145 

Having realised that this PR was merged, I'm wondering why the behaviour persists. I will check version numbers of vscode-R as that may be the way to fix it in geocompx/docker after we added support for `arf`: https://github.com/geocompx/docker/issues/121

Issue #478, opened 2020-11-30
Refactor `any` values to typed object


Ikuyadeu:
Extends from #474

Currently, we disabled many eslint for the new implementation file due to the `any` value problem.
It can be object value by like the following link
Related rule link: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-assignment.md

Target files
* [ ] extension.ts
* [ ] rHelpPanel.ts
* [ ] RHelpProviderBuiltin.ts
* [ ] rstudioapi.ts
* [ ] session.ts
* [ ] test/suite/index.ts

PR #1635, opened 2025-08-10
Performance optimization for builtin `httpgd` plot viewer

TTTPOB:
Hi, this is my first PR to **vscode-r**, so please let me know if I’ve missed anything in the contribution process.
In the field of bioinformatics, R is often used for single-cell RNA-seq analysis, which typically involves visualizing datasets containing tens of thousands of points.

With the current **httpgd** SVG viewer in VS Code, rendering such dense plots can become very laggy and unresponsive.
On the other hand, if we disable `httpgd` and fall back to the naïve PNG viewer, we lose useful features like zoom in/out and dynamic plot resizing.

### Motivation

* Large scatter plots (e.g. tens of thousands of points) cause performance issues in the default SVG renderer of httpgd.
* Users need a way to keep httpgd’s interactive features while avoiding SVG rendering bottlenecks.
* Switching between plots in the httpgd plot history currently triggers a full re-render, even when the plot size and zoom level have not changed.

### Feature Proposal

1. **Support PNG renderer in httpgd plot viewer**

   * Added a setting to allow choosing `"png"` as the renderer for the httpgd viewer, alongside existing formats.
   * This enables users to handle heavy plots more smoothly while retaining httpgd’s interaction features.

2. **Plot history caching (planned, not yet implemented in this PR)**

   * Cache rendered plots with a key `(plotIndex, width, height)` to avoid unnecessary redraws when switching between plots of the same size and zoom level.
   * This will improve navigation speed and reduce computation overhead.

### Notes

* This PR is currently a **draft**.
* Only the PNG renderer support is implemented; the caching mechanism will be addressed in a future commit.
* As this is my first PR to vscode-r, I would appreciate guidance on whether the implementation follows the project’s conventions and where improvements can be made.

Thanks for reviewing!

### DIFF
--- a/html/httpgd/index.ts
+++ b/html/httpgd/index.ts
@@ -52,7 +52,7 @@ function postResizeMessage(userTriggered: boolean = false){
         newHeight = window.innerHeight;
         newWidth = window.innerWidth;
     }
-    if(userTriggered || newHeight !== oldHeight || newWidth !== oldWidth){
+    if(newHeight !== oldHeight || newWidth !== oldWidth){
         const msg: ResizeMessage = {
             message: 'resize',
             height: newHeight,
--- a/html/httpgd/index.ts
+++ b/html/httpgd/index.ts
@@ -27,7 +27,7 @@ let oldWidth = -1;
 
 const handler = document.querySelector('#handler') as HTMLDivElement;
 const largePlotDiv = document.querySelector('#largePlot') as HTMLDivElement;
-const largeSvg = largePlotDiv.querySelector('svg') as SVGElement;
+// The main plot element can be an <svg> or <img>, so don't assume SVG
 const cssLink = document.querySelector('link.overwrites') as HTMLLinkElement;
 const smallPlotDiv = document.querySelector('#smallPlots') as HTMLDivElement;
 
@@ -65,13 +65,7 @@ function postResizeMessage(userTriggered: boolean = false){
     }
 }
 
-function postLogMessage(content: any){
-    console.log(content);
-    vscode.postMessage({
-        message: 'log',
-        body: content
-    });
-}
+//
 
 window.addEventListener('message', (ev: MessageEvent<InMessage>) => {
     const msg = ev.data;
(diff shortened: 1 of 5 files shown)


TTTPOB:
```
ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts
19:32-46
[tsl] ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts(19,33)
      TS2345: Argument of type '{ cwd: string; }' is not assignable to parameter of type '(err: Error | null, matches: string[]) => void'.
  Object literal may only specify known properties, and 'cwd' does not exist in type '(err: Error | null, matches: string[]) => void'.

ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts
20:13-17
[tsl] ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts(20,14)
      TS2339: Property 'then' does not exist on type 'IGlob'.

ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts
20:18-23
[tsl] ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts(20,19)
      TS7006: Parameter 'files' implicitly has an 'any' type.

ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts
22:30-31
[tsl] ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts(22,31)
      TS7006: Parameter 'f' implicitly has an 'any' type.

ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts
37:19-22
[tsl] ERROR in C:\Tools\playground\vscode-R\src\test\suite\index.ts(37,20)
      TS7006: Parameter 'err' implicitly has an 'any' type.

5 errors have detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.

webpack 5.101.0 compiled with 5 errors and 1 warning in 6909 ms
 ELIFECYCLE  Command failed with exit code 1.
```

Also, when I run `pnpm build`, I got this. I don't know if I have to do anything to avoid these errors. I think this is related to #478

TTTPOB:
Also, I am not sure about commit b3de22b. I don't know if this condition can be removed.
The reason I removed it is otherwise it will stuck into a loop of resizing and redrawing the plot.

TTTPOB:
[r-2.8.6.vsix.zip](https://github.com/user-attachments/files/21715625/r-2.8.6.vsix.zip)
I made some adjustment on the test suite issue, so it builds now(changes not in this pr).
This artifact is for review.

Issue #1630, opened 2025-07-27
R linter fails with normalizePath() empty path error in callr subprocess


matgir99:
**Describe the bug**
The R linter fails with a normalizePath() error when trying to lint R files, showing "path[1]="": Impossibile trovare il percorso specificato" (Cannot find the specified path). The error occurs in a callr subprocess and prevents the linter from running, even though the R code itself has no linting issues when run directly.

**To Reproduce**

- Create an R file (e.g., [setup_r_dependencies.R](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) with valid R code
- Open the file in VS Code with the R extension enabled
- The extension automatically attempts to run diagnostics
- See error in the Problems panel: Failed to run diagnostics: ! in callr subprocess. Caused by error in lintr::lint(path, cache = cache, text = content): ! Linter linter() failed in '': Caused by error in normalizePath(): ! path[1]="": Impossibile trovare il percorso specificato

**Can you fix this issue by yourself? (We appreciate the help)**

No - this appears to be an internal issue with how the extension handles file paths when calling the lintr subprocess.

**(If applicable) Please attach `setting.json`**

// Default R extension settings - no custom configuration
"r.rterm.windows": "",
"r.rterm.mac": "/usr/local/bin/R",
"r.rterm.linux": "/usr/bin/R",
"r.rterm.option": [],
"r.source.encoding": "UTF-8",
"r.source.focus": "editor",
"r.alwaysUseActiveTerminal": false,
"r.bracketedPaste": false,
"r.sessionWatcher": true,
"r.rtermSendDelay": 8,

**Expected behavior**
 The linter should run without errors and provide appropriate linting feedback. When running lintr::lint('setup_r_dependencies.R') directly in R, it works correctly and reports "No lints found."

**Environment (please complete the following information):**

OS: Windows
VSCode Version: [your VS Code version]
R Version: 4.5.0 (2025-04-11 ucrt)
vscode-R version: [your extension version]

**Additional context**

The error message is in Italian, suggesting it may be related to locale handling: "Impossibile trovare il percorso specificato" = "Cannot find the specified path"
Direct testing shows the issue is isolated to the VS Code extension's linter integration:
- (ok) lintr::lint('setup_r_dependencies.R') works fine in R console
- (ok) callr::r(function() { lintr::lint('setup_r_dependencies.R') }) works fine in R console
[...]

PR #1681, opened 2026-02-26
fix: prevent linting of deleted temp files

toscm:
This PR fixes stale diagnostics being kept for temporary/deleted documents (for example git diff/revision views), which could accumulate in the 'Problems' panel as `path[1]="": No such file or directory` errors. See issue #1630 for details (including a video and screenshot).

Changes:
- Drop diagnostics when the URI has an empty file path.
- Drop diagnostics for unsupported URI schemes.
- Drop diagnostics for file URIs that no longer exist.
- Restrict single-server document selector to supported schemes only, e.g.:
  - `file:///Users/me/project/script.R` => selected and linted
  - `untitled:Untitled-1` => selected and linted
  - `vscode-notebook-cell:/...` => selected and linted
  - `git:/...` (diff/revision temp document) => not selected
  - other temporary/non-standard schemes => not selected

Validation:
- Reproduced with repeated open/click/close of git diff temporary revisions on my local machine (VSCode=1.109.5, R=4.5.2, MacOS=26.2, Model=MacBook Pro M2 Max)
- Confirmed stale missing-path diagnostics no longer accumulate.
- Confirmed real diagnostics for existing files are still shown correctly.
- Packaged and installed VSIX for manual verification (available for download at [toscm/vscode-R/releases/tag/v2.8.6-tempfix.1](https://github.com/toscm/vscode-R/releases/tag/v2.8.6-tempfix.1))

### DIFF
--- a/src/languageService.ts
+++ b/src/languageService.ts
@@ -2,6 +2,7 @@ import * as os from 'os';
 import { dirname } from 'path';
 import * as net from 'net';
 import { URL } from 'url';
+import { existsSync } from 'fs';
 import { LanguageClient, LanguageClientOptions, StreamInfo, DocumentFilter, ErrorAction, CloseAction, RevealOutputChannelOn } from 'vscode-languageclient/node';
 import { Disposable, workspace, Uri, TextDocument, WorkspaceConfiguration, OutputChannel, window, WorkspaceFolder } from 'vscode';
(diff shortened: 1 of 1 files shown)


randy3k:
merged in https://github.com/REditorSupport/vscode-R/pull/1697

PR #1684, opened 2026-03-14
Next major release 3.0.0-rc.0

randy3k:
The sess package is now bundled with the VSCode extension, we might move it to its own repository in the future.

Everything below this sentence was generated automatically.

# 🚀 Feature: Migrate Session Watcher to WebSockets & JSON-RPC 2.0

## 📝 Overview

This PR introduces a major architectural upgrade to vscode-R by replacing the legacy, file-based session watcher with a high-performance, token-authenticated IPC (Inter-Process Communication) mechanism. 

The new system operates over a pure **WebSocket** architecture using **JSON-RPC 2.0**, enabling reliable, bidirectional communication between the R session and the VS Code extension.

### 1. New `sess` R Package
- Introduced the `sess` package to handle the server-side IPC.
- Provides token-authenticated WebSocket connections instead of writing state to local temporary files.
- Includes comprehensive hooks for the R environment (`sess::register_hooks`) and `rstudioapi` emulation.

### 2. Client Session Manager Modernization
- **`src/session.ts` refactored:** Heavily refactored the extension's session manager to act as a JSON-RPC client/server.
- Added bidirectional RPC dispatching and handling, allowing the extension to actively query the R session instead of just reacting to file system changes.

### 3. Viewer Architecture Overhaul
- **Plot Viewer:** Refactored into a modular architecture (`httpgdViewer.ts` and `standardViewer.ts`).
- **Help & Web Viewers:** Modernized `helpViewer` and `webViewer` implementations to cleanly integrate with the new IPC layer.

### 4. Comprehensive Testing & CI Modernization
- **Integration Tests:** Added new integration test suites specifically for the modernized session (`session.test.ts`) and terminal (`terminal.test.ts`).
- **CI Pipelines:** Updated GitHub Actions (`main.yml`, `pre-release.yml`, `release.yml`) to support the new testing architecture and package building steps.

## 💡 Motivation & Background
The previous file-based session watcher (writing state to `.vscode-R` directories) was prone to race conditions, slow I/O, and was strictly one-way (R -> VS Code). 

By migrating to WebSockets and JSON-RPC 2.0:
- **Reliability:** Communication is synchronous and direct, eliminating I/O bottlenecks.
- **Security:** Sessions are now authenticated via random tokens upon initialization.
- **Extensibility:** The bidirectional protocol lays the foundation for more complex features (e.g., interactive data viewing, precise plotting queries, advanced RStudio API emulation).

## 🧪 Testing Performed
- Validated new integration test suite.
- Tested plotting via `httpgd` and standard viewers.
- Confirmed `rstudioapi` emulation and global environment variable syncing.

*** 
*Note: Ensure to check any new dependencies introduced in `package.json` and `package-lock.json` are properly installed before building.*


renkun-ken:
Thanks for the good work! I'm wondering how to make it work with self-managed R sessions, e.g. a radian console started in a tmux window?

renkun-ken:
I'm also wondering if the dependencies of `sess` could be somehow reduced so that it is more self-contained and could work with more restrictive environments where more limits of package installation are imposed.

```
[sess] Connected to VS Code

> sessionInfo()
R version 4.5.1 (2025-06-13)
Platform: aarch64-apple-darwin20
Running under: macOS Tahoe 26.4

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRblas.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[1] C.UTF-8/en_US.UTF-8/C.UTF-8/C/C.UTF-8/C.UTF-8

time zone: Asia/Hong_Kong
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] websocket_1.4.4   compiler_4.5.1    R6_2.6.1          cli_3.6.5        
 [5] later_1.4.5       tools_4.5.1       Rcpp_1.1.1        httpgd_2.0.4     
 [9] unigd_0.1.3       jsonlite_2.0.0    sess_0.1.0        lifecycle_1.0.5  
[13] rlang_1.1.7       systemfonts_1.3.1
```

randy3k:
> Thanks for the good work! I'm wondering how to make it work with self-managed R sessions, e.g. a radian console started in a tmux window?

If you run `R: Connect to R Session`, it will copy a websocket address. You can then paste it in R (with `sess` installed`) inside `sess_app`. E.g.,

```r
sess::sess_app("ws://......")
```
(at the moment of writing, it may be more convenient to copy the  whoe command with `sess::sess_app(...`.).

Theorietically, we could replace websocket with base R socketConnect but it is not very efficient. I beleive most users will also install languageserver, I guess it is also possible to absorb the package into langugeserver which is kind of bloated already.

renkun-ken:
Yes, that's true. languageserver has many dependencies already. One difference is that languageserver is not loaded in the user session while sess is. For me, less packages loaded in user session could be cleaner :)

randy3k:
> Yes, that's true. languageserver has many dependencies already. One difference is that languageserver is not loaded in the user session while sess is. For me, less packages loaded in user session could be cleaner :)

Sorry for the late reply. `sess` requires     
- websocket,
- later,
- jsonlite,
- rstudioapi,
- svglite

I think minimally we will need `websocket`, `later` and `jsonlite`. It shouldn't be that bad. AFAIK, rstudio and positron have way more dependencies.

randy3k:
I have spent sometime to make live share to work without success, finally realized that we need to whitelite our publisher id: https://github.com/REditorSupport/vscode-R/issues/1218

But It seems that microsoft is not being very active there, we might have to drop the Live share functionality here.

PR #1699, opened 2026-05-05
refactor!: replace WebSocket/TCP transport with Unix socket / Windows named pipe

eitsupi:
Replace the WebSocket-over-TCP IPC between vscode and R sessions with a Unix domain socket (Linux/Mac) or Windows named pipe transport.

### DIFF
--- a/sess/R/server.R
+++ b/sess/R/server.R
@@ -41,6 +41,14 @@ connect <- function(pipe_path = NULL, use_rstudioapi = TRUE, use_httpgd = TRUE) 
     return(invisible(NULL))
   }
 
+  # processx uses the \\?\pipe\ namespace on Windows.
+  # Normalize \\.\pipe\* paths from Node.js to improve compatibility.
+  if (.Platform$OS.type == "windows") {
+    if (startsWith(pipe_path, "\\\\.\\pipe\\")) {
+      pipe_path <- sub("^\\\\\\\\\\.\\\\pipe\\\\", "\\\\\\\\?\\\\pipe\\\\", pipe_path)
+    }
+  }
+
   print_async_msg <- function(msg) {
     prompt <- if (interactive()) getOption("prompt") else ""
     cat(sprintf("\r%s\n\n%s", msg, prompt))
--- a/sess/R/server.R
+++ b/sess/R/server.R
@@ -1,51 +1,42 @@
-#' Start the client R IPC connection
+#' Connect to the VS Code IPC server
 #'
-#' @param port Integer. The port of the VS Code WebSocket server.
-#'   If NULL, it will use SESS_PORT env var.
-#' @param token String. The authentication token. If NULL, it will use SESS_TOKEN env var.
-#' @param use_rstudioapi Logical. Should the rstudioapi emulation layer
-#'   be enabled? Defaults to TRUE.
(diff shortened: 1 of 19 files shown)


sess/R/server.R
```diff
@@ -1,175 +1,199 @@
...
-  if (is.null(token) || is.na(token) || !nzchar(token)) {
-    token <- Sys.getenv("SESS_TOKEN")
+  if (is.null(pipe_path) || is.na(pipe_path)) {
+    pipe_path <- Sys.getenv("SESS_PIPE")
   }
 
-  # Derive file path for fallback/reconnection
+  # Fallback: read from session JSON file written by the extension
   pid <- Sys.getpid()
   home <- path.expand("~")
```
eitsupi, line 27: This should refer to My Documents on Windows, so it's probably not what you intended.

randy3k:
thanks for working on it, didn't know that `_unix_socket` works for windows named pipe. LGTM at the first glance, will take a closer look tonight.

renkun-ken:
This branch works for me on macOS. It looks like the dependencies are reduced 👍 .

renkun-ken:
It would be ok if either https://github.com/eitsupi/vscode-R/pull/1 (which fixes the use case of unmanaged R session) could be merged into this PR, or after this PR being merged into master and I resubmit that PR based on latest master.

eitsupi:
Merged, thank you!

sess/R/server.R
randy3k: Ya, it is not ideal. It may not be the same "~" inferred by VSCode.

src/session.ts
```diff
@@ -256,6 +257,103 @@ export async function getGlobalPipePath(): Promise<string> {
...
+}
+
+export async function getAttachSessionCommand(): Promise<string> {
+    const pipePath = await getGlobalPipePath();
+    const sessPath = extensionContext.asAbsolutePath('sess').replace(/\\/g, '/');
+    const scriptPath = getAttachSessionScriptPath(pipePath);
+    await fs.writeFile(scriptPath, buildAttachSessionScript(pipePath, sessPath), { encoding: 'utf-8' });
+    attachSessionScriptPath = scriptPath;
+
+    return `source(${asRStringLiteral(scriptPath)})`;
```
randy3k, line 310: It is a good idea.

src/session.ts
```diff
@@ -256,6 +257,103 @@ export async function getGlobalPipePath(): Promise<string> {
...
+    if (pipePath.endsWith('.sock')) {
+        return pipePath.replace(/\.sock$/, '.R');
+    }
+    const scriptBase = path.basename(pipePath).replace(/[^a-zA-Z0-9_.-]/g, '_') || 'attach_session';
+    return path.join(tmpDir(), `${scriptBase}.R`);
+}
+
+function buildAttachSessionScript(pipePath: string, sessPath: string): string {
+    return [
+        'local({',
```
randy3k, line 274: i wonder how we could make it consistent with the install_sess.R.

eitsupi:
Is there a reason why the merge can't be done?
I think it would be better to merge it first and then let everyone else proceed with the subsequent work.

randy3k:
SG, let me merge it.

PR #1702, opened 2026-05-09
Security: TLS Certificate Validation Disabled

tomaioo:
## Summary

Security: TLS Certificate Validation Disabled

## Problem

**Severity**: `Critical` | **File**: `src/helpViewer/cran.ts:L17`

In src/helpViewer/cran.ts line 17, the code sets `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'` to bypass TLS certificate verification. This creates a serious security vulnerability allowing man-in-the-middle attacks when fetching packages from CRAN mirrors.

## Solution

Remove the TLS bypass. If there are specific CRAN mirrors with certificate issues, use a custom CA certificate instead of disabling validation globally. The finally block attempts to reset it but this pattern is unsafe.

## Changes

- `src/helpViewer/cran.ts` (modified)

### DIFF
--- a/src/helpViewer/cran.ts
+++ b/src/helpViewer/cran.ts
@@ -25,7 +25,6 @@ export async function getPackagesFromCran(cranUrl: string): Promise<Package[]> {
     for(const site of cranSites){
         try{
             // fetch html
-            process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'; // seems to fail otherwise?
             const res = await fetch(site.url);
             const html = await (res).text();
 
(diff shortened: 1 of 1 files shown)


src/helpViewer/cran.ts
```diff
@@ -25,17 +25,13 @@ export async function getPackagesFromCran(cranUrl: string): Promise<Package[]> {
     for(const site of cranSites){
         try{
             // fetch html
-            process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'; // seems to fail otherwise?
```
renkun-ken, line 28: @ManuelHentschel any idea why we introduced it in the first place?

benz0li: Maybe this was introduced to make it work in corporate networks behind a proxy that does TLS interception (and most likely using self-signed certificates).

If so: This is something the user can take care of.
ℹ️ Simply set the standard Node.js environment variable `NODE_EXTRA_CA_CERTS` so that custom SSL certificates are read.

Cross reference:

* https://github.com/rstudio/rstudio/issues/13718#issuecomment-1925674143

ManuelHentschel (approved):

src/helpViewer/cran.ts
ManuelHentschel: If I remember correctly, fetching the sites did not work on my personal computer/network, unless I set this env variable, but now it seems to be working fine without it.

PR #1717, opened 2026-06-30
feat(dataview): read data rows truely on demand

Fred-Wu:
## Summary

This PR attempts to introduce a true on demand data loading from R.  

The RC already introduced a dynamic data viewer, but R still prepared data for all columns when the viewer opened. This included creating the full row index, splitting matrix columns, and converting an entire Arrow object data to data frame. Values were also formatted as text.

### On-demand loading

- Keep the original R data without preparing all columns.
- Read only column information initially.
- Read and convert only the requested rows during scrolling.
- Support R data frames, matrices, Arrow data and Polars DataFrames.
- Preserve numeric, logical, date and text values instead of formatting everything as text.
- Sorting and filtering now remember the matching row order, so scrolling does not repeat the same work. 

### Data Viewer Changes
- Disables sorting and filtering for:
  - Complex number columns
  - List columns
  - Raw columns
  - The row-number column
- Sorts ordered factors according to their declared R level order.
- Show `(filtered/original)` in the row-number heading when filtering is active.

If this is accepted, and works well, probably could remove those 2 pagination settings as well. Here is an example to open 10M records with 100 columns. 

Current data loading

https://github.com/user-attachments/assets/c5f50515-e4d4-49de-bb57-4270ca8a325a

New data loading

https://github.com/user-attachments/assets/65e87917-5843-4bbb-bd42-e484ac3aa0bd

### DIFF
--- a/sess/DESCRIPTION
+++ b/sess/DESCRIPTION
@@ -16,6 +16,7 @@ Imports:
     methods,
     rstudioapi
 Suggests:
+    bit64,
     jgd,
     svglite,
     tinytest
--- a/sess/R/dispatch.R
+++ b/sess/R/dispatch.R
@@ -4,7 +4,8 @@ ipc_write <- function(data) {
   con <- .sess_env$con
   if (is.null(con)) return(invisible(FALSE))
 
-  line <- paste0(jsonlite::toJSON(data, auto_unbox = TRUE, null = "null", force = TRUE), "\n")
+  json <- jsonlite::toJSON(data, auto_unbox = TRUE, null = "null", force = TRUE)
+  line <- c(charToRaw(enc2utf8(json)), as.raw(0x0a))
   tryCatch(
     {
       remainder <- processx::conn_write(con, line)
--- a/sess/R/handlers.R
+++ b/sess/R/handlers.R
@@ -329,13 +329,21 @@ get_column_def <- function(name, field, value) {
   if (!is.null(units)) {
     tooltip <- sprintf("%s, units: %s", tooltip, toString(units))
   }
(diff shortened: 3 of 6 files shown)


Fred-Wu:
Just made a few more small enhancements. 

1.  A NA value is now displayed as  image

3. Added row position indicator to show current row out of total when clicking on or dragging scrollbar. This also applies to paginated data viewer. 

https://github.com/user-attachments/assets/2b87d90f-61e7-4b71-b73a-d2b0802efef3

I would suggest to remove the pagination setting. The lazying data loading sets the batch number at 500. Opening and scrolling are fast. However, one may set page size to, e.g. 10,000, this actually slows things down as `cacheBlockSize` equals to page size. Transferring data from R to data viewer would then be slower.

Fred-Wu:
Just identified some issues when I built and used this under the new vscode-R architecture. Data with many columns and with free texts may encounter failure to fetch data, although it has no issue if data is loaded via `httpuv` under the old architecture.

Fred-Wu:
The issue should be fixed. 

Previously, large JSON-RPC responses were converted from R character strings to raw bytes inside processx::conn_write(), which delayed transmission.  And the extension repeatedly concatenated and rescanned the accumulated response whenever a new socket chunk arrived. 

The fix converts JSON to UTF-8 raw bytes before writing, and buffers incoming socket data as bytes, decoding each complete message only once. 

With my own data, the data viewer now opens instantly for a data containing 5M rows and 150 columns with free text and grouping text fields.

eitsupi (approved):

eitsupi:
The CI failre is unrelated.

PR #1730, opened 2026-09-19
Pre release chores

Fred-Wu:
1. Fixed a bug introduced in #1713 that R6 members do not appear after expansion in the workspace viewer. 
2. Updated CHANGLOG to reflect other features introduced.


CHANGELOG.md
```diff
@@ -16,11 +17,18 @@
...
+* feat(dataview): load data rows on demand while scrolling, with support for Arrow and Polars DataFrames
+* feat(workspace): support recursive expansion of nested lists, environments, pairlists, S4 objects, and data frames
 
 ### Performance
 
 * perf: optimize package monitoring in helpServer.R
 
+### Other
+
+* Remove obsolete `r.session.objectLengthLimit`, `r.session.objectTimeout`, and `r.session.levelOfObjectDetail` settings following the switch to on-demand workspace inspection
```
eitsupi, line 31: @grantmcdermott I believe that meaningless settings should be deleted, and I think `r.plot.useHttpgd` should be deleted in the same way.

eitsupi (approved):

CHANGELOG.md
grantmcdermott: `r.plot.useHttpgd` is not deprecated yet, so I wouldn't do that with this release. (I want to give users at least one release cycle worth of warning.)

grantmcdermott:
3. Update `sess` package. Both the version, so that it matches the (updated `languageserver` when that lands), as well as some internals (`maintainer` field of the `DESCRIPTION` file, installation details in the README, etc.) I'm obviously pro a CRAN submission too, but we should make it available on R-universe regardless. I couldn't see a https://github.com/reditorsupport/reditorsupport.r-universe.dev repo, though...

eitsupi:
> 3. Update `sess` package. Both the version

My understanding is that it shouldn't be included in the release notes because the sess package wasn't included in previous versions.

CHANGELOG.md
eitsupi: Ok, I've opened #1733

PR #1731, opened 2026-09-19
ci: fix build and tests

eitsupi:
## Summary

- update `@vscode/test-electron` to 3.1.0 so current macOS bundles are launched through `CFBundleExecutable`
- run Xvfb only on Linux and run tests directly on macOS and Windows
- move CI to Node.js 24 and Node 24-based v5 official actions
- remove the obsolete programmatic test runner left after the `vscode-test` CLI migration

## Root cause

VS Code 1.110 renamed the macOS executable from `Electron` to `Code`.
The compatibility symlink was subsequently removed, while `@vscode/test-electron` 2.5.2 continued to launch `Contents/MacOS/Electron`.

As a result, VS Code 1.138 downloaded successfully on macOS ARM64, but failed with `ENOENT` before the Mocha integration suite started.

`@vscode/test-electron` 3.1.0 resolves the executable using the app bundle's `CFBundleExecutable`, while retaining compatibility with older bundles.


eitsupi:
See also #1732

Fred-Wu (approved):

PR #1733, opened 2026-09-19
fix: deprecate useHttpgd setting

eitsupi:
## Summary

- formally deprecate `r.plot.useHttpgd` with VS Code's `markdownDeprecationMessage` and `deprecationMessage`, pointing users to `r.plot.backend`
- describe `r.plot.backend` as the canonical backend selector and clarify the legacy compatibility precedence
- update `r.plot.devArgs` and `r.plot.format` descriptions to refer to the `standard` backend instead of the legacy boolean
- document that `r.plot.useHttpgd` remains supported for existing configurations but will be removed in a future release, without promising a specific version

## Compatibility

The backend resolution behavior is unchanged: `r.plot.useHttpgd: true` still selects httpgd while `r.plot.backend` is `auto`, and an explicit `r.plot.backend` value takes precedence. The integration tests now cover both cases.

## Validation

- `npm ci`
- `npm run pretest`
- `npx eslint src/test/suite/terminal.test.ts`
- `git diff --check`

`npm test` was not run locally because `xvfb-run` is unavailable in this environment. `npm run build` completed the JavaScript build and resource copy, then stopped because the local R installation does not have the `remotes` package; CI installs `remotes` in a separate step.


Fred-Wu (approved):

Fred-Wu:
Not related to this PR, @eitsupi @grantmcdermott  but I think we need to review [Plot viewer](https://github.com/REditorSupport/vscode-R/wiki/Plot-viewer) and `r.plot.defaults.colorTheme` as well given there are 3 backends now.

PR #1734, opened 2026-09-19
Removed the 'showObjectSize' setting and obsolete tooltip support

Fred-Wu:
I think the `sess` rewrite stopped calculating object sizes and reading the `showObjectSize` setting. 

Given the introduction of deferred workspace scans to fix submission delays, calculating object size could cause execution delays again for large objects. So I propose deleting the setting and tooltip support entirely unless we want to wire it back.


Fred-Wu:
I will merge #1730 after this one gets approved or decided.

eitsupi (approved):

PR #695, opened 2021-07-02
Integrate vscode-r-lsp

renkun-ken:
Closes #691 

1. Disable vscode-r-lsp
2. Open an R project and everything should behave in the same way with using vscode-r-lsp.

PR #1735, opened 2026-09-19
feat!: simplify R executable / console path settings

### LINKED ISSUE #597: Combine rterm/rpath settings for different operating systems?
We have separate settings for R paths on different platforms:

- `r.rpath.windows`, `r.rpath.mac`, `r.rpath.linux`
- `r.rterm.windows`, `r.rterm.mac`, `r.term.linux`

But as far as I can tell we don't handle these differently in the code. We check the user's OS, then choose the setting for that OS. So I think we could just have `r.rpath` and `r.rterm`. This would simplify things for the user (fewer settings) and for our code.

The downside to this change would be that a lot of users would already have those settings set. I think settings can be automatically migrated - I believe this was done recently when the vscode-jupyter extension was split out from the vscode-python extension. I could look into how they did it and see whether it seems practical. If the user has multiple of those settings set, we could show a message to ask them to choose which one to retain.

### LINKED ISSUE #1437: Executable name is resolved as relative terminal path
The code in #1398 resolves paths too eagerly as "relative". In particular, a bare executable name such as `radian` is taken to be a relative path and looked up in the current working directory instead of in $PATH (this should only happen for `./radian`).

eitsupi:
Close #597
Close #1437
Supersede #1625

Related to #1130

## Summary

- add `r.executablePath` as the canonical vanilla R executable for language-server and other background processes
- add `r.consolePath` as the canonical interactive console executable, including bare executable names resolved from `PATH` (for example, `arf` or `radian`)
- formally deprecate the six platform-specific `r.rpath.<platform>` and `r.rterm.<platform>` settings while preserving them as compatibility inputs
  - Show a one-time compatibility warning when an installed R Debugger version still registers the legacy `r.rpath.<platform>` settings
- remove the unreleased `r.rterm.preferredConsoles` setting and the previously deprecated, unused `r.helpPanel.rpath` setting
- add host-independent regression coverage for precedence, PATH and registry discovery, substitutions, quoting, and workspace edge cases
- resolve executable settings and `${workspaceFolder}` against the relevant workspace or document resource when one is available

## Why these names?

The names follow the convention used by other VS Code language extensions: for example, Python uses `python.defaultInterpreterPath`, Julia uses `julia.executablePath`, and Ruby LSP uses `rubyLsp.rubyExecutablePath`. `executablePath` clearly identifies the vanilla executable used by the extension, while the parallel `consolePath` name preserves vscode-R's important distinction between background R and an alternative interactive console.

We cannot use the shorter `r.rpath` and `r.rterm` while retaining their legacy platform-specific children: VS Code treats dotted configuration identifiers hierarchically, so scalar parent settings conflict with `r.rpath.windows`, `r.rterm.linux`, and the other existing keys. The `Path` suffix also makes clear that these settings select an executable location (including a name resolved from `PATH`), rather than configuring console behavior more broadly.

## Resolution rules

Background/vanilla R resolves in this order:

1. `r.executablePath`
2. legacy `r.rpath.<current platform>`
3. `R` on `PATH`
4. Windows registry `InstallPath` (Windows only)

The interactive console resolves in this order:

1. `r.consolePath`
2. legacy `r.rterm.<current platform>`
3. an explicitly configured `r.executablePath`
4. system R (`PATH`, then the Windows registry)

The distinction between background R and the interactive console remains intentional: background processes never use `r.consolePath` or legacy `r.rterm.<platform>`. Also, a legacy `r.rpath.<platform>` override does not begin controlling the console; only the new canonical `r.executablePath` can provide that default.

Both canonical settings use `machine-overridable` scope because executable locations are machine-dependent while workspace overrides remain useful. Relevant terminal, language-server, task, R Markdown, C++ properties, and command call sites now pass their workspace or document URI so folder-level settings and `${workspaceFolder}` substitution use the same resource. Empty values are treated as unset. Explicit but unavailable values fail instead of silently selecting a different executable.

## Historical compatibility

The old settings were split by OS when defaults were hard-coded installation paths. The resolver keeps the later discovery behavior added for real-world installations:

- `R` on `PATH` takes priority over the Windows registry (#244, #252, #649)
- background commands use vanilla R, independently of alternative consoles (#607, #695 and commit `5e9c895`)
- `${userHome}`, `${workspaceFolder}`, `${fileWorkspaceFolder}`, and `${fileDirname}` substitutions remain supported (#1398, #1444)
- quoted paths and paths containing spaces remain supported (#884)
- a bare name is looked up on `PATH`, never interpreted as a workspace-relative path (#1347, #1437)

The platform-specific settings remain supported for backward compatibility but are deprecated in favor of the canonical settings and may be removed in a future release.

## Scope boundaries and deliberate follow-ups

- `.github/ISSUE_TEMPLATE/bug_report.md` intentionally remains unchanged even though it contains legacy setting examples. The template needs a broader rewrite, which will be handled separately rather than partially updating it in this PR.
- This PR makes executable configuration lookup resource-aware, but does not introduce a complete per-folder runtime lifecycle. In particular, it does not restart the help subsystem when the active folder changes or redesign single-/multi-server LSP lifecycle behavior.


Fred-Wu:
Deprecation warning for `r.plot.useHttpgd` and `r.rterm.x` works, but seems not for `r.rpath.x`.

eitsupi:
> Deprecation warning for `r.plot.useHttpgd` and `r.rterm.x` works, but seems not for `r.rpath.x`.

Are you using VSCode-R-Debugger too?
The most likely cause is a duplicate setting ID (like #1492).

Fred-Wu:
> > Deprecation warning for `r.plot.useHttpgd` and `r.rterm.x` works, but seems not for `r.rpath.x`.
> 
> Are you using VSCode-R-Debugger too? The most likely cause is a duplicate setting ID (like #1492).

Yes. I have that installed. Uninstalling/Disabling it makes deprecation work.

eitsupi:
Thanks for confirming!

Fred-Wu:
> > Deprecation warning for `r.plot.useHttpgd` and `r.rterm.x` works, but seems not for `r.rpath.x`.
> 
> Are you using VSCode-R-Debugger too? The most likely cause is a duplicate setting ID (like #1492).

Maybe we should provide a one-time warning message if users also have debugger extension installed.

eitsupi:
How about 733dbd3?

Fred-Wu:
> How about [733dbd3](https://github.com/REditorSupport/vscode-R/commit/733dbd3929b1b18e33c508d4433f855862af00e3)?

Tested and LGTM.

Fred-Wu (approved):

eitsupi:
This is a very big change, but I want to merge it now to resolve long-standing issues.

If there are objections, we can revert it before the 3.0.0 release.

grantmcdermott:
@eitsupi two notes on this, since I've just had a chance to test:

1. I don't think backwards compatibility is ensured? I was confronted by the following message after trying to start a new R session after updating. (For the record: I had an existing `r.term.mac` setting that I had to comment out, and then had to explicitly specify `r.consolePath: "arf"` on top of that. Speaking of which...

Screenshot 2026-09-21 at 9 58 55 AM

2. ... the `r.consolePath` dialog box could be improved IMO. Based on the current entry, I don't think it's clear that`"arf"`, `"radian"`, etc. are valid input strings that trigger automatic discovery. (I only knew because I read your PR description above.)
 
Screenshot 2026-09-21 at 10 01 34 AM

eitsupi:
@grantmcdermott Thanks for testing.
One detail I'd like to confirm: the legacy setting is `r.rterm.mac` (with two `r`s), and that key is still explicitly supported.
Could you confirm the exact key and value you had configured?

ChatGPT did find a UX issue regardless: when a legacy setting is present but cannot be resolved, the current error only mentions `r.consolePath`, which makes it look like the legacy setting was ignored.
I'll improve that message and avoid the duplicate error.
I also agree that the `r.consolePath` description should explicitly say that values like `arf` and `radian` are resolved from `PATH`.

grantmcdermott:
> the legacy setting is `r.rterm.mac` (with two `r`s)

Yes, `r.rterm.mac`. Sorry, typo in my original post. I had it pointing to `"/opt/homebrew/bin/arf"`... But I just tried it again (switching between both arf and vanilla R) and now it's working? I'm not sure what happened in between, but I think we can ignore my original comment as noise. (Sorry!)

grantmcdermott:
> I also agree that the `r.consolePath` description should explicitly say that values like `arf` and `radian` are resolved from `PATH`.

P.S. One other little dialog box nit that would be simple to roll into the same fix if you have time: on `r.plot.backend`, it should say _"...For **backwards** compatibility, setting..."_.

In other words add "backwards" (or "legacy") before "compatibility".

eitsupi:
#1762

PR #1738, opened 2026-09-20
Fix/data viewer display

### LINKED ISSUE #1408: View() fails to handle lubridate Period objects within a data structure
**Describe the bug**
In VS Code, View() displays only the seconds of Period objects when in some sort of larger data structure, like a tibble, data frame or a list. Haven't tested with other data structures. 

> View(tibble(x=lubridate::period(num = 20, unit="hours")))

Behaviour in VS Code with the R extension:
image

**To Reproduce**

1. Use VS Code with the R Extension
2. Try to View a Period object within some data structure like so: View(tibble(x=period(num=20, unit="hours")))
3. See View() output

**Can you fix this issue by yourself? (We appreciate the help)**

No :(

**Expected behavior**
Normal Rstudio behaviour:
image

- OS: Ubuntu 22.04.3 LTS
- VSCode Version: 1.81
- R Version: 4.3.0
- vscode-R version: 2.8.1

### LINKED ISSUE #1463: View() displays sets some numerical values to 0
I noticed that `View(dataframe)` sometimes sets small numerical values (such as 2.5e-05) to 0. However, in the same dataframe, I also have values that are much smaller (1e-240) which get displayed correctly. 

I have attached a screenshot of my observations. Here you can see that `p` column for the `SNP` "1_247598484_G_A" is displayed as 0.
view_bug

However, when I subset my data to only contain this value, you can see that the p value is actually different:
view_bug2

The `str` of the `p` column is `num`. 

I have uploaded the dataset for which I observe this to wetransfer:
https://we.tl/t-3WxDaKx1Vn

I am using vscode-R version v2.8.2 and VScode v.1.85.1 on Windows11.

I appreciate any in that matter!

Fred-Wu:
Closes #1463 
Closes #1408 

1. The fix disables JSON's default decimal rounding for all numeric values. AG Grid continues to use JS default formatting, which switches to scientific notation below 1e-6. 

    `1.54e-5` would appear as  `0.0000154` in the data viewer instead of 0.
    `1.54e-7` would appear as `1.54e-7` in the data viewer. 

2. The fix adds one shared formatting function that calls R's `format()` generic to use each class's formatting method when displaying its values. If a class does not implement a `format.xxx()` method, R uses `format.default()` by default.


sess/inst/tinytest/test-dataview.R
eitsupi, line 1: Some tests seem droped (for `dataview_init`, `dataview_dispose`).
Is this intentional?

eitsupi (commented):
Could we also add a regression test for an S4 numeric subclass?
The new synthetic class test is S3, whereas `lubridate::Period` is an S4 class extending numeric, which is the exact shape that previously fell into the numeric-column branch.
A minimal synthetic S4 class would test the new is.object() behavior without adding lubridate as a dependency.

sess/inst/tinytest/test-dataview.R
Fred-Wu: Ah, they were mistakes when I tried to move display tests from test-ipc.R. I will restore them.

Fred-Wu:
> Could we also add a regression test for an S4 numeric subclass? The new synthetic class test is S3, whereas `lubridate::Period` is an S4 class extending numeric, which is the exact shape that previously fell into the numeric-column branch. A minimal synthetic S4 class would test the new is.object() behavior without adding lubridate as a dependency.

I will add a test for that. But the fix should be general for numeric values with an arbitrary class, which is based on both `is.numeric()` and `is.object()` checks.  If both are TRUE, they will enter `textColumn`  branch and values will get formatted based on the class's `format()` method.

eitsupi (approved):

PR #1747, opened 2026-09-22
Add data viewer column panel

### LINKED ISSUE #1586: Data viewer auto width.
Type: <b>Feature Request</b>

In R when we use View(df) some of the column on the far-right of the viewer are quite narrow. So I am requesting a setting to edite the autowidth feature. 
OR a way to select all column and double click on border of one to expand the width as needed (similar to excel). right now you have to do that one by one for each column. 

Thanks

Extension version: 2.8.4
VS Code version: Code 1.96.3 (Universal) (91fbdddc47bc9c09064bf7acf133d22631cbf083, 2025-01-09T18:14:09.060Z)
OS version: Darwin arm64 24.1.0
Modes:

Fred-Wu:
## Summary

Added a column panel to the Data Viewer to expose all columns in the same place with the following features. 

1. Select or de-select all columns. 
2. Select or de-select individual columns. 
3. Reorder columns directly from the panel by dragging them up and down. 

This is a pure UI change, which only modifies AG Grids column state. They do not affect data loading, sorting or filtering. The viewer continues to track columns by column ID rather than display position. It is more close to the new Results Viewer from the SQL Server extension. This enhancement would help those working with wide data or data with many columns, and potentially closes #1586. 

https://github.com/user-attachments/assets/97702d42-ab7f-4011-ae6a-82fdcfcc62e8


eitsupi (commented):
Reviewed with ChatGPT as a second pair of eyes. Overall this looks good to merge, with one small UI consistency concern:

The row-index column (`colId === "0"`) is excluded from the panel, but panel reordering always puts it back at the first position. If moving that column in the grid is currently allowed, this could be surprising. It may be better either to make it explicitly fixed, or preserve its current position.

Also, I don't think this fully closes #1586, since that issue is mainly about auto-sizing column widths. "Related to #1586" may be more accurate.

Other than that, I didn't find any blocking issues, and CI is green.

eitsupi (approved):

